PR #12, opened 2023-06-14
QuickSettings Foundation

davidjiagoogle:
Created the feature/quicksettings module

This module has dependency on feature/settings for SettingsViewModel
In the future, it could eliminate dependency on settings and have its own data module, which would be a lightweight version of data/settings, whose repository only selectively fetch a subset of jcaSettings. But for now, it piggybacks off of existing settingsViewModel
This module is depended by feature/preview because it has to show up in PreviewScreen
Created QuickSettingComponents

QuickSettingsUi is the drop down arrow. It has a composable below that spans the whole screen. When retracted, this composable is empty. When expanded, this composable is populated by ExpandedQuickSettingsUi.
ExpandedQuickSettingsUi is the composable with quickSetting items inside. It can be further expanded when one of its items is clicked.
QuickSettingUiItem is the composable for each quickSetting item.
Created ExpandedQuickSettingsDataModel

When an item is clicked, we need to change the content of ExpandedQuickSettingsUi to display selectable options of this item. ExpandedQuickSettingsDataModel contains all the data information to populate this display.


data/settings/src/main/java/com/google/jetpackcamera/settings/model/QuickSettings.kt
```diff
@@ -0,0 +1,27 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.jetpackcamera.settings.model
+
+data class QuickSettings(
```
jsaund, line 19: A couple nits:
1. use camel case instead of snake case
2. you can set the default value to be false or if you need to know if it's a default instance then you create a `const val DEFAULT_QUICK_SETTINGS = QuickSettings()`

```
data class QuickSettings(val defaultFrontCamera: Boolean = false)
const val DEFAULT_QUICK_SETTINGS = QuickSettings()
```

feature/quicksettings/src/main/java/com/google/jetpackcamera/feature/quicksettings/QuickSettingsComponents.kt
```diff
@@ -0,0 +1,217 @@
...
+import androidx.compose.ui.res.painterResource
+import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.unit.dp
+import androidx.hilt.navigation.compose.hiltViewModel
+import com.google.jetpackcamera.quicksettings.R
+import com.google.jetpackcamera.settings.SettingsViewModel
+
+@Composable
+fun QuickSettingsUi(modifier: Modifier = Modifier) {
+    var isOpen by remember {
```
jsaund, line 60: Do we need to track `isOpen`?
Or should Preview be responsible for this?
For example in the Preview Composable would we have something like:
```
fun Preview(..) {
   if (uiState.isQuickSettingsOpen) {
       QuickSettingsUi(...)
   }
}
```

PR #15, opened 2023-06-27
Fix problem with LaunchedEffect having to depend on cameraState

yasith:
Also add some logging to view model functions.

The coroutine that's suspended awaiting the deferredSurfaceProvider wasn't being resumed because the deferredSurfaceProvider object was different each time.

This fix addresses that.

### DIFF
--- a/feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewScreen.kt
+++ b/feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewScreen.kt
@@ -16,7 +16,6 @@
 
 package com.google.jetpackcamera.feature.preview
 
-import android.provider.MediaStore.Video
 import android.util.Log
 import androidx.camera.core.Preview.SurfaceProvider
 import androidx.compose.foundation.Canvas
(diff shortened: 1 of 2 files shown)


feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewViewModel.kt
```diff
@@ -52,6 +52,7 @@ class PreviewViewModel @Inject constructor(
 
     private fun initializeCamera() {
         // TODO(yasith): Handle CameraUnavailableException
+        Log.d(TAG, "initializeCamera")
```
jsaund, line 55: nit: do you want to remove the log statements? ok to leave them in just wondering if they were for debugging purpose

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewScreen.kt
```diff
@@ -66,13 +67,13 @@ fun PreviewScreen(
 
     val lifecycleOwner = LocalLifecycleOwner.current
 
-    val deferredSurfaceProvider = CompletableDeferred<SurfaceProvider>()
+    val deferredSurfaceProvider by remember { mutableStateOf(CompletableDeferred<SurfaceProvider>())}
```
jsaund, line 70: nit: this probably isn't a mutable state right?
```
val deferredSurfaceProvider = remember { CompletableDeferred<SurfaceProvider>() }
```

jsaund (approved):

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewViewModel.kt
yasith: Intentionally adding them to help with tracking down future issues.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewScreen.kt
yasith: Ah yes, fixed.

PR #16, opened 2023-06-30
Connect CameraUseCase to settings data and implement rear camera flash

Kimblebee:
This is a PR that enables the user to change the default flash mode of the rear camera from the Settings Screen. Here is some info on the more notable changes in the files.

**Misc. Changes**
changed the name of the `Settings` class to `CameraAppSettings`

**Changes to Datastore**
created `flash_mode.proto` enum schema representing flash mode status of on, off, and auto
added `flash_mode_status` value to `jca_settings.proto`
add `flash_mode_status` to `CameraAppSettings`
add `updateFlashModeStatus` to `LocalSettingsRepository`

**Changes connecting cameraUseCase to settings values**
_PreviewViewModel.kt_ & _PreviewUiState.kt_

- added settingsRepository as parameter for PreviewViewModel (read-only) 
- add currentCameraSettings value to PreviewUiState. ViewModel will provide the value of the settingsRepository as the initial value of currentCameraSettings. This can be regarded as our "Quick Settings" values
- add `setFlashMode` function. Changes the value of the flash mode in currentCameraSettings and then calls cameraUseCase.setFlashMode to rebind the use cases with the updated setting.

_CameraXCameraUseCase.kt_

- `setFlashMode` function takes a FlashModeStatus enum and applies it to the imageCaptureUseCase
- adds currentCameraSettings parameter to initialize. when first opening the app this will ensure the default stored flash mode is applied

**UI functionality**
- `setFlashMode` in _SettingsViewmodel_ to update datastore value 
- Flash Setting composable component

### DIFF
--- a/data/settings/src/main/java/com/google/jetpackcamera/settings/LocalSettingsRepository.kt
+++ b/data/settings/src/main/java/com/google/jetpackcamera/settings/LocalSettingsRepository.kt
@@ -18,7 +18,7 @@ package com.google.jetpackcamera.settings
 
 import androidx.datastore.core.DataStore
 import com.google.jetpackcamera.settings.model.DarkModeStatus
-import com.google.jetpackcamera.settings.model.Settings
+import com.google.jetpackcamera.settings.model.CameraAppSettings
 import kotlinx.coroutines.flow.first
 import kotlinx.coroutines.flow.map
 import javax.inject.Inject
@@ -30,9 +30,9 @@ class LocalSettingsRepository @Inject constructor(
     private val jcaSettings: DataStore<JcaSettings>
 ) : SettingsRepository {
 
-    override val settings = jcaSettings.data
+    override val cameraAppSettings = jcaSettings.data
         .map {
-            Settings(
+            CameraAppSettings(
                 default_front_camera = it.defaultFrontCamera,
                 dark_mode_status = when (it.darkModeStatus){
                     DarkModeProto.DARK_MODE_DARK -> DarkModeStatus.DARK
@@ -61,7 +61,7 @@ class LocalSettingsRepository @Inject constructor(
         }
     }
 
-    override suspend fun getSettings(): Settings {
-        return settings.first()
+    override suspend fun getSettings(): CameraAppSettings {
+        return cameraAppSettings.first()
     }
 }
--- a/data/settings/src/main/java/com/google/jetpackcamera/settings/LocalSettingsRepository.kt
+++ b/data/settings/src/main/java/com/google/jetpackcamera/settings/LocalSettingsRepository.kt
@@ -19,6 +19,7 @@ package com.google.jetpackcamera.settings
 import androidx.datastore.core.DataStore
 import com.google.jetpackcamera.settings.model.DarkModeStatus
 import com.google.jetpackcamera.settings.model.CameraAppSettings
+import com.google.jetpackcamera.settings.model.FlashModeStatus
 import kotlinx.coroutines.flow.first
 import kotlinx.coroutines.flow.map
 import javax.inject.Inject
@@ -40,6 +41,13 @@ class LocalSettingsRepository @Inject constructor(
                     DarkModeProto.DARK_MODE_SYSTEM,
                     DarkModeProto.UNRECOGNIZED,
                     null  -> DarkModeStatus.SYSTEM
+                },
+                flash_mode_status = when (it.flashModeStatus){
+                    FlashModeProto.FLASH_MODE_AUTO -> FlashModeStatus.AUTO
+                    FlashModeProto.FLASH_MODE_ON -> FlashModeStatus.ON
+                    FlashModeProto.FLASH_MODE_OFF,
+                    FlashModeProto.UNRECOGNIZED,
+                    null -> FlashModeStatus.OFF
                 }
             )
         }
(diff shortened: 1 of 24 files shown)


data/settings/src/main/java/com/google/jetpackcamera/settings/LocalSettingsRepository.kt
```diff
@@ -61,7 +69,18 @@ class LocalSettingsRepository @Inject constructor(
...
+            FlashModeStatus.AUTO -> FlashModeProto.FLASH_MODE_AUTO
+                FlashModeStatus.ON -> FlashModeProto.FLASH_MODE_ON
+                FlashModeStatus.OFF -> FlashModeProto.FLASH_MODE_OFF
+        }
+        jcaSettings.updateData {
+            it.copy { this.flashModeStatus = newStatus }
+        }
+    }
+
+    override suspend fun getSettings(): CameraAppSettings {
```
jsaund, line 83: nit: you can use expression function here since it's a simple expression
https://kotlinlang.org/docs/functions.html#single-expression-functions

`override suspend fun getSettings(): CameraAppSettings = cameraAppSettings.first()`

data/settings/src/main/java/com/google/jetpackcamera/settings/model/CameraAppSettings.kt
```diff
@@ -19,16 +19,17 @@ package com.google.jetpackcamera.settings.model
...
     val default_front_camera : Boolean,
     val dark_mode_status : DarkModeStatus,
-    val field : String = "todo"
+    val flash_mode_status : FlashModeStatus
 )
 
 
-fun getDefaultSettings(): Settings{
-    return Settings(
+fun getDefaultSettings(): CameraAppSettings{
```
jsaund, line 29: nit / optional 
You could also provide default values in the data class and create a constant `DEFAULT_CAMERA_APP_SETTINGS`

```
data class CameraAppSettings(
  val default_front_camera: Boolean = false,
  val dark_mode_status: DarkModeStatus = DarkModeStatus.SYSTEM,
  val flash_mode_sttatus: FlashModeStatus = FlashModeStatus.OFF
)

const val DEFAULT_CAMERA_APP_SETTINGS = CameraAppSettings()
```

domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
```diff
@@ -147,6 +153,16 @@ class CameraXCameraUseCase @Inject constructor(
...
     }
 
+     override fun setFlashMode(flashModeStatus: FlashModeStatus){
+        val newFlashMode = when (flashModeStatus) {
+            FlashModeStatus.OFF -> ImageCapture.FLASH_MODE_OFF // 2
+            FlashModeStatus.ON -> ImageCapture.FLASH_MODE_ON // 1
+            FlashModeStatus.AUTO -> ImageCapture.FLASH_MODE_AUTO // 0
+        }
+        imageCaptureUseCase.flashMode = newFlashMode
+        Log.d(TAG, "Set flash mode to: " + imageCaptureUseCase.flashMode)
```
jsaund, line 163: nit: prefer using string templates rather than string concatenation or string format here
https://kotlinlang.org/docs/strings.html#string-templates

```
Log.d(TAG, "Set flash mode to: ${imageCaptureUseCase.flashMode}")
```

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewViewModel.kt
```diff
@@ -84,6 +98,22 @@ class PreviewViewModel @Inject constructor(
...
+                previewUiState.value.copy(
+                    currentCameraSettings =
+                        previewUiState.value.currentCameraSettings.copy(
+                            flash_mode_status = flashModeStatus
+                        )
+                )
+            )
+        }
+        // apply to cameraUseCase
+        cameraUseCase.setFlashMode(previewUiState.value.currentCameraSettings.flash_mode_status)
```
jsaund, line 114: Since the flash mode is being read from the preview ui state, it's possible to be reading a stale state.
The `launch` operation may not trigger immediately since it's starting a coroutine. So that means the coroutine has to launch and then the preview ui state has to be updated with the new value. But this line might be called before all that happens and so the value being read here might be the previous value.

One option is to invoke the `cameraUseCase.setFlashMode(flashModeStatus)` first and then call viewModeScope.launch
Or perform everything inside the viewmodelscope

domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
```diff
@@ -147,6 +153,16 @@ class CameraXCameraUseCase @Inject constructor(
         recording?.stop()
     }
 
+     override fun setFlashMode(flashModeStatus: FlashModeStatus){
+        val newFlashMode = when (flashModeStatus) {
```
jsaund, line 157: nit: you can inline the assignment here: `imageCaptureUseCase.flashMode = when (flashModeStatus) { ... } `

domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
jsaund, line 163: nit: prefer string template
`Log.d(TAG, "Set flash mode to: ${imageCaptureUseCase.flashMode}")

PR #21, opened 2023-07-11
connect quick settings

Kimblebee:
So I've connected quick settings to be functional and refactored the setup of the quicksettings screen. 

* Renamed `QuickSettingsComponents` to `QuickSettingsScreen`. 
* moved `isOpen` outside of `QuickSettingsScreen`.  I put it into the previewViewModel, but it can also just be put into the preview screen, or anywhere outside of the QuickSettings Screen. With it like this, other elements in Preview will have access to whether or not the state of quick settings is open, and we can add conditions to change their functionality/visibility accordingly.
* reduced amount of parameters for quick settings/expanded components. any new parameters should just be callback functions for settings to use. 
* Deconstructed different ui elements and ui organization logic into their own independent components/subcomponents.
* Created new `../ui/QuickSettingsComponents.kt` to place all components/subcomponents to be used in the QuickSettingsScreen
* Delete QuickSettingsUiModel. I moved responsibility of self-awareness to each individual component as you create them


feature/quicksettings/src/main/java/com/google/jetpackcamera/feature/quicksettings/ui/QuickSettingsComponents.kt
```diff
@@ -0,0 +1,225 @@
...
+                FlashModeStatus.OFF -> onClick(FlashModeStatus.ON)
+                FlashModeStatus.ON -> onClick(FlashModeStatus.AUTO)
+                FlashModeStatus.AUTO -> onClick(FlashModeStatus.OFF)
+            }
+        }
+    )
+}
+
+@Composable
+fun QuickFlipCamera(flipCamera: (Boolean) -> Unit, currentFacingFront: Boolean) {
```
davidjiagoogle, line 108: I think it's better we use enum instead of boolean here. There can potentially be more than two cameras in the future if we take into account external camera or concurrent cameras

feature/quicksettings/src/main/java/com/google/jetpackcamera/feature/quicksettings/ui/QuickSettingsComponents.kt
```diff
@@ -0,0 +1,225 @@
...
+        { QuickSetRatio(onClick = { setRatio() }, ratio = 1, currentRatio = currentRatio) },
+        { QuickSetRatio(onClick = { setRatio() }, ratio = 2, currentRatio = currentRatio) },
+        { QuickSetRatio(onClick = { setRatio() }, ratio = 3, currentRatio = currentRatio) }
+
+    )
+    ExpandedQuickSetting(quickSettingButtons = buttons)
+}
+
+@Composable
+fun QuickSetRatio(onClick: () -> Unit, ratio: Int, currentRatio: Int) {
```
davidjiagoogle, line 69: we should discuss about QuickSettingUiModel vs. QuickSet***Components. The reason I did it that way is so that we wouldn't have to duplicate code and we can keep track of all the enums and their relationship to the component in a single file. But I see merit doing this way too

feature/quicksettings/src/main/java/com/google/jetpackcamera/feature/quicksettings/QuickSettingsScreen.kt
```diff
@@ -0,0 +1,178 @@
...
+    currentCameraSettings: CameraAppSettings,
+    onLensFaceClick: (lensFacingFront: Boolean) -> Unit,
+    onFlashModeClick: (flashMode: FlashModeStatus) -> Unit,
+    shouldShowQuickSetting: IsExpandedQuickSetting,
+    setVisibleQuickSetting: (IsExpandedQuickSetting) -> Unit
+    // onAspectRatioClick: (aspectRatio: CameraAspectRatio) -> Unit,
+    // onTimerClick: (timer: CameraTimer) -> Unit,
+) {
+
+    Column(
```
davidjiagoogle, line 137: There used to be a width calculation algorithm here to make sure we fit the appropriate number of components in each row according to the screen width. I think we should use that instead of hardcoding the item to 3. This would be especially relevant in landscape mode if we ever support that

feature/quicksettings/src/main/java/com/google/jetpackcamera/feature/quicksettings/ui/QuickSettingsComponents.kt
```diff
@@ -0,0 +1,225 @@
...
+
+// completed components
+
+@Composable
+fun ExpandedQuickSetRatio(
+    setRatio: (/*TODO: insert ratio function*/) -> Unit,
+    currentRatio: Int
+) {
+    val buttons: Array<@Composable () -> Unit> = arrayOf(
+        { QuickSetRatio(onClick = { setRatio() }, ratio = 1, currentRatio = currentRatio) },
```
davidjiagoogle, line 60: Make sure to leave a TODO comment for the placeholders

davidjiagoogle: Same goes for all the ones below to

davidjiagoogle (approved):

feature/quicksettings/src/main/java/com/google/jetpackcamera/feature/quicksettings/QuickSettingsScreen.kt
Kimblebee: I moved it into the `QuickSettingsGrid` component in `QuickSettingsComponents.kt`

feature/quicksettings/src/main/java/com/google/jetpackcamera/feature/quicksettings/ui/QuickSettingsComponents.kt
Kimblebee: yeah like we had spoken about earlier, I think this approach would simplify the process of creating/removing quick settings, and improve readability for anyone new looking at the code

feature/quicksettings/src/main/java/com/google/jetpackcamera/feature/quicksettings/ui/QuickSettingsComponents.kt
Kimblebee: There could be multiple cameras to select from, but this is is to change which direction we are facing, which could either be front or back

feature/quicksettings/src/main/java/com/google/jetpackcamera/feature/quicksettings/ui/QuickSettingsComponents.kt
Kimblebee: I will also write up some bugs for these

PR #22, opened 2023-07-12
fix bug that doesnt let you zoom after flipping camera

Kimblebee:
teeny tiny change :)
self explanatory. 
after flipping the camera using _a quick setting_, the camera var in `CameraXCameraUseCase` isn't updated to acknowledge that. now it does!


jsaund (approved):
good catch!

PR #27, opened 2023-07-25
implement tap to focus

Kimblebee:
In the `PreviewScreen`, I created a value that keeps track of the View information from the CameraPreview composable. I just pass down a callback function from CameraPreview -> PreviewSurface -> CombinedSurface -> Texture to set that value.

When the screen is tapped, then the display, height, and width information of the view is provided to `PreviewViewModel`'s tapToFocus(), which is then provides that info to `cameraUseCase.tapToFocus` to create a metering point and action to focus.

### DIFF
--- a/camera-viewfinder-compose/src/main/java/com/google/jetpackcamera/viewfinder/surface/Texture.kt
+++ b/camera-viewfinder-compose/src/main/java/com/google/jetpackcamera/viewfinder/surface/Texture.kt
@@ -93,7 +93,7 @@ fun Texture(
             }
         }, update = {
             textureView = it
-            setView(textureView!!)
+            setView(it)
             onRequestBitmapReady { -> it.bitmap }
         }
     )
--- a/camera-viewfinder-compose/src/main/java/com/google/jetpackcamera/viewfinder/surface/Texture.kt
+++ b/camera-viewfinder-compose/src/main/java/com/google/jetpackcamera/viewfinder/surface/Texture.kt
@@ -20,6 +20,7 @@ import android.graphics.Bitmap
 import android.graphics.SurfaceTexture
 import android.util.Log
 import android.view.TextureView
+import android.view.View
 import android.view.ViewGroup
 import android.widget.LinearLayout
 import androidx.compose.runtime.Composable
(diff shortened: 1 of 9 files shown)


domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
```diff
@@ -189,6 +192,29 @@ class CameraXCameraUseCase @Inject constructor(
...
+        x: Float,
+        y: Float
+    ) {
+        if (camera != null) {
+            val meteringPoint = DisplayOrientedMeteringPointFactory(
+                display,
+                camera!!.cameraInfo,
+                surfaceWidth.toFloat(),
+                surfaceHeight.toFloat()
+            )
```
yasith, line 208: nit: formatting

camera-viewfinder-compose/src/main/java/com/google/jetpackcamera/viewfinder/surface/Texture.kt
```diff
@@ -91,9 +93,11 @@ fun Texture(
             }
         }, update = {
             textureView = it
+            setView(textureView!!)
```
yasith, line 96: nit, you can use `it` here instead of `textureview!!`

domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
```diff
@@ -189,6 +192,29 @@ class CameraXCameraUseCase @Inject constructor(
...
+        display: Display,
+        surfaceWidth: Int,
+        surfaceHeight: Int,
+        x: Float,
+        y: Float
+    ) {
+        if (camera != null) {
+            val meteringPoint = DisplayOrientedMeteringPointFactory(
+                display,
+                camera!!.cameraInfo,
```
yasith, line 205: avoid `!!` here, you can just do `camera.cameraInfo` since camera is not null here

yasith (approved):
Only a few nits

domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
Kimblebee: could not do that, the error message read: `Smart cast to 'Camera' is impossible, because 'camera' is a mutable property that could have been changed by this time`

PR #29, opened 2023-08-02
Lock the app to portrait orientation

yasith:
Also removing redundant `android:label` (already defined at app level)


Kimblebee (approved):
LGTM :-)

PR #31, opened 2023-08-22
Refactor Preview UI

Kimblebee:
Refactored UI for better appearance, readability, and maintainability; _**There is no change in functionality and minimal change in appearance.**_ 

Here's a brief summary of the changes in this PR:

* Compartmentalized the logic that already exists within the `/preview/PreviewScreen.kt`, and moved most of it into `/preview/ui/PreviewScreenComponents.kt`. Big adjustments on use of containers and modifiers.

* Adjusted some modifiers for `QuickSettingsScreen.kt`; 

* I added a "flip camera" button to the preview screen for easier access

* I also added minor styling changes to the Settings screen; added colors and adjusted some spacing.

### DIFF
--- a/feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewScreen.kt
+++ b/feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewScreen.kt
@@ -19,23 +19,13 @@ package com.google.jetpackcamera.feature.preview
 import android.os.Handler
 import android.os.Looper
 import android.util.Log
-import android.view.View
 import androidx.camera.core.Preview.SurfaceProvider
-import androidx.compose.animation.core.animateFloatAsState
-import androidx.compose.animation.core.tween
-import androidx.compose.foundation.Canvas
-import androidx.compose.foundation.background
-import androidx.compose.foundation.border
-import androidx.compose.foundation.gestures.detectTapGestures
 import androidx.compose.foundation.gestures.rememberTransformableState
-import androidx.compose.foundation.gestures.transformable
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Box
-import androidx.compose.foundation.layout.BoxWithConstraints
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.fillMaxSize
-import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.size
 import androidx.compose.foundation.layout.width
@@ -59,18 +49,18 @@ import androidx.compose.runtime.setValue
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.ExperimentalComposeUiApi
 import androidx.compose.ui.Modifier
-import androidx.compose.ui.draw.alpha
-import androidx.compose.ui.graphics.Color
-import androidx.compose.ui.input.pointer.pointerInput
 import androidx.compose.ui.platform.LocalLifecycleOwner
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.unit.dp
-import androidx.compose.ui.unit.sp
 import androidx.hilt.navigation.compose.hiltViewModel
 import androidx.lifecycle.Lifecycle
 import androidx.lifecycle.repeatOnLifecycle
+import com.google.jetpackcamera.feature.preview.ui.CaptureButton
+import com.google.jetpackcamera.feature.preview.ui.FlipCameraButton
+import com.google.jetpackcamera.feature.preview.ui.PreviewDisplay
+import com.google.jetpackcamera.feature.preview.ui.SettingsNavButton
+import com.google.jetpackcamera.feature.preview.ui.ZoomScaleText
 import com.google.jetpackcamera.feature.quicksettings.QuickSettingsScreen
-import com.google.jetpackcamera.viewfinder.CameraPreview
 import kotlinx.coroutines.CompletableDeferred
 import kotlinx.coroutines.awaitCancellation
 
@@ -79,8 +69,6 @@ private const val TAG = "PreviewScreen"
 /**
  * Screen used for the Preview feature.
  */
-@OptIn(ExperimentalComposeUiApi::class)
-@ExperimentalMaterial3Api
 @Composable
 fun PreviewScreen(
     onNavigateToSettings: () -> Unit,
@@ -93,19 +81,19 @@ fun PreviewScreen(
     val lifecycleOwner = LocalLifecycleOwner.current
 
     val deferredSurfaceProvider = remember { CompletableDeferred<SurfaceProvider>() }
-    val onSurfaceProviderReady: (SurfaceProvider) -> Unit = {
-        Log.d(TAG, "onSurfaceProviderReady")
-        deferredSurfaceProvider.complete(it)
-    }
-    lateinit var viewInfo: View
+
     var zoomScale by remember { mutableStateOf(1f) }
+
     var zoomScaleShow by remember { mutableStateOf(false) }
+
     val zoomHandler = Handler(Looper.getMainLooper())
-    val transformableState = rememberTransformableState(onTransformation = { zoomChange, _, _ ->
-        zoomScale = viewModel.setZoomScale(zoomChange)
-        zoomScaleShow = true
-        zoomHandler.postDelayed({ zoomScaleShow = false }, 3000)
-    })
+
+    val transformableState = rememberTransformableState(
+        onTransformation = { zoomChange, _, _ ->
+            zoomScale = viewModel.setZoomScale(zoomChange)
+            zoomScaleShow = true
+            zoomHandler.postDelayed({ zoomScaleShow = false }, 3000)
+        })
 
     LaunchedEffect(lifecycleOwner) {
         val surfaceProvider = deferredSurfaceProvider.await()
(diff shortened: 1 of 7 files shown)


feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
```diff
@@ -0,0 +1,223 @@
...
+            imageVector = Icons.Filled.Settings,
+            tint = Color.White,
+            contentDescription = stringResource(R.string.settings_content_description),
+            modifier = Modifier.size(72.dp)
+        )
+    }
+}
+
+@Composable
+fun ZoomScaleText(zoomScale: Float, show: Boolean) {
```
yasith, line 175: Instead of passing show here, if show is false, lets skip adding this composable.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
```diff
@@ -0,0 +1,223 @@
...
+    val contentAlpha = animateFloatAsState(
+        targetValue = if (show) 1f else 0f, label = "zoomScaleAlphaAnimation",
+        animationSpec = tween()
+    )
+    Text(
+        modifier = Modifier.alpha(contentAlpha.value),
+        text = String.format("%.1fx", zoomScale),
```
yasith, line 182: nit: `"%.1fx".format(zoomScale)`

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
```diff
@@ -0,0 +1,223 @@
...
+        fontSize = 20.sp,
+        color = Color.White
+    )
+}
+
+@Composable
+fun CaptureButton(
+    modifier: Modifier = Modifier,
```
yasith, line 190: nit: when we pass in modifier to a composable, it should be the first optional parameter, but comes after non-optional parameters.

In this case, we should think about if we need to pass in a modifier for this, since we're not planning on re-using this element with modifier changes that affect the element internally.

feature/quicksettings/src/main/java/com/google/jetpackcamera/feature/quicksettings/QuickSettingsScreen.kt
```diff
@@ -176,6 +170,6 @@ private fun ExpandedQuickSettingsUi(
                 )
             }
         }
-        Spacer(modifier = Modifier.height(dimensionResource(id = R.dimen.quick_settings_spacer_height)))
+        //Spacer(modifier = Modifier.height(dimensionResource(id = R.dimen.quick_settings_spacer_height)))
```
yasith, line 173: remove?

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
Kimblebee: that's true in this case. we don't need the modifier here.

however if we may want to change the attributes of this (or any) composable's placement with respect to its parent container, we will need to designate those modifier values and pass it in.

 It also makes it easier to change the spacing/padding attributes up here vs sifting through the components file every time.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewScreen.kt
```diff
@@ -210,67 +142,56 @@ fun PreviewScreen(
...
+                        //enable only when phone has front and rear camera
+                        enabledCondition =
+                        previewUiState.currentCameraSettings.back_camera_available
+                                && previewUiState.currentCameraSettings.front_camera_available
+                    )
+                    /*todo: close quick settings on start record/image capture*/
+                    CaptureButton(
+                        modifier = Modifier
+                            .fillMaxHeight(),
+                        onClick = { onImageCapture(viewModel) },
```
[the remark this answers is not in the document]
Kimblebee: I put these wrappers here because I wanted to add extra behaviors when performing these actions without bloating this area of the code (not reflected in *this* PR).

i.e. if quick settings screen is open, close it when taking a picture or starting recording. 

or if we wanted to play a sound blip when taking a photo/starting a recording, etc.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewScreen.kt
```diff
@@ -113,71 +98,24 @@ fun PreviewScreen(
...
-                            e.printStackTrace()
-                        }
-                    }
-                )
-            },
-
-            contentAlignment = Alignment.Center
+        // display camera feed. this stays behind everything else
+        PreviewDisplay(
+            viewModel = viewModel,
```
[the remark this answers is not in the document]
Kimblebee: ok, i'll use callbacks for the viewModel functions instead

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
```diff
@@ -0,0 +1,223 @@
...
+    }
+}
+
+@Composable
+fun FlipCameraButton(
+    modifier: Modifier = Modifier,
+    enabledCondition: Boolean,
+    onClick: () -> Unit
+) {
+    Box(modifier = modifier) {
```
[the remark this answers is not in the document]
Kimblebee: the box acts as a spacer for the elements occupying the same row as the capture button.

Since the capture button is centered relative to this box, it will maintain the same relative position regardless of the device it is viewed on

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewScreen.kt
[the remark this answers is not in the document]
yasith: These side effects should happen in response to the changes from the UiState

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
yasith: Sounds good, lets keep the modifiers.

We should make them the first optional parameter.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewScreen.kt
```diff
@@ -194,106 +130,73 @@ fun PreviewScreen(
...
+                                && previewUiState.currentCameraSettings.front_camera_available
+                    )
+                    /*todo: close quick settings on start record/image capture*/
+                    CaptureButton(
+                        onClick = { viewModel.captureImage() },
+                        onLongPress = { viewModel.startVideoRecording() },
+                        onRelease = { viewModel.stopVideoRecording() },
+                        state = previewUiState.videoRecordingState
+                    )
+                    /* spacer is a placeholder to maintain the proportionate location of this row of
```
yasith, line 189: nit: formatting for the comment

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewScreen.kt
```diff
@@ -194,106 +130,73 @@ fun PreviewScreen(
...
-                            onRelease = { viewModel.stopVideoRecording() },
-                            state = previewUiState.videoRecordingState
-                        )
-                    }
+                    FlipCameraButton(
+                        modifier = Modifier
+                            .weight(1f)
+                            .fillMaxHeight(),
+                        onClick = { viewModel.flipCamera() },
+                        //enable only when phone has front and rear camera
```
yasith, line 177: nit: formatting

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewScreen.kt
```diff
@@ -194,106 +130,73 @@ fun PreviewScreen(
...
+                        enabledCondition =
+                        previewUiState.currentCameraSettings.back_camera_available
+                                && previewUiState.currentCameraSettings.front_camera_available
+                    )
+                    /*todo: close quick settings on start record/image capture*/
```
yasith, line 182: nit: formatting

feature/settings/src/main/java/com/google/jetpackcamera/settings/ui/SettingsComponents.kt
```diff
@@ -170,10 +172,11 @@ fun BasicPopupSetting(
 ) {
     val popupStatus = remember { mutableStateOf(false) }
     SettingUI(
+        modifier = Modifier.clickable() { popupStatus.value = true },
         title = title,
         description = description,
         leadingIcon = leadingIcon,
-        onClick = { popupStatus.value = true },
+        //onClick = { popupStatus.value = true },
```
yasith, line 179: nit: remove if not needed

feature/settings/src/main/java/com/google/jetpackcamera/settings/ui/SettingsComponents.kt
```diff
@@ -223,43 +238,25 @@ fun SwitchSettingUI(
...
 fun SettingUI(
+    modifier: Modifier = Modifier,
     title: String,
     description: String? = null,
     leadingIcon: @Composable (() -> Unit)?,
     trailingContent: @Composable (() -> Unit)?,
-    onClick: () -> Unit,
-    enabled: Boolean = true
+    //onClick: () -> Unit,
+    //enabled: Boolean = true
```
yasith, line 247: nit: remove if not needed

feature/settings/src/main/java/com/google/jetpackcamera/settings/ui/SettingsComponents.kt
```diff
@@ -223,43 +238,25 @@ fun SwitchSettingUI(
 @OptIn(ExperimentalMaterial3Api::class)
 @Composable
 fun SettingUI(
+    modifier: Modifier = Modifier,
```
yasith, line 241: make the modifier the first optional argument

yasith (changes requested):

feature/quicksettings/src/main/java/com/google/jetpackcamera/feature/quicksettings/QuickSettingsScreen.kt
```diff
@@ -76,44 +73,43 @@ fun QuickSettingsScreen(
...
+                .fillMaxSize()
+                .background(color = backgroundColor.value)
+                .alpha(alpha = contentAlpha.value)
+                .clickable {
+                    // if a setting is expanded, click on the background to close it.
+                    // if no other settings are expanded, then close the popup
+                    if (shouldShowQuickSetting == IsExpandedQuickSetting.NONE)
+                        toggleIsOpen()
+                    else
+                        shouldShowQuickSetting = IsExpandedQuickSetting.NONE
```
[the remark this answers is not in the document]
Kimblebee: I was using if,else because this part only needs to know if a quicksetting is expanded or not, that way I wouldnt have to keep adding the enum names to essentially the "else" option.

Also I don't think I understand what you mean by using the enum for 2 different things, at least the way currently written.
The way the `shouldShowQuickSetting` works is that when a quick setting we have designated as expandable has an `onClick()` that will change the value of `shouldShowQuickSetting` to display the respective choices.
My intention is that clicking on the dark background should backstep level whatever is currently on the quick settings screen. if a setting is expanded, go back to the whole quick settings menu. if it isn't, then close out of the entire menu.

 I think I could rename the enum if it makes that a bit confusing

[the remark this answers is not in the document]
yasith: Okay, I think I get it now. Would it be more appropriate to call it "currentExpandedQuickSetting"? Or something like that.

Don't have to address that in this PR

yasith (approved):

PR #32, opened 2023-08-23
PostCapture screen implementation with in memory Image storage

yasith:
TODOs for later
- Use image orientation from the transform data of ImageProxy
- Scale image to fit screen width/height
- Overlay buttons for saving/discarding
- Async listening for image to be ready (might be needed for lower end devices)


data/storage/src/main/java/com/google/jetpackcamera/storage/ImageCache.kt
```diff
@@ -0,0 +1,35 @@
...
+
+package com.google.jetpackcamera.storage
+
+import android.graphics.Bitmap
+import android.media.Image
+
+/**
+ * Used to store a single [Image].
+ */
+interface ImageCache {
```
jsaund, line 25: nit: You can replace the set/get with a var

```
interface ImageCache {
    var image: Bitmap?
}

class ImageCacheImpl : ImageCache {
    override var image: Bitmap?
        get() = TODO("Not yet implemented")
        set(value) {}
}
```

jsaund (commented):
I wonder if it makes it cleaner / simpler to rely on a image loading library to manage all this for you.
So if you're using Coil for example, you will want to take the JPEG byte array and save that somewhere
Then in you're Composable you can use AsyncImage and in it's ImageRequest use `.data(jpegByteArray)`

You can then tie to a custom fetcher (I got this from https://github.com/coil-kt/coil/issues/171#issuecomment-554109785)
```
class ByteArrayFetcher : Fetcher<ByteArray> {

    override fun key(data: ByteArray): String? = null

    override suspend fun fetch(
        pool: BitmapPool,
        data: ByteArray,
        size: Size,
        options: Options
    ): FetchResult {
        return SourceResult(
            source = ByteArrayInputStream(data).source().buffer(),
            mimeType = null,
            dataSource = DataSource.MEMORY
        )
    }
}
```
Register the new fetcher -- add to ComponentRegistry when creating the ImageLoader

Now you should get more efficient bitmap pool management as well, resizing, rotation handling

Kimblebee:
There are some issues with orientation of the image captured. front-facing camera is mirrored, and the rear-facing camera is upside-down

feature/postcapture/build.gradle
```diff
@@ -0,0 +1,86 @@
...
+
+    // Compose - Android Studio Preview support
+    implementation 'androidx.compose.ui:ui-tooling-preview'
+    debugImplementation 'androidx.compose.ui:ui-tooling'
+
+    // Compose - Integration with ViewModels with Navigation and Hilt
+    implementation("androidx.hilt:hilt-navigation-compose:1.0.0")
+
+    // Compose - Testing
+    androidTestImplementation "androidx.compose.ui:ui-test-junit4:$compose_version"
```
davidjiagoogle, line 65: where does compose_version come from? Other packages just use:
androidTestImplementation("androidx.compose.ui:ui-test-junit4")

PR #35, opened 2023-09-01
Migrate build configuration from Groovy to Kotlin

yasith:


camera-viewfinder-compose/build.gradle.kts
```diff
@@ -0,0 +1,71 @@
...
+    id("org.jetbrains.kotlin.android")
+}
+
+android {
+    namespace = "com.google.jetpackcamera.camerax"
+    compileSdk = 33
+
+    defaultConfig {
+        minSdk = 21
+        targetSdk = 33
```
jsaund, line 12: I'm assuming these will reference vars in a follow up PR?

jsaund (approved):
Thanks for doing this! This will really help keep consistency with deps and make it easier too

camera-viewfinder-compose/build.gradle.kts
yasith: Yup, sorry should've added that to the description.

Follow up PRs for

- project-wide values
- Moving to version catalogs
- Potentially using convention plugins: https://docs.gradle.org/current/samples/sample_convention_plugins.html

PR #36, opened 2023-09-07
Use mutableStateOf setter instead of CompletableDeferred for SurfaceProvider

yasith:
We received feedback from DevRel that `CompletableDeferred` is unfamiliar. I'm exploring another way of handling the async availability `SurfaceProvider` in this PR.

In the previous implementation 
- We use a `CompletableDeferred<SurfaceProvider>` that's completed when `onSurfaceProviderReady()` is invoked. 
- We await the completion of the `CompletableDeferred`  in order to launch the `repeatOnLifecycle` coroutine.

New proposed implementation
- Store `SurfaceProvider` as a state and update that when the `SurfaceProvider` is ready
- Modify the key of `LaunchedEffect` to include `surfaceProvider` and only launch `repeatOnLifecycle` if `surfaceProvider != null`

### DIFF
--- a/feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewScreen.kt
+++ b/feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewScreen.kt
@@ -42,8 +42,6 @@ import androidx.compose.foundation.layout.width
 import androidx.compose.foundation.shape.CircleShape
 import androidx.compose.material.icons.Icons
 import androidx.compose.material.icons.filled.Settings
-import androidx.compose.material3.Button
-import androidx.compose.material3.ChipColors
 import androidx.compose.material3.ExperimentalMaterial3Api
 import androidx.compose.material3.Icon
 import androidx.compose.material3.IconButton
@@ -71,7 +69,6 @@ import androidx.lifecycle.Lifecycle
 import androidx.lifecycle.repeatOnLifecycle
 import com.google.jetpackcamera.feature.quicksettings.QuickSettingsScreen
 import com.google.jetpackcamera.viewfinder.CameraPreview
-import kotlinx.coroutines.CompletableDeferred
 import kotlinx.coroutines.awaitCancellation
 
 private const val TAG = "PreviewScreen"
@@ -92,11 +89,9 @@ fun PreviewScreen(
 
     val lifecycleOwner = LocalLifecycleOwner.current
 
-    val deferredSurfaceProvider = remember { CompletableDeferred<SurfaceProvider>() }
-    val onSurfaceProviderReady: (SurfaceProvider) -> Unit = {
-        Log.d(TAG, "onSurfaceProviderReady")
-        deferredSurfaceProvider.complete(it)
-    }
+    val (surfaceProvider, onSurfaceProviderReady) =
+        remember { mutableStateOf<SurfaceProvider?>(null) }
+
     lateinit var viewInfo: View
     var zoomScale by remember { mutableStateOf(1f) }
     var zoomScaleShow by remember { mutableStateOf(false) }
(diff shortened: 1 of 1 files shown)


feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewScreen.kt
```diff
@@ -92,11 +89,9 @@ fun PreviewScreen(
 
     val lifecycleOwner = LocalLifecycleOwner.current
 
-    val deferredSurfaceProvider = remember { CompletableDeferred<SurfaceProvider>() }
-    val onSurfaceProviderReady: (SurfaceProvider) -> Unit = {
-        Log.d(TAG, "onSurfaceProviderReady")
-        deferredSurfaceProvider.complete(it)
-    }
+    val (surfaceProvider, onSurfaceProviderReady) =
```
JolandaVerhoef, line 92: Consider renaming this to:
`val surfaceProvider by remember { mutableStateOf<SurfaceProvider?>(null) }`

And then when you call the CameraPreview composable, you use:
```
CameraPreview(
                        ..
                        onSurfaceProviderReady = { surfaceProvider = it },
                        ..
                    )
```

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewScreen.kt
```diff
@@ -107,14 +102,15 @@ fun PreviewScreen(
...
-            } finally {
-                viewModel.stopCamera()
+    LaunchedEffect(lifecycleOwner, surfaceProvider) {
+        if (surfaceProvider != null) {
+            lifecycleOwner.repeatOnLifecycle(Lifecycle.State.STARTED) {
+                viewModel.runCamera(surfaceProvider)
+                try {
+                    awaitCancellation()
+                } finally {
+                    viewModel.stopCamera()
```
JolandaVerhoef, line 112: If I understand this snippet correctly, this means you'd only stop the camera in the on-destroy lifecycle event. Is that okay, or would it need to be stopped during the on-stop lifecycle event?

yasith: We will stop the camera in onStop, `repeatOnLifecycle(Lifecycle.State.STARTED)` launches a new coroutine in the block that will be cancelled when lifecycle state is in STOPPED.

Based this off of: https://developer.android.com/topic/libraries/architecture/coroutines#restart

JolandaVerhoef: Ah yes, you're right. I got confused when reading the documentation on `withStarted` which mentions onDestroy instead of onStop:

```
Run block with this LifecycleOwner's Lifecycle in a Lifecycle.State of at least Lifecycle.State.STARTED and resume with the result. Throws the CancellationException if the lifecycle has reached Lifecycle.State.DESTROYED by the time of the call or before block is able to run.
```

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewScreen.kt
JolandaVerhoef: If you want to use it like this, I'd recommend using the name `setSurfaceProvider`. Using `onSurfaceProviderReady` refers to behavior which is not part of the mutable state itself, rather the way you're using it.

jsaund (approved):
Looks good
We should try how this behaves without restricting to lifecycle state STARTED (maybe a few log statements to try behavior of bg/fg and show preview -> remove preivew and replace with image capture composable -> go back to preview

PR #37, opened 2023-09-20
Capture Mode Setting and Integration

Kimblebee:
Self explanatory. Most interesting changes to review is integration in `CameraXCameraUseCase`

* create enum representing capture type for camera
* create setting to store default value for capture type
* update CameraXCameraUseCase to use enum value

### DIFF
--- a/data/settings/src/main/java/com/google/jetpackcamera/settings/LocalSettingsRepository.kt
+++ b/data/settings/src/main/java/com/google/jetpackcamera/settings/LocalSettingsRepository.kt
@@ -18,7 +18,7 @@ package com.google.jetpackcamera.settings
 
 import androidx.datastore.core.DataStore
 import com.google.jetpackcamera.settings.model.CameraAppSettings
-import com.google.jetpackcamera.settings.model.CaptureModeStatus
+import com.google.jetpackcamera.settings.model.CaptureMode
 import com.google.jetpackcamera.settings.model.DarkModeStatus
 import com.google.jetpackcamera.settings.model.FlashModeStatus
 import kotlinx.coroutines.flow.first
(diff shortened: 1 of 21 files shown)


yasith (changes requested):

data/settings/src/main/java/com/google/jetpackcamera/settings/model/CaptureModeStatus.kt
```diff
@@ -0,0 +1,21 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.jetpackcamera.settings.model
+
+enum class CaptureModeStatus {
```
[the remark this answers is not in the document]
Kimblebee: I think thats ok, though we should be mindful of an auto generated class in the build folder (from proto-datastore) that has the same name, `CaptureMode`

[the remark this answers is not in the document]
yasith: I think it's better for all of them to have the same name, since they're all essentially referring to the same thing.

When we need to specify them by type in the same file, we can use an import as alias if needed.

data/settings/src/main/proto/com/google/jetpackcamera/settings/capture_mode.proto
```diff
@@ -20,6 +20,7 @@ option java_package = "com.google.jetpackcamera.settings";
 option java_multiple_files = true;
 
 enum CaptureModeProto {
-  CAMERA_MODE_DEFAULT = 0;
-  CAMERA_MODE_SINGLE_STREAM = 1;
+  CAMERA_MODE_UNDEFINED = 0;
```
yasith, line 23: These should be "CAPTURE_MODE" instead of "CAMERA_MODE"

data/settings/src/main/proto/com/google/jetpackcamera/settings/capture_mode.proto
```diff
@@ -20,6 +20,7 @@ option java_package = "com.google.jetpackcamera.settings";
 option java_multiple_files = true;
 
 enum CaptureModeProto {
```
yasith, line 22: We should rename this to `CaptureMode` protos shouldn't contain "Proto" in the name.

When we need to differentiate them where we use them, we can potentially use import as aliases.

data/settings/src/main/java/com/google/jetpackcamera/settings/LocalSettingsRepository.kt
```diff
@@ -51,11 +51,11 @@ class LocalSettingsRepository @Inject constructor(
...
                 captureMode = when (it.captureModeStatus) {
-                    CaptureModeProto.CAMERA_MODE_SINGLE_STREAM -> CaptureModeStatus.SINGLE_STREAM
-
-                    CaptureModeProto.CAMERA_MODE_DEFAULT,
+                    CaptureModeProto.CAMERA_MODE_SINGLE_STREAM -> CaptureMode.SINGLE_STREAM
+                    CaptureModeProto.CAMERA_MODE_MULTI_STREAM,
+                    CaptureModeProto.CAMERA_MODE_UNDEFINED,
                     CaptureModeProto.UNRECOGNIZED,
-                    null -> CaptureModeStatus.DEFAULT
+                    null -> CaptureMode.MULTI_STREAM
```
yasith, line 58: combine these 3 in an else branch

yasith (approved):

PR #38, opened 2023-09-20
Aspect Ratio Setting

Kimblebee:
Add set default aspect ratio to settings screen

### DIFF
--- a/data/settings/src/main/java/com/google/jetpackcamera/settings/LocalSettingsRepository.kt
+++ b/data/settings/src/main/java/com/google/jetpackcamera/settings/LocalSettingsRepository.kt
@@ -17,6 +17,7 @@
 package com.google.jetpackcamera.settings
 
 import androidx.datastore.core.DataStore
+import com.google.jetpackcamera.settings.model.AspectRatio
 import com.google.jetpackcamera.settings.model.CameraAppSettings
 import com.google.jetpackcamera.settings.model.DarkModeStatus
 import com.google.jetpackcamera.settings.model.FlashModeStatus
@@ -50,7 +51,17 @@ class LocalSettingsRepository @Inject constructor(
                     null -> FlashModeStatus.OFF
                 },
                 front_camera_available = it.frontCameraAvailable,
-                back_camera_available = it.backCameraAvailable
+                back_camera_available = it.backCameraAvailable,
+                aspect_ratio = when (it.aspectRatioStatus) {
+                    AspectRatioProto.ASPECT_RATIO_NINE_SIXTEEN -> AspectRatio.NINE_SIXTEEN
+                    AspectRatioProto.ASPECT_RATIO_ONE_ONE -> AspectRatio.ONE_ONE
+                    // defaults to 3:4 aspect ratio
+                    AspectRatioProto.ASPECT_RATIO_THREE_FOUR,
+                    AspectRatioProto.UNRECOGNIZED,
+                    null -> AspectRatio.THREE_FOUR
+
+
+                }
             )
         }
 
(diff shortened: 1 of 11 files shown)


data/settings/src/main/proto/com/google/jetpackcamera/settings/aspect_ratio.proto
```diff
@@ -0,0 +1,26 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+syntax = "proto3";
+
+option java_package = "com.google.jetpackcamera.settings";
+option java_multiple_files = true;
+
+enum AspectRatioProto {
```
yasith, line 22: Lets drop "Proto" from the name.

We should also rename the others later.

data/settings/src/main/proto/com/google/jetpackcamera/settings/aspect_ratio.proto
```diff
@@ -0,0 +1,26 @@
...
+  ASPECT_RATIO_THREE_FOUR = 0;
```
yasith, line 23: Google proto guidelines say we should reserve 0 for UNDEFINED.

Lets add
`ASPECT_RATIO_UNDEFINED = 0`

And move everything else down by one

data/settings/src/main/java/com/google/jetpackcamera/settings/LocalSettingsRepository.kt
```diff
@@ -105,4 +116,17 @@ class LocalSettingsRepository @Inject constructor(
...
+        val newStatus = when (aspectRatio) {
+            AspectRatio.NINE_SIXTEEN -> AspectRatioProto.ASPECT_RATIO_NINE_SIXTEEN
+            AspectRatio.THREE_FOUR -> AspectRatioProto.ASPECT_RATIO_THREE_FOUR
+            AspectRatio.ONE_ONE -> AspectRatioProto.ASPECT_RATIO_ONE_ONE
+        }
+        jcaSettings.updateData { currentSettings ->
+            currentSettings.toBuilder()
+                .setAspectRatioStatus(newStatus)
+                .build()
+        }
```
yasith, line 130: We can use the more kotlin idiomatic version with the `copy` function.

```
currentSettings.copy {
     aspectRatio = newAspectRatio
}
```

data/settings/src/main/java/com/google/jetpackcamera/settings/LocalSettingsRepository.kt
```diff
@@ -50,7 +51,17 @@ class LocalSettingsRepository @Inject constructor(
...
                 front_camera_available = it.frontCameraAvailable,
-                back_camera_available = it.backCameraAvailable
+                back_camera_available = it.backCameraAvailable,
+                aspect_ratio = when (it.aspectRatioStatus) {
+                    AspectRatioProto.ASPECT_RATIO_NINE_SIXTEEN -> AspectRatio.NINE_SIXTEEN
+                    AspectRatioProto.ASPECT_RATIO_ONE_ONE -> AspectRatio.ONE_ONE
+                    // defaults to 3:4 aspect ratio
+                    AspectRatioProto.ASPECT_RATIO_THREE_FOUR,
+                    AspectRatioProto.UNRECOGNIZED,
+                    null -> AspectRatio.THREE_FOUR
```
yasith, line 61: nit: We should make this a function on `model.AspectRatoi`, it'll help make this area of the code cleaner when we add more settings.

```
def fromProto(proto: AspectRatioProto) = when...
```

yasith (changes requested):

data/settings/src/main/proto/com/google/jetpackcamera/settings/aspect_ratio.proto
Kimblebee: we'll have a name conflict with the AspectRatio enum from `com.google.jetpackcamera.settings.model`, as well as another auto-generated class (from using proto datastore) in the build folder since we were using; 
I was intentional about keeping proto in the name to help differentiate from the enum equivalent we use for the rest of the app, see `LocalSettingsRepository.kt`

data/settings/src/main/java/com/google/jetpackcamera/settings/LocalSettingsRepository.kt
Kimblebee: i had written the update functions like that previously, but it seems the best practice way is to follow the builder method

feature/settings/src/main/res/values/strings.xml
```diff
@@ -48,4 +48,15 @@
...
+    <!-- Aspect Ratio  setting strings -->
+    <string name="aspect_ratio_title">Set Aspect Ratio</string>
+
+    <string name="aspect_ratio_selector_9_16">9:16</string>
+    <string name="aspect_ratio_selector_3_4">3:4</string>
+    <string name="aspect_ratio_selector_1_1">1:1</string>
+
+    <string name="aspect_ratio_description_9_16">Aspect Ratio is 9:16</string>
+    <string name="aspect_ratio_description_3_4">Aspect Ratio is 3:4</string>
+    <string name="aspect_ratio_description_1_1">Aspect Ratio is 1:1</string>
```
davidjiagoogle, line 60: How do these read out loud? We should make sure the accessibility actually says "nine to sixteen"

davidjiagoogle (approved):

feature/settings/src/main/res/values/strings.xml
Kimblebee: this is actually just the description text that is visible.
I actually still need to update the composable parameters to take in accessibility text for ALL settings. 
I'll put this into a todo and write a bug for that.

PR #39, opened 2023-09-20
Migrate to GraphicsSurface and EmbeddedGraphicsSurface

yasith:
- This will break our current implementation of tap to focus. This is unavoidable right now since GraphicsSurface/EmbeddedGraphicsSurface don't expose any View. Will follow up with a fix for tap to focus using a new API for exposing a MeteringPoint.


settings.gradle.kts
```diff
@@ -9,7 +9,7 @@ dependencyResolutionManagement {
     repositoriesMode.set(RepositoriesMode.FAIL_ON_PROJECT_REPOS)
     repositories {
         maven {
-            setUrl("https://androidx.dev/snapshots/builds/10707469/artifacts/repository")
+            setUrl("https://androidx.dev/snapshots/builds/10836739/artifacts/repository")
```
yasith, line 12: I didn't switch to `latest` since the package versions can change. eg. camera changed from 1.3.0 to 1.4.0 from previous version to latest.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewScreen.kt
```diff
@@ -171,12 +171,6 @@ fun PreviewScreen(
                         modifier = Modifier
                             .fillMaxSize(),
                         onSurfaceProviderReady = onSurfaceProviderReady,
-                        onRequestBitmapReady = {
```
temcguir, line 174: Is this also functionality that we'll be losing by migrating? Do we need to request a feature for this?

PR #44, opened 2023-09-29
Transform TextureView according to aspect ratio

davidjiagoogle:
Added SurfaceTransformatiomUtil which contains transformation logics borrowed from CameraX's ViewFinderTransformation.

The Texture is scaled to fill center the parent view.

Tested on Samsung, Pixel, and foldable devices.


domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
```diff
@@ -72,7 +74,10 @@ class CameraXCameraUseCase @Inject constructor(
 
     //TODO apply flash from settings
     private val imageCaptureUseCase = ImageCapture.Builder().build()
-    private val previewUseCase = Preview.Builder().build()
+    private val previewUseCase: Preview = Preview.Builder().setResolutionSelector(
+        ResolutionSelector.Builder()
+            .setResolutionStrategy(ResolutionStrategy.HIGHEST_AVAILABLE_STRATEGY).build()
```
temcguir, line 79: Is there a reason we need to use `HIGHEST_AVAILABLE_STRATEGY`? I would think we would be better off relying on the default resolution.

camera-viewfinder-compose/src/main/java/com/google/jetpackcamera/viewfinder/surface/SurfaceTransformationUtil.kt
```diff
@@ -0,0 +1,207 @@
...
+                // Viewport is not based on the PreviewView or 3) both.
+                getPreviewViewViewportRectForMismatchedAspectRatios(
+                    transformationInfo, previewViewSize, resolution
+                )
+            }
+        val resolutionRect = RectF(0F, 0F, resolution.width.toFloat(), resolution.height.toFloat())
+        val matrix = TransformUtils.getRectToRect(
+            resolutionRect, previewViewCropRect, transformationInfo.rotationDegrees
+        )
+        if (isFrontCamera && transformationInfo.hasCameraTransform()) {
```
temcguir, line 158: There seems to be an inssue when switching to "single stream" where the image gets stretched. I think it may have something to do with rotation somehow affecting the expected aspect ratio (i.e., 16:9 rather than 9:16). We may need to do a similar `if` condition to this one where we check `hasCameraTransform` to see if rotation has already been applied so we scale along the correct axes.

temcguir (changes requested):
I left a comment about incorrect scaling for "Single Stream" that should probably be fixed before merging, but it could also be ok to handle it in a separate PR if you prefer.

temcguir (approved):
Thanks for the changes!

PR #45, opened 2023-10-06
Create a Configuration data class to use with CameraUseCase implementations

yasith:
This is the 1st PR in a chain to refactor CameraUseCase to enforce UDF (Unidirectional Data Flow).

This data class will act as the holder for camera related configurations.

We can observe the StateFlow<Config> from ViewModel in order to generate the necessary values for views, while respecting UDF.

PR #46, opened 2023-10-06
Update CameraUseCase functions to use CameraUseCase.Config

yasith:
Change CameraUseCase functions to use the newly created Config data class and the related enums for configuration changes

### DIFF
--- a/domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
+++ b/domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
@@ -22,7 +22,6 @@ import android.util.Log
 import android.view.Display
 import androidx.camera.core.Camera
 import androidx.camera.core.CameraSelector
-import androidx.camera.core.CameraSelector.LensFacing
 import androidx.camera.core.DisplayOrientedMeteringPointFactory
 import androidx.camera.core.FocusMeteringAction
 import androidx.camera.core.ImageCapture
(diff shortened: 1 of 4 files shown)


yasith:
Current dependencies on/for this PR:
* main
  * **PR #45** <a href="https://app.graphite.dev/github/pr/google/jetpack-camera-app/45" target="_blank">Graphite</a> 
    * **PR #46** <a href="https://app.graphite.dev/github/pr/google/jetpack-camera-app/46" target="_blank">Graphite</a>  👈
      * **PR #47** <a href="https://app.graphite.dev/github/pr/google/jetpack-camera-app/47" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/google/jetpack-camera-app/46?utm_source=stack-comment).

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewViewModel.kt
```diff
@@ -170,7 +192,7 @@ class PreviewViewModel @Inject constructor(
                     )
                 )
                 // apply to cameraUseCase
-                cameraUseCase.flipCamera(previewUiState.value.currentCameraSettings.isFrontCameraFacing)
+                cameraUseCase.setLensFacing(cameraUseCase.config.value.lensFacing.next())
```
temcguir, line 195: If using the `next()` extension function, do we still need the `isFacingFront` boolean passed in to this method? Also, if we end up adding another lens facing enum, such as EXTERNAL, then `flipCamera` might not be the best name for this method since we could be switching between 3 cameras.

domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
```diff
@@ -182,12 +174,11 @@ class CameraXCameraUseCase @Inject constructor(
 
     private fun getZoomState(): ZoomState? = camera?.cameraInfo?.zoomState?.value
 
-    // flips the camera to the designated lensFacing direction
-    override suspend fun flipCamera(isFrontFacing: Boolean) {
-        this.isFrontFacing = isFrontFacing
-        rebindUseCases(
-
+    override suspend fun setLensFacing(lensFacing: CameraUseCase.LensFacing) {
+        _config.emit(
```
temcguir, line 178: If performing an action that needs to rebind use cases, we should emit after rebinding since `rebindUseCases` may throw an exception. This will probably make more sense once we have the prescriptive config since right now the config here is used in `rebindUseCase` and is acting as both the prescriptive and descriptive config.

temcguir (approved):
Overall looks good. I just had a few comments that can probably be addressed in future PRs. For now, maybe only the `flipCamera` comment should be addressed here.

PR #58, opened 2023-10-11
Close camera on backgrounding the app

davidjiagoogle:
Before the change, PreviewViewModel.stopCamera() only cancels the job. If the camera is already open and the useCases already bound, it doesn't do anything, resulting in the camera still being open even when the app is backgrounded.

In this PR, cameraProvider.unbindAll() is called in stopCamera so that existing bound useCases would unbind and the camera would stop. Since this is evoked in lifeCycle's onStop, the crash is fixed


feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewViewModel.kt
```diff
@@ -100,6 +100,9 @@ class PreviewViewModel @Inject constructor(
                 cancel()
             }
         }
+        viewModelScope.launch {
```
yasith, line 103: I think we should handle this from inside the `CameraUseCase` without needing the ViewModel's manual intervention.

When `runningCameraJob.cancel()` is called I think we'll get to `awaitCancellation()` in CameraXCameraUseCase.

From there we can do a finally block, and call `cameraProvider.unbindAll()`

Lets wait to hear from @temcguir on this too.

yasith (commented):
As an aside, I think some of your older branches are attached to your current pull requests for some reason.

If you pull on main, and branch from there I think it'll be fixed for future PRs.

Not a big problem, but when merging, you should make sure to remove the unnecessary details from the commit description.

PR #60, opened 2023-10-12
Add loading animations for MainActivity and Camera

Kimblebee:
a loading animation when the app initially launches and fetches datastore information, and animation while cameraState is not ready


app/src/main/java/com/google/jetpackcamera/MainActivity.kt
```diff
@@ -66,6 +76,16 @@ class MainActivity : ComponentActivity() {
         setContent {
             when (uiState) {
                 Loading -> {/* TODO(b/283812742):  a loading screen */
```
yasith, line 78: Can we remove this TODO, now that you're adding the loading screen?

yasith (approved):

PR #63, opened 2023-10-12
Backgrounding crash fix

davidjiagoogle:
A better fix than https://github.com/google/jetpack-camera-app/pull/58

When camera is flipped, it is not assigned to a runningCameraJob, this caused the CorotineCameraProvider to not be able to unbind camera on job cancelled


temcguir (approved):
Good catch. This looks pretty fragile, so we should probably have a test for backgrounding the app and returning using Espresso. But that can be a separate PR.

PR #64, opened 2023-10-16
Enable Preview Stabilization if devices support for JCA

kailianc:


domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
```diff
@@ -268,6 +279,34 @@ constructor(
...
+    @OptIn(ExperimentalCamera2Interop::class)
+    private fun createPreviewUseCase() : Preview {
+        var isPreviewStabilizationSupported = false
+
+        // TODO: find a better way to access camera info of front/back camera before binding
+        //  use cases
+        val availableCameraInfo = cameraProvider.availableCameraInfos
+        if (availableCameraInfo.isNotEmpty()) {
+            for (cameraInfo in availableCameraInfo) {
+                val lensFacing = Camera2CameraInfo.from(cameraInfo).getCameraCharacteristic(
```
temcguir, line 291: You can avoid Camera2Interop here by using `CameraSelector.filter`. For instance, you could write something like:

```
val cameraSelector = if (isFrontFacing) {
    CameraSelector.DEFAULT_FRONT_CAMERA
} else {
    CameraSelector.DEFAULT_BACK_CAMERA
}
val isPreviewStabilizationSupported =
    cameraSelector.filter(availableCameraInfo).firstOrNull()?.let {
        Preview.getPreviewCapabilities(it).isStabilizationSupported
    } ?: false
```

domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
```diff
@@ -280,6 +319,7 @@ constructor(
                 getLensFacing(isFrontFacing)
             )
         cameraProvider.unbindAll()
+
```
temcguir, line 322: nit: unnecessary additional whitespace

domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
```diff
@@ -268,6 +279,34 @@ constructor(
...
+        //  use cases
+        val availableCameraInfo = cameraProvider.availableCameraInfos
+        if (availableCameraInfo.isNotEmpty()) {
+            for (cameraInfo in availableCameraInfo) {
+                val lensFacing = Camera2CameraInfo.from(cameraInfo).getCameraCharacteristic(
+                    CameraCharacteristics.LENS_FACING)
+                if ((isFrontFacing && CameraCharacteristics.LENS_FACING_FRONT == lensFacing)
+                    || (!isFrontFacing && CameraCharacteristics.LENS_FACING_BACK == lensFacing)) {
+                    isPreviewStabilizationSupported =
+                        Preview.getPreviewCapabilities(cameraInfo)
```
temcguir, line 296: Looks like `Preview.getPreviewCapabilities()` is restricted. You'll need to suppress the warning and we should track updating this when it is no longer restricted.

temcguir (changes requested):

domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
kailianc: Thanks. This is much cleaner now to avoid Camera2CameraInfo.

domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
kailianc: Done

domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
kailianc: The suppress warning is because of Camera2CameraInfo, so this is not necessary now after using `CameraSelector.filter`.

temcguir: Android studio still gives me a red squiggly under `getPreviewCapabilities()`, so you may need `@SuppressLint("RestrictedApi")` to suppress it.

kailianc: I unhide this API several days ago, aosp/2786388.
Are you using the old build id snapshot?

temcguir: Ah it looks like my project was still syncing :)
Looks good now.

temcguir (approved):

domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
kailianc: Thanks!

yasith (approved):

PR #68, opened 2023-10-18
Fix the preview stream freeze issue when changing aspect ratio and ca…

kailianc:
…pture mode


temcguir (approved):
Thanks for fixing this!

PR #71, opened 2023-10-19
Stop RuningCameraJob on all rebindUseCase() instances

davidjiagoogle:
setCaptureMode() and setAspectRatio() also invokes rebindUseCase(), before which the old runningCameraJob needs to be stopped and updated

### DIFF
--- a/app/src/androidTest/java/com/google/jetpackcamera/BackgroundDeviceTest.kt
+++ b/app/src/androidTest/java/com/google/jetpackcamera/BackgroundDeviceTest.kt
@@ -1,0 +1,89 @@
+/*
+ * Copyright (C) 2023 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.jetpackcamera
+
+import android.content.Context
+import android.view.KeyEvent
+import androidx.test.InstrumentationRegistry.getTargetContext
+import androidx.test.core.app.ActivityScenario
+import androidx.test.core.app.ApplicationProvider
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import androidx.test.platform.app.InstrumentationRegistry
+import androidx.test.rule.GrantPermissionRule
+import androidx.test.uiautomator.By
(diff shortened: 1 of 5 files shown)


yasith (approved):

temcguir (changes requested):
Can we add some tests to background the app and return (including a test to change aspect ratio before backgrounding)?

This should be possible with UiAutomator.

davidjiagoogle:
@temcguir I'll add the tests in a separate PR

app/src/androidTest/java/com/google/jetpackcamera/BackgroundDeviceTest.kt
```diff
@@ -0,0 +1,85 @@
...
+    @get:Rule
+    val cameraPermissionRule: GrantPermissionRule =
+        GrantPermissionRule.grant(android.Manifest.permission.CAMERA)
+
+    private val instrumentation = InstrumentationRegistry.getInstrumentation()
+    private val uiDevice = UiDevice.getInstance(instrumentation)
+
+    private fun backgroundThenForegroundApp() {
+        uiDevice.pressHome()
+        Thread.sleep(1500)
```
temcguir, line 41: Instead of using `sleep`, can you use `UiDevice.waitForIdle(timeout)`? You can probably use the same sleep duration for the timeout.

app/src/androidTest/java/com/google/jetpackcamera/BackgroundDeviceTest.kt
```diff
@@ -0,0 +1,85 @@
...
+        uiDevice.click(uiDevice.displayWidth / 2, uiDevice.displayHeight / 2)
+        Thread.sleep(2000)
+
+    }
+
+    @Before
+    fun setUp() {
+        ActivityScenario.launch(MainActivity::class.java)
+        Thread.sleep(2000)
```
temcguir, line 52: This sleep can be removed since every method in `UiDevice` internally calls `waitForIdle()`. If you want to timeout waiting for the device to idle, you can replace the sleep with `UiDevice.waitForIdle(timeout)`.

app/src/androidTest/java/com/google/jetpackcamera/BackgroundDeviceTest.kt
```diff
@@ -0,0 +1,85 @@
...
+        backgroundThenForegroundApp()
+    }
+
+    @Test
+    fun setAspectRatio_then_background_foreground() {
+        uiDevice.findObject(By.res("QuickSettingDropDown")).click()
+        uiDevice.findObject(By.res("QuickSetAspectRatio")).click()
+        uiDevice.findObject(By.res("QuickSetAspectRatio1:1")).click()
+        uiDevice.findObject(By.res("QuickSettingDropDown")).click()
+        InstrumentationRegistry.getInstrumentation().waitForIdleSync()
```
temcguir, line 75: Might as well also replace these `waitForIdleSync` calls with `UiDevice.waitForIdle(timeout)` for consistency and so we can add timeouts.

temcguir (changes requested):

temcguir (approved):

PR #73, opened 2023-10-19
Apply all spotless fixes that required manual intervention

temcguir:
Ran `spotlessApply` and fixed all issues that caused errors. Some
 automatic fixes were applied after fixing manual issues.

 This required temporarily turning off
 `ratchetFrom("origin/main")`

### DIFF
--- a/app/src/main/java/com/google/jetpackcamera/ui/Routes.kt
+++ b/app/src/main/java/com/google/jetpackcamera/ui/Routes.kt
@@ -13,10 +13,9 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
 package com.google.jetpackcamera.ui
 
 object Routes {
-    const val PreviewRoute = "preview"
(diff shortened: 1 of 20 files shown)


yasith (approved):

app/src/main/java/com/google/jetpackcamera/ui/Routes.kt
```diff
@@ -13,10 +13,9 @@
...
  */
-
 package com.google.jetpackcamera.ui
 
 object Routes {
-    const val PreviewRoute = "preview"
-    const val SettingsRoute = "settings"
-}
\ No newline at end of file
+    const val PREVIEW_ROUTE = "preview"
```
yasith, line 19: I think you need to change where these are used too

temcguir: Yeah apparently `spotlessApply` renamed this automatically but didn't update references. I've updated the references now.

PR #80, opened 2023-11-15
foundation of JCA macrobenchmark integration

Kimblebee:


app/proguard-benchmark-rules.pro
```diff
@@ -0,0 +1 @@
+-keep class androidx.datastore.*.** {*;}
```
temcguir, line 1: I'm not seeing this file be used in the `benchmark` build type in `build.gradle.kts`. Is it actually necessary?

benchmark/build.gradle.kts
```diff
@@ -0,0 +1,71 @@
...
+        sourceCompatibility = JavaVersion.VERSION_1_8
+        targetCompatibility = JavaVersion.VERSION_1_8
+    }
+
+    kotlinOptions {
+        jvmTarget = "1.8"
+    }
+
+    defaultConfig {
+        minSdk = 23
```
temcguir, line 36: Since macrobenchmark has a minSdk of 23, but our app has a minSdk of 21, please put a comment here explaining that.

benchmark/build.gradle.kts
```diff
@@ -0,0 +1,71 @@
...
+        targetSdk = 33
```
temcguir, line 37: Can this be bumped to 34 to match the app `targetSdk`?

benchmark/build.gradle.kts
```diff
@@ -0,0 +1,71 @@
...
+        // for easy local/CI testing.
+        create("benchmark") {
+            isDebuggable = true
+            signingConfig = getByName("debug").signingConfig
+            matchingFallbacks += listOf("release")
+        }
+    }
+
+    targetProjectPath = ":app"
+    experimentalProperties["android.experimental.self-instrumenting"] = true
```
temcguir, line 57: If this is required for macrobenchmark, maybe add a comment stating so. You could also link to https://source.android.com/docs/core/tests/development/instr-self-e2e

benchmark/build.gradle.kts
```diff
@@ -0,0 +1,71 @@
...
+}
+
+dependencies {
+    implementation("androidx.test.ext:junit:1.1.5")
+    implementation("androidx.test.espresso:espresso-core:3.5.1")
```
temcguir, line 62: Do we need this espresso dependency? I don't see any espresso imports in the test file.

benchmark/build.gradle.kts
```diff
@@ -0,0 +1,71 @@
...
+    implementation("androidx.test.uiautomator:uiautomator:2.2.0")
```
temcguir, line 63: Similar question here; do we need this yet? If we need UiAutomator in a subsequent PR we can add it then.

benchmark/build.gradle.kts
```diff
@@ -0,0 +1,75 @@
...
+        create("benchmark") {
+            isDebuggable = true
+            signingConfig = getByName("debug").signingConfig
+            matchingFallbacks += listOf("release")
+        }
+    }
+
+    targetProjectPath = ":app"
+    experimentalProperties["android.experimental.self-instrumenting"] = true
+    // required for benchmark:
```
temcguir, line 60: nit: Usually we put the comments above the line they are referring to. So it should be between line 58 and 59 here.

temcguir (approved):

PR #82, opened 2023-11-22
Update CI to run on pull requests instead of push

yasith:
Some of our CI tests are not triggering when there's a new pull request from outside of the repository.

This should be fixed by changing the event it's triggered on to `pull_request` instead of `push`


Zeronfinity (approved):
Thanks for the update!

PR #83, opened 2023-11-28
Settings to enable preview/video stabilization

Kimblebee:
These are settings to designate the stabilization mode for the preview and video. The 3 available modes are on, high-quality, or off.

Refer to [this PR](#94) to review the extra commentary and changes regarding this feature's development

- CameraXCameraUseCase sets preview/video stabilization support information to datastore
- cameraAppSettings condenses datastore stabilization support info to an enum to make it easier to use around the app
- disables stabilization setting if unsupported by device
- disables On option if preview stabilization is unavailable
- disables High Qualityoption if video stabilization is unavailable
- disables entire setting if neither preview nor video stabilization is supported
- display an icon to preview screen when stabilization is active


domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
```diff
@@ -289,8 +319,12 @@ constructor(
             } ?: false
 
         val previewUseCaseBuilder = Preview.Builder()
-        if (isPreviewStabilizationSupported) {
-            previewUseCaseBuilder.setPreviewStabilizationEnabled(true)
+        if (isPreviewStabilizationSupported && shouldStabilizePreview != Stabilization.UNDEFINED) {
+            val isStabilized = when (shouldStabilizePreview) {
+                Stabilization.ON -> true
+                else -> false
```
yasith, line 325: nit: use Stabilization.OFF branch instead of else. We don't need an else because it's an enum

feature/settings/src/main/java/com/google/jetpackcamera/settings/SettingsScreen.kt
```diff
@@ -78,6 +80,18 @@ fun SettingsList(uiState: SettingsUiState, viewModel: SettingsViewModel) {
         setCaptureMode = viewModel::setCaptureMode
     )
 
+    // todo(b/313647247): query device and disable setting if stabilization isn't supported
```
yasith, line 83: nit: go/todo-style

feature/settings/src/main/java/com/google/jetpackcamera/settings/SettingsViewModel.kt
```diff
@@ -126,4 +127,28 @@ class SettingsViewModel @Inject constructor(
...
     }
+
+    fun setPreviewStabilization(stabilization: Stabilization) {
+        viewModelScope.launch {
+            settingsRepository.updatePreviewStabilization(stabilization)
+
+            Log.d(
+                TAG,
+                "set preview stabilization: " +
+                    settingsRepository.getCameraAppSettings().previewStabilization
```
yasith, line 138: Use Kotlin idiomatic String formatting with ${}

feature/settings/src/main/java/com/google/jetpackcamera/settings/ui/SettingsComponents.kt
```diff
@@ -240,6 +243,92 @@ fun CaptureModeSetting(currentCaptureMode: CaptureMode, setCaptureMode: (Capture
...
+                    selected = currentPreviewStabilization == Stabilization.ON,
+                    onClick = { setVideoStabilization(Stabilization.ON) }
+                )
+                SingleChoiceSelector(
+                    text = stringResource(id = R.string.stabilization_selector_off),
+                    selected = currentPreviewStabilization == Stabilization.OFF,
+                    onClick = { setVideoStabilization(Stabilization.OFF) }
+                )
+                SingleChoiceSelector(
+                    text = stringResource(id = R.string.stabilization_selector_undefined),
```
temcguir, line 323: While I can appreciate trying to mirror the APIs, we should consider that the APIs are meant for developers whereas UI is meant for app users. An app user probably has no idea what "`UNDEFINED`" means.

Looking at the table:
image

Really, there are only 3 options that users should care about:
1. Preview AND the recorded video are stabilized (WYSIWYG) [rows 1, 3]
2. "High quality" recorded video stabilization (Preview _might_ be stabilized, depending on device) [row 7]
3. Stabilization off [all other rows]

I would only display these 3 options in the UI (we could call them `ON`, `HIGH QUALITY`, `OFF`), but add longer descriptions underneath each option to explain the difference between `ON` and `HIGH_QUALITY`.

feature/settings/src/main/java/com/google/jetpackcamera/settings/SettingsScreen.kt
```diff
@@ -78,6 +80,18 @@ fun SettingsList(uiState: SettingsUiState, viewModel: SettingsViewModel) {
         setCaptureMode = viewModel::setCaptureMode
     )
 
+    // todo: b/313647247 - query device and disable setting if stabilization isn't supported.
+    PreviewStabilizeSetting(
```
temcguir, line 84: Can we also add a UI affordance (a small icon) to `PreviewScreen` to show when stabilization is enabled? We already have the `video_stable` material icon we can use. This will also allow you to write some tests to check that changing the settings in the SettingsScreen has some effect in the PreviewScreen.

temcguir (changes requested):

feature/settings/src/main/java/com/google/jetpackcamera/settings/SettingsScreen.kt
Kimblebee: yes but it will need to have information from CameraXCameraUseCase to know if stabilization is actually supported.

Right now the UI on preview/quicksettings reflect the information in the CameraAppSettings (currentCameraSettings). I think for now we can wait until the UDF refactor is complete, and add the supported information to the config

temcguir: As discussed offline, we should check once when starting up which stabilization modes are available and only present those as options in the quick settings / settings. Once those settings are selected in the UI, we can assume they're being applied and show an icon, since we've already checked that they're supported.

This may be sufficient, but later on we should also handle cases where we set a setting and CameraXCameraUseCase tells us that the camera wasn't able to apply the setting. This will likely be handled by showing some kind of error in the UI.

temcguir (approved):
As discussed offline, please follow up with a UiAutomator test.

PR #86, opened 2023-12-07
Target FPS settings

Kimblebee:
Adds settings to set default target FPS
**_Since preview/video stabilization affects target fps, it will be disabled when a target fps other than "none" is selected._**

added target fps options of 15, 30, and 60.

The application of the target fps logic occurs in `CameraXCameraUseCase`

everything else  relates to datastore or Settings Screen UI

_Current known issue: The change in the default setting screen currently does not apply to the current camera session, and is applied on app start._


data/settings/src/main/java/com/google/jetpackcamera/settings/model/CameraAppSettings.kt
```diff
@@ -25,7 +25,9 @@ data class CameraAppSettings(
     val darkMode: DarkMode = DarkMode.SYSTEM,
     val flashMode: FlashMode = FlashMode.OFF,
     val captureMode: CaptureMode = CaptureMode.MULTI_STREAM,
-    val aspectRatio: AspectRatio = AspectRatio.NINE_SIXTEEN
+    val aspectRatio: AspectRatio = AspectRatio.NINE_SIXTEEN,
+    val targetFrameRate: TargetFrameRate = TargetFrameRate.TARGET_FPS_NONE
+
```
yasith, line 30: nit: extra new line

data/settings/src/main/java/com/google/jetpackcamera/settings/model/TargetFrameRate.kt
```diff
@@ -0,0 +1,53 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.jetpackcamera.settings.model
+
+import android.util.Range
+import com.google.jetpackcamera.settings.TargetFrameRate as TargetFrameRateProto
+
+enum class TargetFrameRate(val range: Range<Int>) {
+    TARGET_FPS_NONE(Range(0, 0)),
```
yasith, line 22: why do we have a range, if both values are the same?

data/settings/src/main/proto/com/google/jetpackcamera/settings/target_frame_rate.proto
```diff
@@ -0,0 +1,30 @@
...
+option java_multiple_files = true;
+
+enum TargetFrameRate {
+  TARGET_FRAME_RATE_UNDEFINED = 0;
+  TARGET_FRAME_RATE_NONE = 1;
+  TARGET_FRAME_RATE_15 = 2;
+  TARGET_FRAME_RATE_30 = 3;
+  TARGET_FRAME_RATE_60 = 4;
+  TARGET_FRAME_RATE_100 = 5; // excessively high option to target the supported upper ranges
+
```
yasith, line 29: nit: extra new line

domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
```diff
@@ -289,7 +301,7 @@ constructor(
             } ?: false
 
         val previewUseCaseBuilder = Preview.Builder()
-        if (isPreviewStabilizationSupported) {
+        if (isPreviewStabilizationSupported && shouldStabilize) {
```
yasith, line 304: Do we also have to set `setPreviewStabilizationEnabled` to `false` if we don't meet the criteria?

yasith (approved):

data/settings/src/main/java/com/google/jetpackcamera/settings/model/TargetFrameRate.kt
```diff
@@ -0,0 +1,53 @@
...
+    TARGET_FPS_15(Range(15, 15)),
+    TARGET_FPS_30(Range(30, 30)),
+    TARGET_FPS_60(Range(60, 60)),
+    TARGET_FPS_100(Range(100, 100)); // excessively high option to target the supported upper ranges
```
temcguir, line 26: There won't be any devices that support this. I don't think we should even call it out as an option.

feature/settings/src/main/java/com/google/jetpackcamera/settings/ui/SettingsComponents.kt
```diff
@@ -240,6 +242,61 @@ fun CaptureModeSetting(currentCaptureMode: CaptureMode, setCaptureMode: (Capture
     )
 }
 
+@Composable
+fun TargetFpsSetting(
+    modifier: Modifier = Modifier,
+    currentTargetFps: TargetFrameRate,
+    setTargetFps: (TargetFrameRate) -> Unit
+) {
+    BasicPopupSetting(
```
temcguir, line 251: We need to check against the supported FPS ranges in `CameraInfo.getSupportedFrameRateRanges()` to check which options we should show here. We don't need to show all of the ranges, but at least need to match them up to 15, 30, 60.

temcguir (changes requested):

feature/settings/src/main/java/com/google/jetpackcamera/settings/ui/SettingsComponents.kt
Kimblebee: right now i dont think the default settings are connected to the camera. I'll need to figure out how to send that information from the camera to settings, since its only source of information currently comes from the datastore.

I could have the cameraxCameraUseCase populate the datastore with information regarding device available fps at init, similar to the information for front or rear camera availability.

domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
Kimblebee: I intentionally chose to leave it alone as unspecified, since we aren't explicitly saying to leave it off. setPreviewStabilizationEnabled can be set to true (on), false (off), or leaving it alone to be unspecified, since setting preview or video stabilization may be different. 

Noting that, something I still need to test is if the fps selection issue still occurs when *only* video capture is stabilized.

image

data/settings/src/main/java/com/google/jetpackcamera/settings/model/TargetFrameRate.kt
temcguir: Gotcha. I think it might cause more confusion than anything. Even if it's choosing the highest possible FPS, it's still not choosing "100fps", which is what a user who is not familiar with cameras or CameraX would expect. Thanks for taking it out.

feature/settings/src/main/java/com/google/jetpackcamera/settings/ui/SettingsComponents.kt
temcguir: That's a good idea. We should populate the datastore from CameraUseCase somehow. We should then check if the 3 options of 15, 30 and 60 are supported, and hide them as an option in the UI if they're not supported.

data/settings/src/main/java/com/google/jetpackcamera/settings/model/TargetFrameRate.kt
```diff
@@ -0,0 +1,49 @@
...
+
+enum class TargetFrameRate(val range: Range<Int>) {
+    TARGET_FPS_NONE(Range(0, 0)),
+    TARGET_FPS_15(Range(15, 15)),
+    TARGET_FPS_30(Range(30, 30)),
+    TARGET_FPS_60(Range(60, 60));
+
+    companion object {
+        // converts our kotlin enum to a proto enum
+        fun toProto(targetFrameRate: TargetFrameRate): TargetFrameRateProto {
```
jsaund, line 29: it might be easier to use if this is moved out of the companion object (but leave the `fromProto` since it's a constructor of this enum type)

```
enum class TargetFrameRate(val range: Range<Int>) {
    ...
    fun toProto(): TargetFrameRateProto = when (this) {
        ...
    }
}
```

Then it can be used as:
`TargetFrameRate.TARGET_FPS_15.toProto()`

domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
```diff
@@ -134,7 +132,18 @@ constructor(
...
-        videoCaptureUseCase = createVideoUseCase()
+
+        videoCaptureUseCase = when (currentCameraSettings.targetFrameRate) {
+            TargetFrameRate.TARGET_FPS_NONE -> {
+                VideoCapture.withOutput(recorder)
+            }
+
+            else -> {
+                VideoCapture.Builder(recorder)
+                    .setTargetFrameRate(currentCameraSettings.targetFrameRate.range)
```
temcguir, line 143: It looks like this is only happening once during initialization. Shouldn't we be reacting to changes in this frame rate and updating the use cases / rebinding?

Kimblebee: before the update introducing perpetual settings, the settings would automatically be applied whenever navigating back to the preview screen.
like i mentioned offline, it now has the same issue as #131 where the setting is only applied on app startup, so I will need to still to take care of that

feature/settings/src/main/res/values/strings.xml
```diff
@@ -83,5 +83,24 @@
     <string name="stabilization_description_on">Stabilization On</string>
     <string name="stabilization_description_high_quality">Stabilization High Quality</string>
     <string name="stabilization_description_off">Stabilization Off</string>
-    <string name="stabilization_description_unsupported">Stabilization unsupported</string>
+    <string name="stabilization_description_unsupported_device">Stabilization unsupported by device</string>
+    <string name="stabilization_description_unsupported_fps">Stabilization unsupported due to frame rate</string>
+
+
+    <!-- Target Fps setting strings -->
+    <string name="fps_title">Set Target Frame Rate</string>
```
temcguir, line 91: We should drop the "Target" word usage from the UI strings. In JCA, we want to always hit the target FPS, and should disable features if we can't. So if an FPS is set and the camera doesn't provide it, that should be considered a bug. We can keep the "Target" word in variable names, etc., but just don't use it in the UI.

feature/settings/src/main/java/com/google/jetpackcamera/settings/SettingsViewModel.kt
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2023 The Android Open Source Project
+ * Copyright (C) 2023-2024 The Android Open Source Project
```
temcguir, line 2: Leave this as 2023.

data/settings/src/main/proto/com/google/jetpackcamera/settings/target_frame_rate.proto
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+syntax = "proto3";
+
+option java_package = "com.google.jetpackcamera.settings";
+option java_multiple_files = true;
+
+enum TargetFrameRate {
+  TARGET_FRAME_RATE_UNDEFINED = 0;
+  TARGET_FRAME_RATE_NONE = 1;
```
temcguir, line 24: Why do we need UNDEFINED and NONE? What is the difference?

data/settings/src/main/proto/com/google/jetpackcamera/settings/jca_settings.proto
```diff
@@ -44,4 +45,6 @@ message JcaSettings {
   DynamicRange dynamic_range_status = 13;
   repeated DynamicRange supported_dynamic_ranges = 14;
   LensFacing default_lens_facing = 15;
+  TargetFrameRate target_frame_rate = 16;
```
temcguir, line 48: Is there a reason this can't just be an int32 rather than an enum? It's possible some devices might not support 15fps, for instance, and may support something like 12 fps. It will be much easier to just keep things as ints so that we can more easily support cases like this in the future.

data/settings/src/main/java/com/google/jetpackcamera/settings/test/FakeSettingsRepository.kt
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2023 The Android Open Source Project
+ * Copyright (C) 2023-2024 The Android Open Source Project
```
temcguir, line 2: Leave as 2023.

data/settings/src/main/java/com/google/jetpackcamera/settings/test/FakeSettingsRepository.kt
```diff
@@ -116,4 +117,16 @@ object FakeSettingsRepository : SettingsRepository {
         currentCameraSettings =
             currentCameraSettings.copy(aspectRatio = aspectRatio)
     }
+
+    override suspend fun updateTargetFrameRate(targetFrameRate: TargetFrameRate) {
+        TODO("Not yet implemented")
```
temcguir, line 122: Please add implementation:
```
currentCameraSettings =
        currentCameraSettings.copy(targetFrameRate = targetFrameRate)
```

feature/settings/src/main/res/values/strings.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!--
-  ~ Copyright (C) 2023 The Android Open Source Project
+  ~ Copyright (C) 2023-2024 The Android Open Source Project
```
temcguir, line 3: Leave as 2023.

feature/settings/src/main/res/values/strings.xml
```diff
@@ -83,5 +83,24 @@
...
-    <string name="stabilization_description_unsupported">Stabilization unsupported</string>
+    <string name="stabilization_description_unsupported_device">Stabilization unsupported by device</string>
+    <string name="stabilization_description_unsupported_fps">Stabilization unsupported due to frame rate</string>
+
+
+    <!-- Target Fps setting strings -->
+    <string name="fps_title">Set Target Frame Rate</string>
+
+    <string name="fps_description_unavailable">No fixed frame rates supported</string>
+    <string name="fps_description_none">No target fps</string>
```
temcguir, line 94: When we drop "Target", we should just call this "Auto Frame Rate".

feature/settings/src/main/java/com/google/jetpackcamera/settings/ui/SettingsComponents.kt
```diff
@@ -325,14 +415,16 @@ fun StabilizationSetting(
                 )
 
                 // high quality selector
+                // disabled if target fps = 60
```
temcguir, line 418: We need to expand on this comment why we turn this off for FPS == 60.

feature/settings/src/main/java/com/google/jetpackcamera/settings/ui/SettingsComponents.kt
```diff
@@ -311,11 +393,19 @@ fun StabilizationSetting(
             Column(Modifier.selectableGroup()) {
                 Spacer(modifier = Modifier.height(10.dp))
 
-                // on selector
+                // on (preview) selector
+                // disabled if target fps != (30 or off)
```
temcguir, line 397: Expand this comment on why this is the case.

data/settings/src/main/proto/com/google/jetpackcamera/settings/target_frame_rate.proto
Kimblebee: so `NONE` is the actual user/app defined choice to not use a target frame rate.
`UNDEFINED` is there to follow proto best practices for enums, to leave a default unspecified value to ensure consistent expected behavior when adding new values.
https://protobuf.dev/programming-guides/dos-donts/#unspecified-enum
>Enums should include a default `FOO_UNSPECIFIED` value as the first value in the declaration . When new values are added to a proto2 enum, old clients will see the field as unset and the getter will return the default value or the first-declared value if no default exists 

I could add a comment to the enum files with that explanation, and maybe rename undefined to remove the ambiguity between none?

data/settings/src/main/proto/com/google/jetpackcamera/settings/jca_settings.proto
Kimblebee: you're right, this could just be an int. i think when i was originally writing the datastore values i wanted to keep it congruent with the data model used for `CameraAppSettings`. 

I think alternatively, since target frame rate is a range, the datastore can hold onto two ints to represent the range's upper and lower values. this isn't necessary right now because the implementation is only looking at fixed values 15, 30, or 60 fps, but I think it'd still be something to consider if we'd ever wanted to dynamically change the options available on the setting.

data/settings/src/main/proto/com/google/jetpackcamera/settings/target_frame_rate.proto
Kimblebee: following this [recommended practice](https://protobuf.dev/programming-guides/dos-donts/#unspecified-enum) for proto enums by setting an unspecified value for the first enum ordinal. 
>When new values are added to a proto2 enum, old clients will see the field as unset and the getter will return the default value or the first-declared value if no default exists . For consistent behavior with [proto enums](https://protobuf.dev/programming-guides/proto2/#enum), the first declared enum value should be a default FOO_UNSPECIFIED value and should use tag 0.

I _could_ rename it or add a comment to address the ambiguity, but based on [this comment](https://github.com/google/jetpack-camera-app/pull/86#discussion_r1518444032) I'll likely scrap that enum anyways.

feature/settings/src/main/res/values/strings.xml
Kimblebee: let me know if the update is how you imagined the changes

data/settings/src/main/proto/com/google/jetpackcamera/settings/target_frame_rate.proto
temcguir: Thanks for the context. I'll make sure to add UNDEFINED for all my proto enums.

But yes, I don't think this is needed; we could just use integers. So in that case you can remove this.

data/settings/src/main/java/com/google/jetpackcamera/settings/model/TargetFrameRate.kt
```diff
@@ -0,0 +1,46 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.jetpackcamera.settings.model
+
+import android.util.Range
+
+enum class TargetFrameRate(val range: Range<Int>) {
```
temcguir, line 20: We shouldn't need this enum. We could just use `Int` directly. If you need to call out specific frame rates in code, you can put them in constants.

feature/settings/src/main/res/values/strings.xml
```diff
@@ -83,5 +83,24 @@
...
+
+
+    <!-- Target Fps setting strings -->
+    <string name="fps_title">Set Frame Rate</string>
+
+    <string name="fps_description_unavailable">No fixed frame rates supported</string>
+    <string name="fps_description_none">Auto Frame Rate</string>
+    <string name="fps_description_15">15 fps </string>
+    <string name="fps_description_30">30 fps</string>
+    <string name="fps_description_60">60 fps</string>
```
temcguir, line 97: You should only need a single format string for all of these:
`<string name="fps_description">%d fps</string>`

You can then get get the string with: `stringResources(R.string.fps_description, integerFps)`

data/settings/src/main/java/com/google/jetpackcamera/settings/SettingsRepository.kt
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2023 The Android Open Source Project
+ * Copyright (C) 2023-2024 The Android Open Source Project
```
temcguir, line 2: nit: just 2023

feature/settings/src/main/java/com/google/jetpackcamera/settings/ui/SettingsComponents.kt
```diff
@@ -253,6 +256,69 @@ fun CaptureModeSetting(currentCaptureMode: CaptureMode, setCaptureMode: (Capture
...
+                    onClick = { setTargetFps(TargetFrameRate.TARGET_FPS_30) },
+                    enabled = supportedFps.contains(30)
+                )
+
+                SingleChoiceSelector(
+                    text = stringResource(id = R.string.fps_selector_60),
+                    selected = currentTargetFps == TargetFrameRate.TARGET_FPS_60,
+                    onClick = { setTargetFps(TargetFrameRate.TARGET_FPS_60) },
+                    enabled = supportedFps.contains(60)
+                )
```
temcguir, line 316: You can just have a loop here to add the values from supportedFps:

```
supportedFps.forEach { fpsOption ->
  SingleChoiceSelector(
     text = "%d".format(fpsOption), // Should localize automatically
     selected = currentTargetFps == fpsOption,
     onClick = { setTargetFps(fpsOption) }
     // enabled = supportedFps.contains(30) <--- not needed since the option won't exist if its not supported
  )
```

Kimblebee: Thanks! but i think i'd still like to keep the disabled out fps options. It follows the pattern of the other settings having greyed out options and I think having a visible indicator is nice.
I think its nice to have a visible indication of "we checked for this and its unavailable"
Instead of only telling "this is what is available to you"

temcguir: SGTM. This is probably better for now anyways since it will make us have to think harder about the layout if we add many more choices.

temcguir (approved):

PR #87, opened 2023-12-11
Create a Toast message with compose

Kimblebee:
🍞 _it is time ⏰ for toast (yum)_ 🍞

Here are the notable things about the change:
Create the option to have toasts in the preview screen!
in this PR it displays a toast message when the image capture is successful or unsuccessful

New **PreviewUiState** property -`toastMessageToShow : ToastMessage`, which is default to null
New class - `ToastMessage.kt`, which holds the information necessary to create a toast 
New composable component - `ShowToast`, takes in a `ToastMessage` and `onToastShown()` (which resets toast ui state to null). Whats special about it is that it has a TestTag and a changeable description, so you can target the presence of a toast or a specific toast when testing.
 
Implementation in `PreviewScreen` -  if the toast property is not null for preview ui state, show composable and reset ui state.

image

note:
* you can still click the capture button through the toast
* starting from android 12 the toast messages include the app icon

### DIFF
--- a/feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
+++ b/feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
@@ -58,7 +58,6 @@ import androidx.compose.ui.semantics.semantics
 import androidx.compose.ui.unit.dp
 import androidx.compose.ui.unit.sp
 import com.google.jetpackcamera.feature.preview.R
-import com.google.jetpackcamera.feature.preview.ToastMessage
 import com.google.jetpackcamera.feature.preview.VideoRecordingState
 import com.google.jetpackcamera.settings.model.AspectRatio
 import com.google.jetpackcamera.viewfinder.CameraPreview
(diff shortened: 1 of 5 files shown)


feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
```diff
@@ -60,8 +65,27 @@ import kotlinx.coroutines.CompletableDeferred
...
+ * @param onToastShown called once the Toast has been displayed.
+ */
+@Composable
+fun ShowToast(modifier: Modifier = Modifier, toastMessage: ToastMessage, onToastShown : () -> Unit){
+    Box(
+        modifier.testTag(toastMessage.testTag).semantics { contentDescription = toastMessage.testDesc }
+    ) {
+        Toast.makeText(LocalContext.current, toastMessage.message, toastMessage.toastLength).show()
+        onToastShown()
+    }
```
temcguir, line 82: I think this is just creating an empty `Box`, since Toasts aren't Composables. That empty box will get removed almost immediately since `onToastShown` will then cause recomposition to no longer draw the Box.  But maybe that's sufficient for UiAutomator to catch it?

Just checking if that's the intended behavior.

temcguir (approved):

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
Kimblebee: yeah the intention was creating an "invisible box" to both be the vehicle to launch a toast message and also to identify its presence (with the tags). 

I'm not 100% sure if the UiAutomator will be able to catch it because of how fast it is, but i could include a delay  before it calls `ShowToast()` if it ends up being necessary to detect

temcguir: Since you're not 100% sure, it would be good to add a simple UiAutomator test for the `ShowToast` composable. Actually it's a good idea to add a test even if you are 100% sure, but especially if you have doubts.

PR #88, opened 2023-12-11
External Intent Image Capture

davidjiagoogle:
ITS broadcasts an android.media.action.IMAGE_CAPTURE intent with ContentValues extra to launch JCA and go through image capture flow. Refer to https://docs.google.com/document/d/1RWv5G-DVtw6_GAusTlLLQmnfmYgiiWUsTaxTdhprx2o/edit. This PR implements the external intent flow in JCA. While the intent that launches the app must strictly be android.media.action.IMAGE_CAPTURE, and the extra must be in the format of ContentValues, the source of the intent is generic and not limited to ITS.

In this PR:
- Add a BroadcastReceiver(ImageCaptureReceiver) to receive IMAGE_CAPTURE intent.
- ImageCaptureReceiver uses the extra ContentValues (if any) to launch the MainActivity
- When user goes through the image capture flow, if the app is launched externally and contains extra values, JCA will use these values to save the image. If not, JCA will use default values.
- If launched externally, after image capture succeeds or fails, JCA will finish with the proper activity result (OK or CANCELED).


app/build.gradle.kts
```diff
@@ -73,6 +73,7 @@ android {
 }
 
 dependencies {
+    implementation(project(":domain:camera"))
```
yasith, line 76: nit: move this to the bottom, similar to the other build.gradle.kts files for internal project deps

app/src/main/java/com/google/jetpackcamera/MainActivity.kt
```diff
@@ -65,6 +72,19 @@ class MainActivity : ComponentActivity() {
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
+        val externalContentValues: ContentValues? =
+            if (intent.extras == null) null
```
yasith, line 76: nit: add the braces for if 

https://developer.android.com/kotlin/style-guide#braces

app/src/main/java/com/google/jetpackcamera/MainActivity.kt
```diff
@@ -65,6 +72,19 @@ class MainActivity : ComponentActivity() {
...
+            else if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
+                intent.extras!!.getParcelable(
+                    MediaStore.EXTRA_OUTPUT, ContentValues::class.java
+                )
+            } else {
+                intent.extras!!.getParcelable(MediaStore.EXTRA_OUTPUT)
+            }
+        val shouldFinishAfterCapture =
```
yasith, line 84: nit: braces

app/src/main/java/com/google/jetpackcamera/MainActivity.kt
```diff
@@ -101,7 +121,27 @@ class MainActivity : ComponentActivity() {
...
+                                        }
+                                    }
+
+                                    override fun onError() {
+                                        if (shouldFinishAfterCapture) {
+                                            setResult(RESULT_CANCELED)
+                                            finish()
+                                        }
+                                    }
+
```
yasith, line 142: nit: empty line

domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/TakePictureCallback.kt
```diff
@@ -0,0 +1,25 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.jetpackcamera.domain.camera
+
+import android.net.Uri
+
+interface TakePictureCallback {
```
yasith, line 21: It would be more kotlin idiomatic to have a sealed interface with two classes for the different results.

CombinedSurface.kt has an example

domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
```diff
@@ -130,25 +134,65 @@ constructor(
...
+                } else if (value is String) {
+                    eligibleContentValues.put(key, value)
+                } else if (value is Int) {
+                    eligibleContentValues.put(key, value)
+                } else if (value is Boolean) {
+                    eligibleContentValues.put(key, value)
+                } else if (value is Long) {
+                    eligibleContentValues.put(key, value)
+                } else if (value is Float) {
+                    eligibleContentValues.put(key, value)
```
yasith, line 180: This seems a bit weird, is this how it's supposed to be? If so that's fine

feature/preview/src/test/java/com/google/jetpackcamera/feature/preview/PreviewViewModelTest.kt
```diff
@@ -63,7 +63,7 @@ class PreviewViewModelTest {
     fun captureImage() = runTest(StandardTestDispatcher()) {
         val surfaceProvider: SurfaceProvider = mock()
         previewViewModel.runCamera(surfaceProvider)
-        previewViewModel.captureImage()
+        previewViewModel.captureImage(null)
```
yasith, line 66: Looks like this takes in 3 arguments?

domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
davidjiagoogle: Yeah, unfortunately there is no other way to copy ContentValues except iterating through every key-value pair, and this has to be done because put doesn't accept generic object type

temcguir: Could you just use [`ContentValues.putAll(ContentValues)`](https://developer.android.com/reference/android/content/ContentValues#putAll(android.content.ContentValues))?

davidjiagoogle:
Addressed comments and added two device tests

app/src/main/java/com/google/jetpackcamera/receiver/ImageCaptureReceiver.kt
```diff
@@ -0,0 +1,47 @@
...
+import android.content.Intent
+import android.os.Build
+import android.provider.MediaStore
+import com.google.jetpackcamera.MainActivity
+
+class ImageCaptureReceiver : BroadcastReceiver() {
+    override fun onReceive(context: Context, intent: Intent) {
+        val activityIntent = Intent(context, MainActivity::class.java)
+        val externalContentValues = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
+            intent.getParcelableExtra(MediaStore.EXTRA_OUTPUT, ContentValues::class.java)
```
temcguir, line 30: Based on the documentation for [ACTION_TAKE_PICTURE](https://developer.android.com/reference/android/provider/MediaStore#ACTION_IMAGE_CAPTURE), the standard contract is to take a `Uri` as the `EXTRA_OUTPUT`, not a `ContentValues`.

This means we'll want to use the `OutputStream` version of `ImageCapture.OutputFileOptions` where we open the Uri with [`ContentResolver.openOutputStream(Uri)`](https://developer.android.com/reference/android/content/ContentResolver.html#openOutputStream(android.net.Uri))

temcguir (changes requested):
This needs to be updated to use the standard contract for capturing an image from ACTION_IMAGE_CAPTURE. By following that contract, JCA would also be able to act as a default camera app for capturing images.

domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
temcguir: Actually this should no longer be needed if we're saving to a user-provided `Uri`.

app/src/androidTest/java/com/google/jetpackcamera/ImageCaptureDeviceTest.kt
```diff
@@ -0,0 +1,112 @@
...
+                MediaStore.EXTRA_OUTPUT,
+                contentValues
+            )
+        activityScenario = ActivityScenario.launch(intent)
+        uiDevice.waitForIdle(5000)
+        uiDevice.findObject(By.res("CaptureButton")).click()
+        Thread.sleep(2000)
+        directory = File(dirPath)
+        val pictureTaken = (files.size + 1) == directory.listFiles().size
+        assert(pictureTaken)
```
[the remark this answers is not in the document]
davidjiagoogle: It's still good to verify that an extra image was taken

app/src/main/java/com/google/jetpackcamera/receiver/ImageCaptureReceiver.kt
davidjiagoogle: This actually can be done via ActivityScenario. ActivityScenario.result will WAIT for the activity that was launched from ActivityScenario.launchActivityForResult(), and we can verify the result code then. Check the updated ImageCaptureDeviceTest

domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
```diff
@@ -144,27 +150,85 @@ constructor(
...
+        } else {
+            try {
+                val outputStream = contentResolver.openOutputStream(imageCaptureUri)
+                if (outputStream != null) {
+                    outputFileOptions =
+                        OutputFileOptions.Builder(
+                            contentResolver.openOutputStream(imageCaptureUri)!!
+                        ).build()
+                } else {
+                    val e = RuntimeException("Output stream is null.")
```
temcguir, line 176: From the `ContentResolver` [documentation](https://developer.android.com/reference/android/content/ContentResolver#openOutputStream(android.net.Uri)), it looks like this happens when the "provider has recently crashed", so maybe put that in the exception message.

app/src/androidTest/java/com/google/jetpackcamera/ImageCaptureDeviceTest.kt
```diff
@@ -0,0 +1,160 @@
...
+        assert(result == Activity.RESULT_OK)
+        val pictureTaken = (files.size + 1) == directory.listFiles().size
+        assert(pictureTaken)
+        if (pictureTaken) {
+            deleteFilesInDirAfterTimestamp(timeStamp)
+        }
+    }
+
+    @Test
+    fun image_capture_external_no_uri() = runTest {
```
temcguir, line 105: According to the [contract](https://developer.android.com/reference/android/provider/MediaStore#ACTION_IMAGE_CAPTURE), in this case "a small sized image is returned as a Bitmap object in the extra field".

At a minimum, you should create a bug/TODO to implement that in a follow-up PR.

app/src/main/java/com/google/jetpackcamera/MainActivity.kt
```diff
@@ -101,13 +107,67 @@ class MainActivity : ComponentActivity() {
...
+        ) {
+            null
+        } else if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
+            intent.extras!!.getParcelable(
+                MediaStore.EXTRA_OUTPUT,
+                Uri::class.java
+            )
+        } else {
+            @Suppress("DEPRECATION")
+            intent.extras!!.getParcelable(MediaStore.EXTRA_OUTPUT)
```
temcguir, line 153: It looks like you should be using `getParcelable(EXTRA_OUTPUT, Uri::class.java)` here on API 33+.

app/src/main/java/com/google/jetpackcamera/MainActivity.kt
```diff
@@ -101,13 +107,67 @@ class MainActivity : ComponentActivity() {
...
                     }
                 }
             }
         }
     }
+
+    private fun getImageCaptureUriIfExists(): Uri? {
+        val isExternalImageCapture = isExternalImageCapture()
+        val uri = if (!isExternalImageCapture || intent.extras == null ||
+            !intent.extras!!.containsKey(MediaStore.EXTRA_OUTPUT)
```
temcguir, line 143: The [contract](https://developer.android.com/reference/android/provider/MediaStore#ACTION_IMAGE_CAPTURE) also says the URI might come from `ClipData` set on the intent, rather than from `EXTRA_OUTPUT`.

app/src/androidTest/java/com/google/jetpackcamera/ImageCaptureDeviceTest.kt
```diff
@@ -0,0 +1,160 @@
...
+        val timeStamp = System.currentTimeMillis()
+        var directory = File(DIR_PATH)
+        val files = directory.listFiles()
+        val launchIntent = Intent()
+        launchIntent.setAction(MediaStore.ACTION_IMAGE_CAPTURE)
+        launchIntent.putExtra(MediaStore.EXTRA_OUTPUT, TEST_URI)
+        launchIntent.setComponent(
+            ComponentName("com.google.jetpackcamera", "com.google.jetpackcamera.MainActivity")
+        )
+        activityScenario = ActivityScenario.launchActivityForResult(launchIntent)
```
temcguir, line 85: The [documentation](https://developer.android.com/reference/androidx/test/core/app/ActivityScenario#launchActivityForResult(android.content.Intent)) for this method says we should be using `androidx.activity.result.ActivityResult` instead of this method and links to https://developer.android.com/training/basics/intents/result#test. Since JCA is supposed to be following best practices, we should be using the methods suggested there.

app/build.gradle.kts
```diff
@@ -124,6 +125,9 @@ dependencies {
     // Settings Screen
     implementation(project(":feature:settings"))
 
+    // Domain Camera
+    implementation(project(":domain:camera"))
```
temcguir, line 129: I'm not sure if the app should have a direct dependency on `domain:camera`. Maybe we need a layer of abstraction? @yasith do you know?

app/src/main/java/com/google/jetpackcamera/MainActivity.kt
```diff
@@ -101,13 +107,67 @@ class MainActivity : ComponentActivity() {
...
+
+                                        is CameraUseCase.ImageCaptureEvent.ImageCaptureError -> {
+                                            if (shouldFinishWithResult) {
+                                                setResult(RESULT_CANCELED)
+                                                finish()
+                                            }
+                                        }
+                                    }
+                                }
+                            )
```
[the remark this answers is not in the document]
davidjiagoogle: woulnd't this violate the  MediaStore contract? If we use this mode class, then it would also encapsulate the uri that's supposed to be at EXTRA_OUTOUT or clipData?

app/src/main/java/com/google/jetpackcamera/MainActivity.kt
```diff
@@ -101,13 +107,67 @@ class MainActivity : ComponentActivity() {
...
+                                        is CameraUseCase.ImageCaptureEvent.ImageSaved -> {
+                                            if (shouldFinishWithResult) {
+                                                setResult(RESULT_OK)
+                                                finish()
+                                            }
+                                        }
+
+                                        is CameraUseCase.ImageCaptureEvent.ImageCaptureError -> {
+                                            if (shouldFinishWithResult) {
+                                                setResult(RESULT_CANCELED)
```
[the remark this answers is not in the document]
davidjiagoogle: I'm not sure that's the best UX. I think it could cause a lot of confusion by potentially making errors happening repeatedly within JCA. With the current UX, the user would be forced to examine the RESULT_CANCEL within there own app. I think with this logic, would could consider incorporating the error message or exception in the result intent.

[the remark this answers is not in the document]
temcguir: Agreed that it's not the best UX, but it may be better than the alternative.
`ACTION_IMAGE_CAPTURE` does not specify any result codes or any extras returned in the result that would denote an error, so we're stuck with the `RESULT_OK` and `RESULT_CANCELED`. Sure we could put an error message or exception into the result, but since it isn't documented anywhere for `ACTION_IMAGE_CAPTURE`, apps can't reasonably investigate the results from an intent because they don't know what to look for. Remember, JCA is only one of many apps that can handle `ACTION_IMAGE_CAPTURED`, so apps won't know specifics of what is returned unless it is part of the documentation.

Since *we* know what the error message is and what it means, we have enough context to show the error in the UI in JCA. It's also possible it could be a transitive error, in which case trying again would be faster if we stayed in the app. If the user navigates away, the `RESULT_CANCELED` is actually the correct error code for the calling app, since it was the user's decision to cancel the request.

Note this is how the AOSP camera app handles this scenario [here](https://cs.android.com/android/platform/superproject/main/+/main:packages/apps/Camera2/src/com/android/camera/PhotoModule.java;l=1289;drc=96ae7ed133f130b84cfa6e71ab6b5c23f806f046).

app/src/main/java/com/google/jetpackcamera/MainActivity.kt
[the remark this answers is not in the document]
davidjiagoogle: Ah I see what you are saying. The mode is generated in mainActivity from the extra (or lack of) then passed into JCAApp. Yeah we could certainly do it that way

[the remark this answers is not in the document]
temcguir: It shouldn't change anything about how we fulfill the contract, it's just a way to organize the code without having to add new arguments to the `JcaApp` constructor for every intent `ACTION_*` that we handle.

You could even just encapsulate the arguments that you already have in the `ExternalImageCaptureMode` subclass:
```
sealed interface JcaAppMode

class StandardMode() : JcaAppMode

data class ExternalImageCaptureMode(
    val contentResolver: ContentResolver,
    val imageCaptureUri: Uri?,
    val onImageCapture: (ImageCaptureEvent) -> Unit
) : JcaAppMode
```

Then the constructor for JcaApp is:
```
class JcaApp(
    onPreviewViewModel = { previewViewModel = it },
    val appMode: JcaAppMode = StandardMode
)
```

Then for the `captureImage` call, you can do:
```
when (appMode) {
  is StandardMode -> captureImage() // previous "captureImage" overload with no args
  is ExternalImageCaptureMode ->
      captureImage(appMode.contentResolver, appMode.imageCaptureUri, appMode.onImageCapture)
}
```

Eventually we can use the `JcaAppMode` to determine how the UI looks. For instance, in `ExternalImageCapture` mode, there is no need to allow users to capture video, so we may show a special icon for the capture button that makes it more clear that video is disabled.

app/build.gradle.kts
davidjiagoogle: ImageCaptureEvent can be moved into feature:preview to prevent this line of change

app/src/main/java/com/google/jetpackcamera/MainActivity.kt
davidjiagoogle: this is for version below 33 where that api was unavailable. Our min version is 21

app/src/androidTest/java/com/google/jetpackcamera/ImageCaptureDeviceTest.kt
davidjiagoogle: I read through https://developer.android.com/training/basics/intents/result#test a couple times and I don't think it can be applied in our test. Our test is lacking a start activity to call launchActivityForResult(), which is what ActivityScenario is for. In the documentation, however, the new way to launch an activity for result is still predicated on having a starting ComponentActivity, with which you can call registerForActivityResult() to get a Launcher to launch an activity externally. In the Test section, the documentation simply shows a way to mock/override a result being returned by the launched activity. 

Since we are trying to do a device test here, we don't need to override the result. Instead, we need a way to launch JCA externally, and the examples in the documentation don't help us. I also couldn't find any examples on how to obtain a launcher in a test in other places

davidjiagoogle: done

app/src/androidTest/java/com/google/jetpackcamera/ImageCaptureDeviceTest.kt
```diff
@@ -0,0 +1,176 @@
...
+        )
+
+    private val instrumentation = InstrumentationRegistry.getInstrumentation()
+    private var activityScenario: ActivityScenario<MainActivity>? = null
+    private val uiDevice = UiDevice.getInstance(instrumentation)
+
+    @Test
+    fun image_capture_external() = run {
+        val timeStamp = System.currentTimeMillis()
+        val launchIntent = Intent()
```
temcguir, line 59: You can replace this with the contract:
```
val contract = object: ActivityResultContracts.TakePicture() {
  override fun createIntent(context: Context, uri: Uri) {
    return super.createIntent(context, uri).apply {
      component = ComponentName("com.google.jetpackcamera", "com.google.jetpackcamera.MainActivity")
    }
  }
}
```

Then pass this in to `.register("key", contract)`. 

You could even make the above contract into a constant (`TEST_CONTRACT`, or `TAKE_PICTURE_CONTRACT`).

app/src/androidTest/java/com/google/jetpackcamera/ImageCaptureDeviceTest.kt
```diff
@@ -0,0 +1,176 @@
...
+
+    private fun deleteFilesInDirAfterTimestamp(timeStamp: Long) {
+        for (file in File(DIR_PATH).listFiles()) {
+            if (file.lastModified() > timeStamp) {
+                file.delete()
+            }
+        }
+    }
+
+    private fun getTestRegistry(launch: () -> Int): ActivityResultRegistry {
```
temcguir, line 151: If you change the signature to `launch: (Intent) -> Instrumentation.ActivityResult`, then you can make everything go through the contract.

You would need to change the `onLaunch` code below to:
```
override fun <I, O> onLaunch(
   requestCode: Int,
   contract: ActivityResultContract<I, O>,
   input: I,
   options: ActivityOptionsCompat?
   ) {
    val launchIntent = contract.getIntent(ApplicationProvider.getApplicationContext<Context>(), input)
    val result: ActivityResult = launch(launchIntent)
    dispatchResult(requestCode, result.resultCode, result.resultData)
  }
```

app/src/main/AndroidManifest.xml
```diff
@@ -25,6 +25,10 @@
         android:required="false" />
 
     <uses-permission android:name="android.permission.CAMERA" />
+    <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE"/>
+    <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE" />
+    <uses-permission android:name="com.google.android.c2dm.permission.RECEIVE" />
```
temcguir, line 30: What is this permission used for?

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewViewModel.kt
```diff
@@ -235,6 +237,45 @@ class PreviewViewModel @Inject constructor(
...
     }
 
+    fun captureImage(
+        contentResolver: ContentResolver,
+        imageCaptureUri: Uri?,
+        onImageCapture: (ImageCaptureEvent) -> Unit
+    ) {
+        Log.d(TAG, "captureImageWithUri")
+        viewModelScope.launch {
+            traceAsync(IMAGE_CAPTURE_TRACE, 0) {
```
temcguir, line 247: nit: You should either add a benchmark to `ImageCaptureLatencyBenchmark` or create a TODO to add one.

domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
```diff
@@ -163,10 +170,75 @@ constructor(
...
+                        imageDeferred.complete(outputFileResults)
+                    }
+
+                    override fun onError(exception: ImageCaptureException) {
+                        Log.d(TAG, "takePicture onError: $exception")
+                        imageDeferred.completeExceptionally(exception)
+                    }
+                }
+            )
+            imageDeferred.await()
```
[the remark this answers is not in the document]
davidjiagoogle: Oh I see. Changed to throwing exceptions. This also guarantees outputFileOptions not being null

app/src/main/AndroidManifest.xml
davidjiagoogle: leftover from broadcaster receiver. Removed.

domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
```diff
@@ -177,11 +177,11 @@ constructor(
     override suspend fun takePicture(contentResolver: ContentResolver, imageCaptureUri: Uri?) {
         val imageDeferred = CompletableDeferred<ImageCapture.OutputFileResults>()
         val eligibleContentValues = getEligibleContentValues()
-        var outputFileOptions: OutputFileOptions? = null
+        val outputFileOptions: OutputFileOptions
```
temcguir, line 180: Does this need an `=` at the end of the line?

temcguir (approved):
Thanks for making all of these changes!

domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
davidjiagoogle: No. It's set on line 189. All other cases lead to exception being thrown

PR #94, opened 2024-01-12
Kim/stabilization UI logic

Kimblebee:
changes for [stabilization setting PR](https://github.com/google/jetpack-camera-app/pull/83) , i thought that it may be better to just put it into a separate PR for visibility of the changes.

here are the big changes to come of it:
* `CameraXCameraUseCase` sets preview/video stabilization support information to datastore
* cameraAppSettings condenses datastore stabilization support info to an enum to make it easier to use around the app
* disables stabilization setting if unsupported by device
* disables `On` option if preview stabilization is unavailable
* disables `High Quality`option if video stabilization is unavailable
* disables entire setting if neither preview nor video stabilization is supported
* display an icon to preview screen when stabilization is active

image 
image


domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
```diff
@@ -369,19 +377,8 @@ constructor(
     }
 
     private fun createPreviewUseCase(): Preview {
-        val availableCameraInfo = cameraProvider.availableCameraInfos
-        val cameraSelector = if (isFrontFacing) {
-            CameraSelector.DEFAULT_FRONT_CAMERA
-        } else {
-            CameraSelector.DEFAULT_BACK_CAMERA
-        }
-        val isPreviewStabilizationSupported =
```
temcguir, line 378: Why aren't we checking preview stabilization separately from video stabilization? These two stabilization modes should account for:
Preview -> Stabilization ON
Video -> Stabilization HIGH QUALITY

data/settings/src/main/java/com/google/jetpackcamera/settings/test/FakeJcaSettingsSerializer.kt
```diff
@@ -31,6 +36,14 @@ class FakeJcaSettingsSerializer(
...
         .setDarkModeStatus(DarkMode.DARK_MODE_SYSTEM)
         .setDefaultFrontCamera(false)
+        .setBackCameraAvailable(true)
+        .setFrontCameraAvailable(true)
+        .setFlashModeStatus(FlashMode.FLASH_MODE_OFF)
+        .setAspectRatioStatus(AspectRatio.ASPECT_RATIO_NINE_SIXTEEN)
+        .setCaptureModeStatus(CaptureMode.CAPTURE_MODE_MULTI_STREAM)
+        .setStabilizePreview(PreviewStabilization.PREVIEW_STABILIZATION_UNDEFINED)
+        .setStabilizeVideo(VideoStabilization.VIDEO_STABILIZATION_UNDEFINED)
+        .setStabilizationSupported(false)
```
temcguir, line 46: Instead of a `boolean`, can this be a list of the supported stabilization modes such as `ON` and `HIGH_QUALITY`?

feature/settings/src/main/java/com/google/jetpackcamera/settings/ui/SettingsComponents.kt
```diff
@@ -272,24 +272,35 @@ private fun getStabilizationStringRes(
...
+    )
     BasicPopupSetting(
         title = stringResource(R.string.video_stabilization_title),
         leadingIcon = null,
-        description = stringResource(
-            id = getStabilizationStringRes(
-                previewStabilization = currentPreviewStabilization,
-                videoStabilization = currentVideoStabilization
+        enabled = isStabilizationSupported,
+        description = when (isStabilizationSupported) {
```
temcguir, line 294: nit: instead of `when` with `true` and `false` branches here, just use an `if` statement.

data/settings/src/main/java/com/google/jetpackcamera/settings/test/FakeJcaSettingsSerializer.kt
temcguir: Ok- storing them as separate booleans is fine. I was just suggesting a list since it's easier to pass around as a single variable. I don't know if we need to convert them to a separate enum. In the settings composable can we just do something like:

```
Column(Modifier.selectableGroup()) {

  if (isPreviewStabilizationSupported) {
    // on selector
    SingleChoiceSelector(...)
  }

  if (isVideoStabilizationSupported) {
    // high quality selector
    SingleChoiceSelector(...)
  }

  // none selector
  SingleChoiceSelector(...)
}

File: feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewScreen.kt [hunk 4]
```diff
@@ -184,19 +188,35 @@ fun PreviewScreen(
...
+                            text = stringResource(
+                                when (previewUiState.currentCameraSettings.captureMode) {
+                                    CaptureMode.SINGLE_STREAM -> R.string.capture_mode_single_stream
+                                    CaptureMode.MULTI_STREAM -> R.string.capture_mode_multi_stream
+                                }
+                            )
                         )
-                    )
+                        if (previewUiState.currentCameraSettings.supportedStabilizationMode !=
+                            SupportedStabilizationMode.UNSUPPORTED &&
```
temcguir, line 209: Shouldn't it be the case that if the current stabilization mode is `UNSUPPORTED` then the stabilization mode can't be set to `ON`? If so, then we can get rid of the check for `UNSUPPORTED`, we should just go with whatever the stabilization mode is set to.

feature/settings/src/main/java/com/google/jetpackcamera/settings/ui/SettingsComponents.kt
```diff
@@ -271,37 +272,40 @@ private fun getStabilizationStringRes(
...
-                    color = MaterialTheme.colorScheme.onPrimaryContainer
-                )
-
                 Spacer(modifier = Modifier.height(10.dp))
 
                 // on selector
                 SingleChoiceSelector(
                     text = stringResource(id = R.string.stabilization_selector_on),
                     secondaryText = stringResource(id = R.string.stabilization_selector_on_info),
+                    enabled = supportedStabilizationMode == SupportedStabilizationMode.FULL,
```
temcguir, line 308: Do you need to do something similar for the "high quality selector" below?

feature/settings/src/main/java/com/google/jetpackcamera/settings/ui/SettingsComponents.kt
```diff
@@ -271,37 +272,40 @@ private fun getStabilizationStringRes(
...
                     selected = (currentPreviewStabilization == Stabilization.ON) &&
                         (currentVideoStabilization != Stabilization.OFF),
                     onClick = {
```
temcguir, line 311: Technically setting both VideoStabilization and PreviewStabilization to `ON` will work here, but only because CameraX will internally ignore the VideoStabilization part.

Let's set VideoStabilization to `UNDEFINED` to better represent what's actually going on under the hood and so we don't confuse developers into thinking they have to set both to `ON` to get preview stabilization.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewScreen.kt
```diff
@@ -184,19 +188,35 @@ fun PreviewScreen(
...
+                            previewUiState.currentCameraSettings.videoCaptureStabilization ==
```
temcguir, line 210: I know we're using the same icon here for `ON` and `HIGH_QUALITY`, but we should at least change the `contentDescription` to specify whether `ON` or `HIGH_QUALITY` is selected. In that case, you'll also need to check `previewStabilization`.

feature/settings/src/main/java/com/google/jetpackcamera/settings/ui/SettingsComponents.kt
Kimblebee: Not necessary for the UI's current implementation. if the device can't support video stabilization at all, then the popup  @won't be able to open in the first place. the whole button itself would be greyed out and unclickable.
image

temcguir: Aren't video stabilization and preview stabilization mutually exclusive?

If preview stabilization is supported, but video stabilization isn't, then you should be able to choose between `ON` and `OFF`.

Alternatively if preview stabilization isn't supported but video stabilization is, you should be able to select between `HIGH_QUALITY` and `OFF`.

Kimblebee: Thats true,  I assumed if the output feed couldn't be stabilized, the user may not want to bother with stabilization settings. I think to avoid confusion with `ON` meaning: "preview and video" OR "preview only if video isn't supported", I could add an extra option/enum `PREVIEW_ONLY`. 

The setting would only be completely disabled if neither preview nor video were supported.

temcguir: > I could add an extra option/enum PREVIEW_ONLY. It can be a hidden option only displayed when the device doesn't support video.

I think there's still some confusion here about preview stabilization. Preview stabilization means that Preview AND Video will be stabilized (it's WYSIWYG, hence why we call it "`ON`"). You don't need to enable video stabilization to stabilize the video, the preview stabilization will be sufficient to stabilize both. When you enable both preview and video stabilization at the same time, the preview stabilization takes precedence and stabilizes both, and the video stabilization setting on the VideoCapture use case will be ignored, hence why its ok to set it to `UNSPECIFIED`. There shouldn't be a `PREVIEW_ONLY` setting since preview stabilization will always apply a (lower quality) stabilization to the video.

> The setting would only be completely disabled if neither preview nor video were supported.

That is the correct behavior. It means it's not possible to stabilize Preview AND Video with preview stabilization. It's also not possible to get high quality video stabilization with just video stabilization.

feature/preview/src/main/res/values/strings.xml
```diff
@@ -23,4 +23,8 @@
 
     <string name="toast_image_capture_success">Image Capture Success</string>
     <string name="toast_capture_failure">Image Capture Failure</string>
+    <string name="stabilization_icon_description_preview_and_video">Preview and video are stabilized</string>
+    <string name="stabilization_icon_description_preview_only">Only Preview is Stabilized</string>
```
temcguir, line 27: This shouldn't be a description. As discussed in other comments, preview can't be stabilized by itself.

data/settings/src/main/java/com/google/jetpackcamera/settings/model/SupportedStabilizationMode.kt
```diff
@@ -0,0 +1,31 @@
...
+    FULL,
+
+    /** Device supports only Video stabilization.*/
+    VIDEO_ONLY,
+
+    /** Device supports only Preview stabilization.*/
+    PREVIEW_ONLY,
+
+    /** Device doesn't support any stabilization.*/
+    UNSUPPORTED
```
temcguir, line 30: This seems like it will become difficult to maintain as separate states if we were to add one more. For instance, if we added another stabilization mode called `SUPER_STABLE`, then we'll need 4 more enums: `SUPER_STABLE_ONLY`, `SUPER_STABLE_AND_VIDEO_ONLY`, `SUPER_STABLE_AND_PREVIEW_ONLY` `PREVIEW_AND_VIDEO_ONLY`. If you add _even more_ modes, the number of enums needed will start to become untenable. 

How about we just make `SupportedStabilizationMode` have `ON` and `HIGH_QUALITY`. We can then pass it around as `val supportedStabilizationModes: List<SupportedStabilizationMode>`. Then in our settings, we disable the entire stabilization setting if `supportedStabilizationModes.isEmpty()`, and for each of the individual choices, `enabled == supportedStabilizationModes.contains(SupportedStabilizationMode.***)`. We could still leave the protos as individual settings, but in LocalSettingsRepository, we just have:

```
private fun getSupportedStabilization(
        previewSupport: Boolean,
        videoSupport: Boolean
    ): List<SupportedStabilizationMode> {
  return buildList {
    if (previewSupport) add(SupportedStabilizationMode.ON)
    if (videoSupport) add(SupportedStabilizationMode.HIGH_QUALITY)
  }
}
```

domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
```diff
@@ -104,7 +104,7 @@ constructor(
     private lateinit var stabilizePreviewMode: Stabilization
     private lateinit var stabilizeVideoMode: Stabilization
     private lateinit var surfaceProvider: Preview.SurfaceProvider
-    private lateinit var supportedStabilizationMode: SupportedStabilizationMode
+    private lateinit var supportedStabilizationMode: List<SupportedStabilizationMode>
```
temcguir, line 107: nit: supportedStabilizationMode**s**

temcguir (approved):

PR #96, opened 2024-01-12
Preview screen layout refactor

Kimblebee:
**Low priority**
when working on adding an icon in the preview screen for [pr #83](https://github.com/google/jetpack-camera-app/pull/83), I noticed it was a bit difficult to "neatly" insert new elements into the top bar; I made a small refactor to `previewScreen` to make it easier to do so. This will facilitate the addition of future potential elements to the preview screen such as a vertical side bar. 

Most important about the change:
* Preview screen buttons/icons overlay layer has more of a grid-relative layout -- no more free-floating elements in boxes 🎈 
* Top row of icons is a 3-segmented row similar to how the bottom row was implemented
* `QuickSettingsScreen` has been decoupled from its toggle button

Screenshots below use colors to show how the overlay is divided
Magenta = parent column
Green/Blue = rows within the column (the center "empty" space between them is currently a placeholder spacer element)
Grays = shows how the rows are divided
Screenshot 2024-01-12 at 9 40 44 AM Screenshot 2024-01-12 at 9 45 58 AM Screenshot 2024-01-12 at 9 53 14 AM

### DIFF
--- a/feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewScreen.kt
+++ b/feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewScreen.kt
@@ -66,6 +66,7 @@ import com.google.jetpackcamera.feature.preview.ui.TestingButton
 import com.google.jetpackcamera.feature.preview.ui.ZoomScaleText
 import com.google.jetpackcamera.feature.quicksettings.QuickSettingsScreen
 import com.google.jetpackcamera.feature.quicksettings.ui.QuickSettingsIndicators
+import com.google.jetpackcamera.feature.quicksettings.ui.ToggleQuickSettingsButton
 import com.google.jetpackcamera.settings.model.CaptureMode
 import kotlinx.coroutines.CompletableDeferred
 import kotlinx.coroutines.awaitCancellation
@@ -88,7 +89,7 @@ fun PreviewScreen(
     val previewUiState: PreviewUiState by viewModel.previewUiState.collectAsState()
 
     val screenFlashUiState: ScreenFlash.ScreenFlashUiState
-        by viewModel.screenFlash.screenFlashUiState.collectAsState()
+            by viewModel.screenFlash.screenFlashUiState.collectAsState()
 
     val lifecycleOwner = LocalLifecycleOwner.current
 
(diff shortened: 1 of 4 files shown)


feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewScreen.kt
```diff
@@ -137,8 +137,21 @@ fun PreviewScreen(
             aspectRatio = previewUiState.currentCameraSettings.aspectRatio,
             deferredSurfaceProvider = deferredSurfaceProvider
         )
-        // overlay
-        Box(
+        // quick settings screen overlay
+        QuickSettingsScreen(
```
yasith, line 141: nit: We should name this QuickSettinsgScreenOverlay, it's better for the names to be self documenting.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewScreen.kt
```diff
@@ -149,122 +162,156 @@ fun PreviewScreen(
...
-                    modifier = Modifier
-                        .testTag(previewUiState.toastMessageToShow!!.testTag),
-                    toastMessage = previewUiState.toastMessageToShow!!,
-                    onToastShown = viewModel::onToastShown
-                )
-            }
+        }
+        // displays toast when there is a message to show
+        if (previewUiState.toastMessageToShow != null) {
+            ShowTestableToast(
```
yasith, line 309: Is the previewUiState.toastMessageToShow mutable? 

Should add a `requireNotNull` for previewUiState.toastMessageToShow

yasith (approved):

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewScreen.kt
Kimblebee: yes, it stays null when there isn't anything to show. when it is populated, the toast is shown and it returns to null. the "testable" part of the toast is an invisible box that is present for as long as the `toastMessageToShow` isn't null.

Thinking about it again, maybe toastMessageToShow should be a list or queue should other units want to call a toast in quick succession... but i don't think toasts should be used frequently anyways since they're a bit invasive on the UI

PR #97, opened 2024-01-12
Kim/fps UI logic

Kimblebee:
changes for [target fps PR](https://github.com/google/jetpack-camera-app/pull/86) , i thought that it may be better to just put it into a separate PR for visibility of the changes.

here are the 3 big changes to come of it:
* camera sets available select frame rates from all available camera infos to datastore as a list of integers
* Settings UI fades/disables fps options unsupported by device
* function to set max fps to datastore also checks to make sure there isn't an out of bounds target fps
image


domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
```diff
@@ -141,10 +142,34 @@ constructor(
...
 
+    private suspend fun updateMaxFps(currentTargetFrameRate: TargetFrameRate) {
+        var maxFps = 5
+        cameraProvider.availableCameraInfos.forEach { e ->
+            val ranges = e.supportedFrameRateRanges
+            Log.d(TAG, "hewwo $ranges")
+
+            val highest = getMaxFps(currentHighestFps = maxFps, supportedFrameRateRanges = ranges)
+            maxFps = maxFps.coerceAtLeast(highest)
+            Log.d(TAG, "uwu $highest")
```
jsaund, line 158: nit: also this log can be removed too

domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
```diff
@@ -141,10 +142,34 @@ constructor(
...
+        }
+        coroutineScope {
+            settingsRepository.updateMaxFrameRate(maxFps, currentTargetFrameRate)
```
jsaund, line 161: I wonder if we can simplify this method to something like:
```
coroutineScope {
    val highestFps = cameraProvider.availableCameraInfos.map { e.supportedFrameRateRanges.upper }.max()
    settingsRepository.updateMaxFrameRate(
        highestFps,
        currentTargetFrameRate
    )
}

File: data/settings/src/main/java/com/google/jetpackcamera/settings/LocalSettingsRepository.kt [hunk 3]
```diff
@@ -129,6 +130,37 @@ class LocalSettingsRepository @Inject constructor(
         }
     }
 
+    override suspend fun updateMaxFrameRate(
+        maxFrameRate: Int,
+        currentTargetFrameRate: TargetFrameRate
```
jsaund, line 135: I couldn't find where `TargetFrameRate` is defined -- is that part of this PR or already landed?

jsaund: Ah nevermind, I see it now

domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
```diff
@@ -141,10 +142,34 @@ constructor(
...
+        updateMaxFps(currentCameraSettings.targetFrameRate)
 
         return availableCameraLens
     }
 
+    private suspend fun updateMaxFps(currentTargetFrameRate: TargetFrameRate) {
+        var maxFps = 5
+        cameraProvider.availableCameraInfos.forEach { e ->
+            val ranges = e.supportedFrameRateRanges
+            Log.d(TAG, "hewwo $ranges")
```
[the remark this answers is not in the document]
Kimblebee: 🤣 😭 sure thing

data/settings/src/main/proto/com/google/jetpackcamera/settings/jca_settings.proto
```diff
@@ -44,5 +44,5 @@ message JcaSettings {
   bool stabilize_video_supported = 11;
   bool stabilize_preview_supported = 12;
   TargetFrameRate target_frame_rate = 13;
-
+  int32 max_frame_rate = 14;
```
temcguir, line 47: The max frame rate doesn't necessarily mean that every frame rate underneath it is supported. We should check that [15, 15], [30, 30] and [60, 60] are supported explicitly, and add them to this proto as a list. So this field would change to:

`repeated int32 supported_frame_rates = 14;`

domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
```diff
@@ -152,6 +152,17 @@ constructor(
         return availableCameraLens
     }
 
+    private suspend fun updateMaxFps(currentTargetFrameRate: TargetFrameRate) {
+        coroutineScope {
+            val maxFps = cameraProvider.availableCameraInfos.maxOf { cameraInfo ->
+                cameraInfo.supportedFrameRateRanges.maxOf { e ->
```
temcguir, line 158: As mentioned in my other comment, we should populate a list of:
`supportedFramerates: List<Int>` where we add it when `listOf(15, 30, 60).contains(e) && (e.upper == e.lower)`.

data/settings/src/main/proto/com/google/jetpackcamera/settings/jca_settings.proto
Kimblebee: I think this raises a question on what exactly the user may expect the buttons on the settings to convey and do. 
I did my current implementation with this in mind: the device has some fps ranges that reach up to 15/30/60. Selecting one of the options will set the device to find a frame rate that best matches it.

I think limiting the options to only reflect an exact match could potentially disable choices on the setting, despite ranges existing that contain the desired frame rate.
Rather than check for an exact match, what do you think about searching for a range that actually **contains** 15/30/60 instead of the current implementation or searching for an exact match?

temcguir: Video frame rates historically are fixed frame rates. This means [15, 15], [30, 30], or [60, 60] (other standard frame rates are [24, 24] and [25, 25]). These will commonly also be supported frame rates listed in device capabilities. If the user selects 60fps, and we set something like [15, 60], and they're in a dark room, then the AE algorithm may actually choose something closer to 15fps. This will seem like broken behavior to the user who is expecting a 60fps video.

If we want to allow variable frame rates, we need to explicitly call it out as something like "Auto" frame rate or "Variable" frame rate. But there isn't a consistent definition of "variable" frame rate across all devices, so for a consistent behavior we'll need to either add more options for the user to select or explicitly call out frame rate ranges.

Let's stick with the standard fixed frame rates for now. If there is a request for a variable frame rate setting, we can think about how to implement it later.

domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
temcguir: If we want to be more explicit about these being fixed frame rates, we could call it `supportedFixedFrameRates`.

data/settings/src/main/java/com/google/jetpackcamera/settings/JcaSettingsSerializer.kt
```diff
@@ -35,7 +35,8 @@ object JcaSettingsSerializer : Serializer<JcaSettings> {
         .setStabilizeVideo(VideoStabilization.VIDEO_STABILIZATION_UNDEFINED)
         .setStabilizePreviewSupported(false)
         .setStabilizeVideoSupported(false)
-        .setMaxFrameRate(1)
+        .clearSupportedFrameRates()
```
temcguir, line 38: Since this is a `newBuilder()`, calling `clearSupportedFrameRates()` is unnecessary.

domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
```diff
@@ -151,12 +151,16 @@ constructor(
...
     private suspend fun updateMaxFps(currentTargetFrameRate: TargetFrameRate) {
         coroutineScope {
-            val maxFps = cameraProvider.availableCameraInfos.maxOf { cameraInfo ->
-                cameraInfo.supportedFrameRateRanges.maxOf { e ->
-                    e.upper
+            var supportedFixedFrameRates = mutableSetOf<Int>()
+            cameraProvider.availableCameraInfos.forEach() { cameraInfo ->
+                cameraInfo.supportedFrameRateRanges.forEach() { e ->
+                    if (listOf(15,30,60).contains(e.upper)
+                        && e.upper == e.lower) {
```
temcguir, line 158: nit: Reverse this to `e.upper == e.lower && listOf(15, 30, 60).contains(e.upper)`.

The comparison operation is much quicker than `contains` and will short-circuit the condition faster when that first comparison is false.

domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
```diff
@@ -151,12 +151,16 @@ constructor(
 
     private suspend fun updateMaxFps(currentTargetFrameRate: TargetFrameRate) {
         coroutineScope {
-            val maxFps = cameraProvider.availableCameraInfos.maxOf { cameraInfo ->
-                cameraInfo.supportedFrameRateRanges.maxOf { e ->
-                    e.upper
+            var supportedFixedFrameRates = mutableSetOf<Int>()
+            cameraProvider.availableCameraInfos.forEach() { cameraInfo ->
+                cameraInfo.supportedFrameRateRanges.forEach() { e ->
+                    if (listOf(15,30,60).contains(e.upper)
```
temcguir, line 157: nit: Pull `listOf(15, 30, 60)` into a constant.

temcguir (approved):

domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
Kimblebee: nice catch, thanks for pointing this out!

temcguir (approved):

PR #110, opened 2024-02-13
Get rid of the complexity around passing the surface provider

JolandaVerhoef:


feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/CameraXViewfinder.kt
```diff
@@ -53,18 +59,25 @@ import kotlinx.coroutines.launch
...
     implementationMode: ImplementationMode = ImplementationMode.PERFORMANCE,
-    onSurfaceProviderReady: (Preview.SurfaceProvider) -> Unit = {}
+    lifecycleOwner: LifecycleOwner = LocalLifecycleOwner.current,
+    onSurfaceProviderReady: (Preview.SurfaceProvider) -> Unit = {},
+    onSurfaceProviderDisposed: () -> Unit = {}
 ) {
+    var surfaceProvider by remember { mutableStateOf<Preview.SurfaceProvider?>(null) }
+
+    LifecycleStartEffect(surfaceProvider, lifecycleOwner) {
+        surfaceProvider?.let { onSurfaceProviderReady(it) }
```
temcguir, line 69: Should we pull `surfaceProvider?.let` outside `LifecycleStartEffect`? It seems we'd be launching an extra side-effect for a `null` `SurfaceProvider` that will be cancelled immediately.

In fact, the more I think about this, maybe we don't even need a `State` variable for `surfaceProvider`. Calling `onSurfaceProviderDisposed()` when the `produceState` block completes or is cancelled seems to make the most sense to me. It wouldn't cause any unnecessary recompositions, and would actually pair `onSurfaceProviderReady` with `onSurfaceProviderDisposed` when the `implementationMode` changes.

temcguir: Thinking about this again, I still think it makes sense from an API perspective for `CameraXViewfinder` to add `onSurfaceProviderDisposed` to the existing `produceState`. However, I think that would only be for API completeness, and I don't think it would be used in JCA in its current state.

Ultimately, I don't think we should be pushing more state, including the lifecycle, into the `CameraXViewfinder` composable. It makes the surrounding code cleaner in the case of JCA, but doesn't necessarily represent what's actually going on: the `SurfaceProvider` lifecycle isn't actually linked to `lifecycleOwner` argument, though it behaves as if it is.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewScreen.kt
```diff
@@ -111,7 +90,9 @@ fun PreviewScreen(
             onStopVideoRecording = viewModel::stopVideoRecording,
             onToggleCaptureMode = viewModel::toggleCaptureMode,
             onToastShown = viewModel::onToastShown,
-            onSurfaceProviderCreated = { deferredSurfaceProvider.complete(it) }
+//            onSurfaceProviderReady = { surfaceProvider = it },
+            onSurfaceProviderReady = viewModel::runCamera,
```
temcguir, line 94: We could probably clean things up just as well by adding some more code to the ViewModel so that `onSurfaceProviderReady` here would just call `viewModel::setSurfaceProvider`.

We would still need a `LifecycleStartEffect` to start/stop the camera, but the `SurfaceProvider` handling would be removed form composition. I still like the idea of starting/stopping the camera being at the `PreviewScreen` level, since we are then saying "as long as the PreviewScreen is showing, the camera should be on"

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewScreen.kt
```diff
@@ -77,20 +70,6 @@ fun PreviewScreen(
...
-        surfaceProvider = deferredSurfaceProvider.await()
-    }
-
-    LifecycleStartEffect(surfaceProvider) {
-        surfaceProvider?.let { viewModel.runCamera(it) }
-        onStopOrDispose {
-            viewModel.stopCamera()
-        }
-    }
-
```
temcguir, line 93: Another option here that would remove recompositions would be:
```
    val surfaceProviderState = remember { MutableStateFlow<SurfaceProvider?>(null) }

    LifecycleStartEffect(Unit) {
        val startedJob = lifecycleScope.launch {
            surfaceProviderState
                .filterNotNull()
                .onCompletion { viewModel.stopCamera() }
                .collect { viewModel.runCamera(it) }
        }

        onStopOrDispose { startedJob.cancel() }
    }
```

and

```
onSurfaceProviderReady = { surfaceProviderState.value = it }
```

This would also allow the `ImplementationMode` to be changed, and would react correctly.

Of course, this logic can be simplified further if we move the SurfaceProvider state into the ViewModel.

PR #111, opened 2024-02-14
Add HDR settings

temcguir:
- Add dynamic range proto and setting
- Add supported dynamic ranges to settings


data/settings/build.gradle.kts
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2023 The Android Open Source Project
+ * Copyright (C) 2023-2024 The Android Open Source Project
```
Kimblebee, line 2: may want to update the copyright years as per #115

Kimblebee (approved):

data/settings/build.gradle.kts
temcguir: Done. Updated this file and the rest of the files I had edited.

PR #108, opened 2024-02-12
[WIP] Improve Compose usage

JolandaVerhoef:

PR #112, opened 2024-02-14
Set SurfaceProvider directly on PreviewViewModel

temcguir:
Rather than wait for a SurfaceProvider to start the camera, it
 is set eagerly on the ViewModel/CameraUseCase

 The old behavior of waiting for the SurfaceProvider, this was 
 masking a bug where the camera may not have finished
 initialization before starting, so starting the camera now 
 explicitly waits for initialization to be finished


temcguir:
@JolandaVerhoef this PR accomplishes something similar to #110, but without the addition of the `LifecycleOwner` to `CameraXViewfinder`. It also now explicitly calls `startCamera`/`stopCamera` using the `LifecycleStartEffect` you introduced in #108. PTAL.

domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/test/FakeCameraUseCase.kt
```diff
@@ -49,6 +50,9 @@ class FakeCameraUseCase(
     var recordingInProgress = false
 
     var isLensFacingFront = false
+
+    private var surfaceProvider: Preview.SurfaceProvider? = null
```
JolandaVerhoef, line 54: Not used

domain/camera/src/test/java/com/google/jetpackcamera/domain/camera/test/FakeCameraUseCaseTest.kt
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2023 The Android Open Source Project
+ * Copyright (C) 2023-2024 The Android Open Source Project
```
JolandaVerhoef, line 2: Replace with just 2024

JolandaVerhoef (approved):

domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/test/FakeCameraUseCase.kt
temcguir: Good catch. Removed.

domain/camera/src/test/java/com/google/jetpackcamera/domain/camera/test/FakeCameraUseCaseTest.kt
temcguir: This is spotless's doing. I'm going to turn off this behavior in a separate PR and we'll stick with the inception year in each file (as is normally done in Android).

PR #113, opened 2024-02-15
Upgrade dependency versions and cleanup

temcguir:
Upgrades CameraX version so it now depends on stable
 Compose artifacts (1.6.1). This change was made in
 aosp/2960243

 Upgrade a couple other library dependency versions to
 their latest stable/bugfix release

 Removes dangling camera-viewfinder-compose files that
 weren't deleted in a previous PR

 Change single instance of run {} to runTest {} in
 ImageCaptureDeviceTest


Kimblebee (approved):

temcguir:
Thanks, Kim!

PR #114, opened 2024-02-16
Fix stabilization issue

### LINKED ISSUE #104: Stabilization Mode "ON" not available on device that supports it
Running on a Pixel 8 Pro, which is known to support Preview Stabilization, the `ON` stabilization mode is not showing up as available:

image

This device should have `ON`, `HIGH_QUALITY`, and `OFF` as options.

Kimblebee:
the merged PR  did not include the change where the stabilization check is split between preview and video

resolves #104


temcguir (approved):
Thanks, Kim!

PR #115, opened 2024-02-16
Copyright headers should reflect inception year

temcguir:
- Tell spotless to never update existing year in copyright
- Update all copyright headers to file inception year based on git history


Kimblebee (approved):
thanks for this!

PR #117, opened 2024-02-19
Remove dependency on full Guava library

### LINKED ISSUE #116: App should not include full Guava library
It should use the light-weight futures-ktx library instead. See also the [release notes](https://developer.android.com/jetpack/androidx/releases/concurrent#1.1.0) of this library.

JolandaVerhoef:
Fixes #116.

### DIFF
--- a/feature/quicksettings/build.gradle.kts
+++ b/feature/quicksettings/build.gradle.kts
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2023-2024 The Android Open Source Project
+ * Copyright (C) 2023 The Android Open Source Project
  *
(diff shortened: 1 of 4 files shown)


temcguir (approved):
Thanks!

feature/quicksettings/build.gradle.kts
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2023 The Android Open Source Project
+ * Copyright (C) 2023-2024 The Android Open Source Project
```
temcguir, line 2: Now that #115 is merged, please change this to just `2023`

PR #122, opened 2024-02-23
Single source of truth for camera settings

temcguir:
- Update CameraUseCase to reactive model
- CameraUseCase is only place where default settings will be applied
- PreviewViewModel should not edit CameraAppState
- Setting aspect ratio and flipping camera no longer restarts camera job

### DIFF
--- a/domain/camera/build.gradle.kts
+++ b/domain/camera/build.gradle.kts
@@ -54,10 +54,11 @@ android {
 dependencies {
     // Testing
     testImplementation(libs.junit)
-    androidTestImplementation(libs.androidx.junit)
-    androidTestImplementation(libs.androidx.espresso.core)
+    testImplementation(libs.truth)
     testImplementation("org.jetbrains.kotlinx:kotlinx-coroutines-test:1.7.3")
     testImplementation("org.mockito:mockito-core:5.2.0")
+    androidTestImplementation(libs.androidx.junit)
+    androidTestImplementation(libs.androidx.espresso.core)
 
     // Futures
     implementation(libs.futures.ktx)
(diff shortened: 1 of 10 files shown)


domain/camera/build.gradle.kts
```diff
@@ -54,10 +54,11 @@ android {
 dependencies {
     // Testing
     testImplementation(libs.junit)
-    androidTestImplementation(libs.androidx.junit)
-    androidTestImplementation(libs.androidx.espresso.core)
+    testImplementation(libs.truth)
```
davidjiagoogle, line 57: what's the reason for switching to this?

davidjiagoogle: also goes for the same switch in feature/preview

temcguir: Truth is used for more readable assertions. It uses a fluent API which also gives better error messages. It's especially better than JUnit's `assertTrue`/`assertEquals` for things like collection comparisons or float comparisons. It's used extensively in AOSP and AndroidX and is preferred over JUnit assertions for everything except `assertThrows`, which doesn't have an equivalent in Truth. We should prefer using Truth for assertions going forward.

I mainly added it in this PR because I had to modify a lot of tests and some of them required collection / float comparisons, though it could have been its own PR. If you want me to, I can split it out.

These line of code changes are a little confusing as well. It's not replacing espresso or androidx junit, I was just organizing them a little better so the `testImplementation` dependencies were grouped togeter. I just moved the `androidTestImplementation` dependencies down a few lines.

yasith (approved):

PR #125, opened 2024-02-29
Pop from back stack when Settings back arrow is pressed

temcguir:
When the back arrow is pressed on `SettingsScreen`, `navController.navigate(PREVIEW_ROUTE)` was being called, which added `SettingsScreen` to the back stack. Instead, `navController.popBackStack()` can be used to return to the previous screen, mimicking the behavior of using the device's back button (or back gesture).

### DIFF
--- a/app/src/androidTest/java/com/google/jetpackcamera/ComposeTestRuleExt.kt
+++ b/app/src/androidTest/java/com/google/jetpackcamera/ComposeTestRuleExt.kt
@@ -17,10 +17,13 @@ package com.google.jetpackcamera
 
 import android.content.Context
 import androidx.annotation.StringRes
+import androidx.compose.ui.test.SemanticsMatcher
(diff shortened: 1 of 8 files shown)


Kimblebee:
taking a look at this, I had intentionally changed it from using `navcontroller.popBackStack()` way back in #28 

> "Fix crashing bug that occurs when attempting to navigate back to the preview screen after using a quick setting that rebinds use cases"

temcguir:
> taking a look at this, I had intentionally changed it from using `navcontroller.popBackStack()` way back in #28
> 
> > "Fix crashing bug that occurs when attempting to navigate back to the preview screen after using a quick setting that rebinds use cases"

Thanks for the context. I think the old behavior (popBackStack()) is what we actually want. Do you have repro steps for the old bug? It could be that the cause of the bug was actually fixed since #28.

Kimblebee (approved):

PR #126, opened 2024-03-04
small visual updates

Kimblebee:
This change has two (mostly) cosmetic changes.
1.)  **zoom scale text** is now  centered on the screen instead of being shifted to the left
 --> 

2.) When using`TestableToast`, the display of a system toast is now optional. 
The composable responsible for triggering the Toast display will still be visible to UiAutomator (UI automator only checks for the presence of the composable, not the visibility).

### DIFF
--- a/feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewScreen.kt
+++ b/feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewScreen.kt
@@ -169,7 +169,6 @@ private fun ContentScreen(
         TestableToast(
             modifier = Modifier.testTag(previewUiState.toastMessageToShow.testTag),
             toastMessage = previewUiState.toastMessageToShow,
-            shouldShowToast = false,
             onToastShown = onToastShown
         )
     }
--- a/feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewScreen.kt
+++ b/feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewScreen.kt
@@ -44,7 +44,7 @@ import androidx.lifecycle.compose.LifecycleStartEffect
 import com.google.jetpackcamera.feature.preview.ui.CameraControlsOverlay
 import com.google.jetpackcamera.feature.preview.ui.PreviewDisplay
 import com.google.jetpackcamera.feature.preview.ui.ScreenFlashScreen
-import com.google.jetpackcamera.feature.preview.ui.ShowTestableToast
+import com.google.jetpackcamera.feature.preview.ui.TestableToast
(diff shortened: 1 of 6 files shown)


feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewScreen.kt
```diff
@@ -166,9 +166,10 @@ private fun ContentScreen(
 
     // displays toast when there is a message to show
     if (previewUiState.toastMessageToShow != null) {
-        ShowTestableToast(
+        TestableToast(
             modifier = Modifier.testTag(previewUiState.toastMessageToShow.testTag),
             toastMessage = previewUiState.toastMessageToShow,
+            shouldShowToast = false,
```
temcguir, line 172: Looks like this is hidden by default? Let's unhide for now so we have some visual feedback.

Kimblebee: done 👍🏾

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
```diff
@@ -84,17 +85,17 @@ fun ShowTestableToast(
             .size(20.dp)
             .testTag(toastMessage.testTag)
     ) {
-        // prevents toast from being spammed
-        if (!toastShownStatus.value) {
+        // checking toastShownStatus prevents toast visual from being spammed
```
temcguir, line 88: It seems like we're fighting against compose here a bit. Since `Toast` is not `@Composable`, we probably want to put it into a side-effect. And since side-effects will only run when a key changes, we can use `toastMessage` as a key:
```
@Composable
fun TestableToast(
    toastMessage: ToastMessage,
    onToastShown: () -> Unit
) {
    Box(
        // box seems to need to have some size to be detected by UiAutomator
        modifier = modifier
            .size(20.dp)
            .testTag(toastMessage.testTag)
    )

    LaunchedEffect(toastMessage) {
        if (toastMessage.shouldShowToast) {
            Toast.makeText(
                context,
                context.getText(toastMessage.stringResource),
                toastMessage.toastLength
            ).show()
        }
        onToastShown()
    }
}
```

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewViewModel.kt
```diff
@@ -271,6 +271,7 @@ class PreviewViewModel @Inject constructor(
             _previewUiState.emit(
                 previewUiState.value.copy(
                     quickSettingsIsOpen = isOpen
+
```
JolandaVerhoef, line 274: nit; remove empty line

temcguir (approved):

PR #128, opened 2024-03-04
Make back press exit out of quick settings

davidjiagoogle:
When quick settings screen is open, back press would toggle it to be closed

### DIFF
--- a/feature/quicksettings/src/main/java/com/google/jetpackcamera/feature/quicksettings/QuickSettingsScreen.kt
+++ b/feature/quicksettings/src/main/java/com/google/jetpackcamera/feature/quicksettings/QuickSettingsScreen.kt
@@ -15,6 +15,7 @@
  */
 package com.google.jetpackcamera.feature.quicksettings
 
+import androidx.activity.compose.BackHandler
 import androidx.compose.animation.animateColorAsState
 import androidx.compose.animation.core.animateFloatAsState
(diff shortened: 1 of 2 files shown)


feature/quicksettings/src/main/java/com/google/jetpackcamera/feature/quicksettings/QuickSettingsScreen.kt
```diff
@@ -81,6 +82,9 @@ fun QuickSettingsScreenOverlay(
         )
 
     if (isOpen) {
+        BackHandler {
+            toggleIsOpen()
```
temcguir, line 86: Do we want to add something similar for expanded quick settings? Rather than toggling `isOpen`, should we return to the non-expanded state?

temcguir (commented):
Can you add a quick test to app/src/androidTest/java/com/google/jetpackcamera/NavigationTest.kt now that #125 is merged? You should be able to use the other tests in that file as a template.

feature/quicksettings/src/main/java/com/google/jetpackcamera/feature/quicksettings/QuickSettingsScreen.kt
davidjiagoogle: done

PR #129, opened 2024-03-04
Fix quick settings back press

davidjiagoogle:
Make it so pressing back button in quick settings screen would navigate back instead of closing the app.

Added two more tests in NavigationTest

### DIFF
--- a/feature/quicksettings/src/main/java/com/google/jetpackcamera/feature/quicksettings/QuickSettingsScreen.kt
+++ b/feature/quicksettings/src/main/java/com/google/jetpackcamera/feature/quicksettings/QuickSettingsScreen.kt
@@ -15,6 +15,7 @@
  */
 package com.google.jetpackcamera.feature.quicksettings
 
+import androidx.activity.compose.BackHandler
 import androidx.compose.animation.animateColorAsState
 import androidx.compose.animation.core.animateFloatAsState
 import androidx.compose.animation.core.tween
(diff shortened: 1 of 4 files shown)


feature/quicksettings/src/main/java/com/google/jetpackcamera/feature/quicksettings/QuickSettingsScreen.kt
```diff
@@ -81,6 +82,13 @@ fun QuickSettingsScreenOverlay(
         )
 
     if (isOpen) {
+        val onBack = {
+            when (shouldShowQuickSetting) {
+                IsExpandedQuickSetting.NONE -> toggleIsOpen()
+                else -> shouldShowQuickSetting = IsExpandedQuickSetting.NONE
+            }
```
temcguir, line 89: nit: Can you just put this inline in `BackHandler`?

```
BackHandler {
    when (shouldShowQuickSetting) {
        IsExpandedQuickSetting.NONE -> toggleIsOpen()
        else -> shouldShowQuickSetting = IsExpandedQuickSetting.NONE
    }
}
```

temcguir (approved):

feature/quicksettings/src/main/java/com/google/jetpackcamera/feature/quicksettings/QuickSettingsScreen.kt
davidjiagoogle: onBack is used again below so I made it into a val

temcguir: I only see it being used in the `BackHandler` below. Where else is it being used?

PR #133, opened 2024-03-05
Use LensFacing instead of flipCamera

temcguir:
This PR switches our model to explicitly set the LensFacing direction rather than relying on booleans that specify whether it is the front camera that is showing. This will allow for more flexibility in the future for pulling out camera constraints and for supporting more than 2 lens facing directions.

Changes:
- Add LensFacing to model
- Remove unused CameraUseCase data classes
- Change CameraUseCase.flipCamera() to setLensFacing()
- Use of setLensFacing rather than flipCamera

### DIFF
--- a/data/settings/src/androidTest/java/com/google/jetpackcamera/settings/LocalSettingsRepositoryInstrumentedTest.kt
+++ b/data/settings/src/androidTest/java/com/google/jetpackcamera/settings/LocalSettingsRepositoryInstrumentedTest.kt
@@ -28,6 +28,7 @@ import com.google.jetpackcamera.settings.model.DEFAULT_CAMERA_APP_SETTINGS
 import com.google.jetpackcamera.settings.model.DarkMode
 import com.google.jetpackcamera.settings.model.DynamicRange
 import com.google.jetpackcamera.settings.model.FlashMode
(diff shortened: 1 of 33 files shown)


data/settings/src/androidTest/java/com/google/jetpackcamera/settings/LocalSettingsRepositoryInstrumentedTest.kt
```diff
@@ -108,15 +109,15 @@ class LocalSettingsRepositoryInstrumentedTest {
 
     @Test
     fun can_update_default_to_front_camera() = runTest {
-        // default to front camera starts false
-        val initialFrontCameraDefault = repository.getCameraAppSettings().isFrontCameraFacing
-        repository.updateDefaultToFrontCamera()
+        // default to front camera starts as back
```
yasith, line 112: nit: these 2 comments can probably be updated to say default lens facing starts as back, and default lens facing is now front.

temcguir: Thanks for the catch. I fixed these comments and a couple others in the test method below.

PR #134, opened 2024-03-05
First frame benchmark

Kimblebee:
Benchmark to calculate time to first frame.

**_Whats being measured?_ ❓** 
The benchmark tracks the app startup time and the earliest time that an image success callback fires.
For Cold startups, it tracks the time between `.onCreate` in `MainActivity.kt` and `onCaptureCompleted()` in `CameraXCameraUseCase` AND the time between `PreviewScreen` composing and `onCaptureCompleted()` in `CameraXCameraUseCase`
For Hot startup, it tracks the time between the `PreviewScreen` composing and `onCaptureCompleted()` in `CameraXCameraUseCase`

**_How's it done?_** ❓ 
New class `TraceManager.kt` in the `core:common` module. Location of create trace functions to be used throughout the app.
Use of timestamps to enable preview trace to be performed multiple times in a session. necessary for hot startup testing.

 

**_Anything to note?_** ❓ 
Since the traces tracking startup  and the image success callback cannot be combined into a single trace, taking the sum of the `firstFrameTracePreview` and `Android App Startup` traces could be used to calculate the total time from starting the app to the callback. Could create a script to parse the output json files.
Screenshot 2024-06-27 at 1 43 12 PM
Screenshot 2024-06-27 at 1 31 14 PM


app/src/main/java/com/google/jetpackcamera/MainActivity.kt
```diff
@@ -81,6 +84,11 @@ class MainActivity : ComponentActivity() {
         super.onCreate(savedInstanceState)
         var uiState: MainActivityUiState by mutableStateOf(Loading)
 
+        // start trace between app starting and the earliest possible completed capture
+        // todo: only run traces on a specific build version or flavor(?)
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
```
temcguir, line 89: There shouldn't be a need to limit it. Traces are essentially no-ops when atrace isn't turned on. They're also very lightweight otherwise.

If this is due to API level restrictions of `android.os.Trace`, we should use `androidx.tracing` instead.

app/src/main/java/com/google/jetpackcamera/MainActivity.kt
```diff
@@ -67,6 +68,8 @@ import kotlinx.coroutines.flow.onEach
 import kotlinx.coroutines.launch
 
 private const val TAG = "MainActivity"
+const val FIRST_FRAME_TRACE = "FirstFrameTrace"
```
temcguir, line 71: Can we put this in a `TestTags.kt` file in the `app` module?

app/src/main/java/com/google/jetpackcamera/MainActivity.kt
```diff
@@ -67,6 +68,8 @@ import kotlinx.coroutines.flow.onEach
...
+const val FIRST_FRAME_COOKIE = 12345
```
temcguir, line 72: Assuming we only start 1 first frame trace, we don't really need a cookie. You can just put 0 for the cookie in `begin/endAsyncSection`.

domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
```diff
@@ -123,6 +132,33 @@ constructor(
...
+        val captureCallback = object : CameraCaptureSession.CaptureCallback() {
+            override fun onCaptureCompleted(
+                session: CameraCaptureSession,
+                request: CaptureRequest,
+                result: TotalCaptureResult
+            ) {
+                super.onCaptureCompleted(session, request, result)
+                try {
+                    if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.Q) {
+                        Trace.endAsyncSection(FIRST_FRAME_TRACE, FIRST_FRAME_COOKIE)
```
temcguir, line 149: Having cross-module begin/end could be problematic if the code changes in one spot or the other. I would recommend plumbing up a callback to the spot where the trace is started. So instead of calling `Trace.endAsyncSection` here, you would just invoke the callback. It could be named something like `firstFrameComplete()` and can be passed into the `CameraUseCase.runCamera()` function.

That way you can call `begin/endSection()` at the same spot, and only need to define the trace tag in a single spot.

domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
```diff
@@ -123,6 +132,33 @@ constructor(
...
+                    }
+                } catch (_: Exception) {}
+            }
+        }
+
+        // Create an Extender to attach Camera2 options
+        val previewExtender = Camera2Interop.Extender(imageCaptureBuilder)
```
temcguir, line 156: Should this extender be on `Preview` rather than `ImageCapture`?

domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
```diff
@@ -123,6 +132,33 @@ constructor(
 
     private lateinit var imageCaptureUseCase: ImageCapture
 
+    /**
+     * Applies a CaptureCallback to the provided image capture builder
+     */
+    @OptIn(ExperimentalCamera2Interop::class)
+    private fun onCaptureCompletedCallback(imageCaptureBuilder: ImageCapture.Builder) {
```
temcguir, line 139: nit: call this method `setOnCapturecompletedCallback`.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewScreen.kt
```diff
@@ -75,6 +78,11 @@ fun PreviewScreen(
     viewModel: PreviewViewModel = hiltViewModel<PreviewViewModel, PreviewViewModel.Factory>
         { factory -> factory.create(previewMode) }
 ) {
+    // start trace between preview screen starting and the earliest possible completed capture
+    // todo: only run traces on a specific build version or flavor(?)
+    if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.Q) {
+        Trace.beginAsyncSection(FIRST_FRAME_TRACE, FIRST_FRAME_COOKIE)
```
temcguir, line 84: If you want a separate trace for the PreviewScreen -> First Frame, you'll need to use a separate cookie if you're using the same trace tag.

If you're able to plumb up a callback as I mentioned in another comment, it will make this cookie management easier since you can have 2 separate cookies for `endAsyncSection`.

domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
temcguir, line 149: If this is on Preview, you'll only want to call this once. Since it's in a callback, you could just have a boolean instance variable to check if it's already been called, and set it to `true` once you call it the first time.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/quicksettings/QuickSettingsScreen.kt
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2024 The Android Open Source Project
+ * Copyright (C) 2023 The Android Open Source Project
```
Jaehwa-Noh, line 2:
```suggestion
 * Copyright (C) 2024 The Android Open Source Project
```

Does it need to change years?

domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
```diff
@@ -203,12 +239,14 @@ constructor(
...
 
-        imageCaptureUseCase = ImageCapture.Builder()
+        val imageCaptureBuilder = ImageCapture.Builder()
             .setResolutionSelector(
                 getResolutionSelector(
                     settingsRepository.defaultCameraAppSettings.first().aspectRatio
                 )
-            ).build()
+            )
+        onCaptureCompletedCallback(imageCaptureBuilder)
```
yasith, line 248: We'll probably need to have capture completed callbacks for Preview and ImageCapture Usecases. So maybe it's best to inline the callbacks where we setup the builders.

For this specific benchmark, I think we need to measure the Preview usecase.

domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
Kimblebee: i think this was a naming mistake

domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
Kimblebee: I ended up creating a singleton TraceManager class to cleanly facilitate tracing anywhere in the app. Just inject it where you need it.
I think for future traces, it may become messy to repeatedly plumb up callbacks throughout the app, so I think this may save some future legibility

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewScreen.kt
Kimblebee: So I didn't want to separate the two traces. The issue i kept finding was that the first frame tests using `StartupMode.HOT` weren't being found. The trace start wasnt being called, so I needed to be able to call it in the second best location, which I think is here.

Calling beginAsync on top of each other using the same trace and cookie should be ignored, so there aren't issues with cold startups where it would be "called" twice.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/quicksettings/QuickSettingsScreen.kt
Kimblebee: I don't think so. I think the year was supposed to reflect the date the file was created, not last edited

core/common/src/main/java/com/google/jetpackcamera/core/common/TraceManager.kt
```diff
@@ -0,0 +1,79 @@
...
+@Singleton
+class TraceManager @Inject constructor(@ApplicationContext context: Context) {
+    private var firstFrameTraceStatus = TraceStatus.NOT_STARTED
+
+    // if it is benchmark build, multiple attempts to start traces is allowed in different locations.
+    // especially important for "hot startup" benchmark tests
+    init {
+        if (context.resources.getBoolean(R.bool.is_benchmark_build)) {
+            isBenchmarkBuild = true
+        }
```
Jaehwa-Noh, line 43: I think you can simplify it with remove `if` like directly assigns the value to `isBenchmarkBuild`.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewScreen.kt
```diff
@@ -73,9 +77,14 @@ fun PreviewScreen(
...
     modifier: Modifier = Modifier,
     onRequestWindowColorMode: (Int) -> Unit = {},
+    onFirstFrameCaptureCompleted: () -> Unit = {},
     viewModel: PreviewViewModel = hiltViewModel<PreviewViewModel, PreviewViewModel.Factory>
         { factory -> factory.create(previewMode) }
 ) {
+    // only run this trace in benchmark build. difference from release is by name only
+    // this check allows it to run in the cold run...
+    // if (booleanResource(id = R.bool.is_benchmark_build)) {
+
```
temcguir, line 87: Looks like this can be deleted.

feature/permissions/src/main/java/com/google/jetpackcamera/permissions/PermissionsViewModel.kt
```diff
@@ -25,6 +25,7 @@ import dagger.assisted.Assisted
 import dagger.assisted.AssistedFactory
 import dagger.assisted.AssistedInject
 import dagger.hilt.android.lifecycle.HiltViewModel
+import kotlin.collections.removeFirst as ktRemoveFirst
```
temcguir, line 28: nit: add a comment linking to https://issuetracker.google.com/348683480 to explain why we have to do this.

core/common/src/main/res/values/bools.xml
temcguir, line 1: If `LocalTraceManager` is removed, this can be deleted as well.

core/common/src/benchmark/res/values/bools.xml
temcguir, line 1: If `LocalTraceManager` is removed, this can be removed as well.

app/src/main/java/com/google/jetpackcamera/MainActivity.kt
```diff
@@ -81,6 +84,9 @@ class MainActivity : ComponentActivity() {
         super.onCreate(savedInstanceState)
         var uiState: MainActivityUiState by mutableStateOf(Loading)
 
+        // start trace between app starting and the earliest possible completed capture
+        // todo: only run traces on a specific build version or flavor(?)
```
temcguir, line 88: Can this comment be removed (or moved)?

.idea/androidTestResultsUserPreferences.xml
temcguir, line 1: Please delete.

temcguir (approved):
LGTM. The merge conflict with main needs to be resolved, however.

.idea/androidTestResultsUserPreferences.xml
Jaehwa-Noh: After delete it, what about add it in `.gitignore`?

PR #136, opened 2024-03-12
Save Image on capture

davidjiagoogle:
Use PreviewViewModel::captureImagewithUri instead of captureImage so that the captured image is saved onto disk. Added a device test.


feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/CameraControlsOverlay.kt
```diff
@@ -249,7 +251,18 @@ private fun CaptureButton(
             multipleEventsCutter.processEvent {
                 when (previewMode) {
                     is PreviewMode.StandardMode -> {
-                        onCaptureImage()
+                        val timeStamp = System.currentTimeMillis()
+                        onCaptureImageWithUri(
+                            context.contentResolver,
+                            Uri.fromFile(
+                                File(
+                                    Environment.getExternalStoragePublicDirectory(
```
temcguir, line 259: This method was deprecated in API 29. We should be using something similar to the core test app on API 29+: https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:camera/integration-tests/coretestapp/src/main/java/androidx/camera/integration/core/CameraXActivity.java;l=975;drc=92cfd0386ecb506581413d8ecee586c8f648da7c

Note: you can still do this by generating a URI with ContentResolver.resolve() and use the same `onCaptureImageWithUri()` method. You'll probably want to create a helper function for getting the URI based on API level checks.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/CameraControlsOverlay.kt
```diff
@@ -249,7 +251,18 @@ private fun CaptureButton(
             multipleEventsCutter.processEvent {
                 when (previewMode) {
                     is PreviewMode.StandardMode -> {
-                        onCaptureImage()
+                        val timeStamp = System.currentTimeMillis()
```
temcguir, line 254: Let's use something a little more human-readable. Something like "JCA-YYYY-MM-dd-HH-mm-ss-SSS".

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/CameraControlsOverlay.kt
davidjiagoogle: I changed the case in CameraXCameraUseCase.takePicture() where the provided uri is null to generate the same outputFileOptions as in the linked code. The result is the same: image is saved to the pictures directory.

domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
```diff
@@ -259,9 +261,19 @@ constructor(
...
         val outputFileOptions: OutputFileOptions
         if (imageCaptureUri == null) {
-            val e = RuntimeException("Null Uri is provided.")
-            Log.d(TAG, "takePicture onError: $e")
-            throw e
+            val formatter = SimpleDateFormat(
+                "yyyy-MM-dd-HH-mm-ss-SSS",
+                Locale.US
+            )
+            val fileName = ("CoreTestApp-" + formatter.format(Calendar.getInstance().time))+ ".jpg"
```
temcguir: Replace `CoreTestApp` with `JCA`.

domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
```diff
@@ -259,9 +261,19 @@ constructor(
...
         val outputFileOptions: OutputFileOptions
         if (imageCaptureUri == null) {
-            val e = RuntimeException("Null Uri is provided.")
-            Log.d(TAG, "takePicture onError: $e")
-            throw e
+            val formatter = SimpleDateFormat(
+                "yyyy-MM-dd-HH-mm-ss-SSS",
+                Locale.US
+            )
+            val fileName = (formatter.format(Calendar.getInstance().time)) + ".jpg"
```
temcguir, line 268: Use kotlin string formatting:
val filename = "JCA-${formatter.format(...)}.jpg"

domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
```diff
@@ -259,9 +261,19 @@ constructor(
         val eligibleContentValues = getEligibleContentValues()
         val outputFileOptions: OutputFileOptions
         if (imageCaptureUri == null) {
-            val e = RuntimeException("Null Uri is provided.")
-            Log.d(TAG, "takePicture onError: $e")
-            throw e
```
temcguir, line 264: Since this no longer throws an exception on null, we need to add a null check for the intent flow outside of this method.

davidjiagoogle: Can we just make this the default behavior? If null url is provided, we just save it  to the default location. I added a javaDoc on CameraUseCase.takePicture() for this behavior

temcguir: It's fine for that to be the behavior for CameraUseCase.takePicture(), but we still need to validate the URI isn't null when we are handling `ACTION_IMAGE_CAPTURE`. Technically we should be returning a small Bitmap in that case according to https://developer.android.com/reference/android/provider/MediaStore#ACTION_IMAGE_CAPTURE, but since we don't plan on doing that currently, we should instead show an error. We don't want to have the side-effect of saving a picture that wasn't requested.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/CameraControlsOverlay.kt
```diff
@@ -249,7 +249,7 @@ private fun CaptureButton(
             multipleEventsCutter.processEvent {
                 when (previewMode) {
                     is PreviewMode.StandardMode -> {
-                        onCaptureImage()
+                        onCaptureImageWithUri(context.contentResolver, null) {}
```
temcguir, line 252: In standard mode, when this completes successfully, we should broadcast `ACTION_NEW_PICTURE` with the captured URI. That means you should be returning the Uri from this method as well.

davidjiagoogle: Updated. Instead of having the method returning uri, I made it so that ImageSaved event has the savedUri, which can be picked up by the callback in PreviewMode.StandardMode in the MainActivity for broadcasting

domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraUseCase.kt
```diff
@@ -51,11 +53,15 @@ interface CameraUseCase {
...
+     * Takes a picture with the camera. If ignoreUri is set to true, the picture taken will be saved
+     * at the default directory for pictures on device. Otherwise, it will be saved at the uri
+     * location if the uri is not null. If it is null, an error will be thrown.
      */
-    suspend fun takePicture(contentResolver: ContentResolver, imageCaptureUri: Uri?)
+    suspend fun takePicture(
+        contentResolver: ContentResolver,
+        imageCaptureUri: Uri?,
+        ignoreUri: Boolean = false
+    ): CompletableDeferred<ImageCapture.OutputFileResults>
```
temcguir, line 64: We shouldn't be returning `CompletableDeferred` from a `suspend` function. If anything it should be `ImageCapture.OutputFileResults`.

app/src/main/java/com/google/jetpackcamera/MainActivity.kt
```diff
@@ -126,7 +128,15 @@ class MainActivity : ComponentActivity() {
 
     private fun getPreviewMode(): PreviewMode {
         if (intent == null || MediaStore.ACTION_IMAGE_CAPTURE != intent.action) {
-            return PreviewMode.StandardMode
+            return PreviewMode.StandardMode { event ->
+                if (event is PreviewViewModel.ImageCaptureEvent.ImageSaved) {
+                    val intent = Intent(Camera.ACTION_NEW_PICTURE)
+                    if (event.savedUri != null) {
```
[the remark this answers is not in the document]
davidjiagoogle: OutputFileResults.savedUri is nullable. This makes it still broadcast in that case, just with no data

domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
```diff
@@ -254,11 +256,29 @@ constructor(
...
+            val filename = "JCA-${formatter.format(Calendar.getInstance().time)}.jpg"
+            val contentValues = ContentValues()
+            contentValues.put(MediaStore.MediaColumns.DISPLAY_NAME, filename)
+            contentValues.put(MediaStore.MediaColumns.MIME_TYPE, "image/jpeg")
+            outputFileOptions = OutputFileOptions.Builder(
+                contentResolver,
+                MediaStore.Images.Media.EXTERNAL_CONTENT_URI,
+                contentValues
+            ).build()
+        } else if (imageCaptureUri == null) {
```
temcguir, line 281: Am I missing something here? It looks like this if condition is:
```
if (imageCaptureUri == null) {
} else if (imageCaptureUri == null) {
} else {
}
```
It looks like the first `else` branch is redundant.

app/src/main/java/com/google/jetpackcamera/MainActivity.kt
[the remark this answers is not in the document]
temcguir: According to the documentation for `OutputFileResults.savedUri`:

> Returns null if the [OutputFileOptions](https://developer.android.com/reference/androidx/camera/core/ImageCapture.OutputFileOptions) is constructed with [#Builder(OutputStream)](https://developer.android.com/reference/androidx/camera/core/ImageCapture.OutputFileOptions.Builder).

AFAIK, we only use `Builder(OutputStream)` for the external image capture flow. However, in that case we know that the URI won't be null, since it's passed in. So in both cases, we should never actually have a null URI. We should write the code accordingly to expect that we never have a null URI. This may require moving away from returning a `OutputFileResults` from `takePicture`, and just return the URI directly.

domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
davidjiagoogle: oh good catch. The first if branch is supposed to be if (ignoreUri)

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewViewModel.kt
```diff
@@ -307,7 +310,9 @@ class PreviewViewModel @Inject constructor(
     }
 
     sealed interface ImageCaptureEvent {
-        object ImageSaved : ImageCaptureEvent
+        data class ImageSaved(
+            val savedUri: Uri? = null
```
temcguir, line 314: Can this now be made non-null?

temcguir (approved):

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewViewModel.kt
davidjiagoogle: This is null in the case of external capture right now

PR #138, opened 2024-03-14
Add RefCounted class

temcguir:
This is used for shared state between threads.

 It will allow us to track and release Surface usage
 between a GL thread and the main thread.

Example of how it is being used [here](https://github.com/google/jetpack-camera-app/pull/132/files#diff-2c5c256dae5a6352542cb2bad6539b3bb4a36935bc7d845c014315339c4bd9ab)

### DIFF
--- a/core/common/src/main/java/com/google/jetpackcamera/core/common/RefCounted.kt
+++ b/core/common/src/main/java/com/google/jetpackcamera/core/common/RefCounted.kt
@@ -29,7 +29,7 @@ import kotlinx.atomicfu.loop
  */
 class RefCounted<T : Any>(
     private val debugRefCounts: Boolean = false,
-    private val onRelease: (T) -> Unit = {}
+    private val onRelease: (T) -> Unit
 ) {
     private val refCounted = atomic(uninitialized<T>())
 
@@ -37,6 +37,9 @@ class RefCounted<T : Any>(
      * Initializes the ref-count managed object with the object being managed.
      *
      * This also initializes the implicit ref-count to 1.
+     *
+     * All calls to this function must be paired with [release] to ensure the initial implicit
+     * ref count is decremented and the `onRelease` callback can be called.
      *
      */
     fun initialize(newValue: T) {
@@ -105,7 +108,7 @@ class RefCounted<T : Any>(
      */
     fun release() {
         check(refCounted.value != uninitialized<T>()) {
-            "Ref-count managed object has not yet been initialized. Unable to acquire."
+            "Ref-count managed object has not yet been initialized. Unable to release."
         }
 
         refCounted.loop { old ->
--- a/core/common/src/main/java/com/google/jetpackcamera/core/common/RefCounted.kt
+++ b/core/common/src/main/java/com/google/jetpackcamera/core/common/RefCounted.kt
@@ -1,0 +1,159 @@
+/*
+ * Copyright (C) 2024 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


core/common/src/main/java/com/google/jetpackcamera/core/common/RefCounted.kt
```diff
@@ -0,0 +1,159 @@
...
+class RefCounted<T : Any>(
+    private val debugRefCounts: Boolean = false,
+    private val onRelease: (T) -> Unit = {}
+) {
+    private val refCounted = atomic(uninitialized<T>())
+
+    /**
+     * Initializes the ref-count managed object with the object being managed.
+     *
+     * This also initializes the implicit ref-count to 1.
```
avirakesh, line 39: nit: explicitly mention that `release` must be called after `initialize`

core/common/src/main/java/com/google/jetpackcamera/core/common/RefCounted.kt
```diff
@@ -0,0 +1,159 @@
...
+     * Retrieves the underlying managed object, increasing the ref-count by 1.
+     *
+     * This increases the ref-count if the object has not already been released. If the object has
+     * been released, `null` is returned.
+     *
+     * All calls to this function must be paired with [release], unless `null` is returned.
+     */
+    fun acquire(): T? {
+        check(refCounted.value != uninitialized<T>()) {
+            "Ref-count managed object has not yet been initialized. Unable to acquire."
```
avirakesh, line 68: nit: "Unable to release."

core/common/src/main/java/com/google/jetpackcamera/core/common/RefCounted.kt
```diff
@@ -0,0 +1,159 @@
...
+ * A thread-safe, lock-free class for managing the lifecycle of an object using ref-counting.
+ *
+ * The object that is ref-counted can be initialized late using [initialize].
+ *
+ * @param[debugRefCounts] whether to print debug log statements.
+ * @param[onRelease] a block that will be invoked once when the ref-count reaches 0.
+ */
+class RefCounted<T : Any>(
+    private val debugRefCounts: Boolean = false,
+    private val onRelease: (T) -> Unit = {}
```
avirakesh, line 32: From an API perspective, I would force this to be a required parameter. After all, Kotlin doesn't really need reference counting if onRelease is a no-op. People coming from other languages may try to use RefCouted where `shared_ptr` or `sp` would typically be used (even though it is not needed).

avirakesh (approved):
Generally LGTM! I really wish Kotlin had better resource support handling, but this seems like a decent way to go about it!

core/common/src/main/java/com/google/jetpackcamera/core/common/RefCounted.kt
temcguir: Good catch. Made this change in the `release()` function.

core/common/src/main/java/com/google/jetpackcamera/core/common/RefCounted.kt
temcguir: Done.

core/common/src/main/java/com/google/jetpackcamera/core/common/RefCounted.kt
temcguir: Yeah that's probably better- changed it to be required.

The only case I can think of where you would want a no-op `onRelease` is for flow control. That is, wrap the object and perform tasks only if you can `acquire()`. However, there are probably better ways to do that, and you can always pass in a no-op `onRelease` if that's what you want.

temcguir:
Thanks for the review, Avi!

PR #144, opened 2024-03-15
Show version information in SettingsScreen

temcguir:
Display version information from gradle's generated BuildConfig
 in the app setting screen

 Also adds ability to preview settings screen using Compose previews

image


Kimblebee (approved):
LGTM. One thought I had is if we'd want to include version number in the capture filename/metadata.

temcguir:
Yes that's something we should do. I would prefer to put it in the exif metadata since the filename could be provided by the user when the app is run via external image capture intent.

Issue #119, opened 2024-02-21
Stabilization modes not available on Galaxy S23 Ultra


temcguir:
Video Stabilization modes do not show up as available on the Galaxy S23 Ultra:
![image](https://github.com/google/jetpack-camera-app/assets/1455403/84ea60cd-7941-4056-9ebd-2f288c8e94c9)

This device supports stabilization on the rear-facing camera, but not the front-facing camera.

PR #147, opened 2024-03-19
Supported Stabilization modes reflect all cameras on device

Kimblebee:
closes [#119 Stabilization modes not available on Galaxy S23 Ultra ](#119)

The issue was that when starting JCA, the supported stabilization modes were only reflective of the lens direction when opening the camera.

Now the supported stabilization modes reflect the union of the cameras. The setting currently does not distinguish the supported modes between the rear or front camera.

Tested on s23 with front or back set as the default lens facing

### DIFF
--- a/domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
+++ b/domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
@@ -35,7 +35,6 @@ import androidx.camera.core.Preview
 import androidx.camera.core.SurfaceRequest
 import androidx.camera.core.UseCaseGroup
 import androidx.camera.core.ViewPort
-import androidx.camera.core.impl.stabilization.StabilizationMode
 import androidx.camera.lifecycle.ProcessCameraProvider
 import androidx.camera.video.MediaStoreOutputOptions
 import androidx.camera.video.Recorder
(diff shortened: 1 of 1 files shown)


domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
```diff
@@ -198,19 +218,23 @@ constructor(
...
+                // get device-supported stabilization modes
+                val supportedStabilizationModes = getDeviceSupportedStabilizations()
+
                 settingsRepository.updatePreviewStabilizationSupported(
-                    isPreviewStabilizationSupported(cameraInfo)
+                    supportedStabilizationModes.contains(SupportedStabilizationMode.ON)
                 )
                 settingsRepository.updateVideoStabilizationSupported(
-                    isVideoStabilizationSupported(cameraInfo)
+                    supportedStabilizationModes.contains(SupportedStabilizationMode.ON)
```
temcguir, line 234: This should be:
`supportedStabilizationModes.contains(SupportedStabilizationMode.HIGH_QUALITY)`, correct?

temcguir (approved):
Thanks for adding the popup disclaimer. When we've added more robust constraints, we can make that disclaimer only pop up when it detects that a setting is chosen that won't apply to all camera lenses (using the material3 onError color)

PR #150, opened 2024-03-26
Fix settings screen transition animation

temcguir:
- The settings screen column didn't have a background, so during
   the transition animation, any elements without a background
   use the old background until the transition is complete. This
   adds a background to the settings screen column.

**Before**:
image

**After**:
image


Kimblebee (approved):

temcguir:
Broken test is unrelated. Will be fixed by #154

PR #153, opened 2024-03-27
Show snack bar on video record

davidjiagoogle:
This branch is based on the PR for showing snack bar on image capture: https://github.com/google/jetpack-camera-app/pull/149/


domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
```diff
@@ -385,6 +389,18 @@ constructor(
                 .start(ContextCompat.getMainExecutor(application)) { videoRecordEvent ->
```
temcguir, line 389: nit: the main executor will probably be used anyways, but it's a good idea to use the calling coroutine's dispatcher as an executor here:
```
val callbackExecutor =
        (currentCoroutineContext()[ContinuationInterceptor] as? CoroutineDispatcher)?.asExecutor()
            ?: Runnable::run
```

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewScreen.kt
```diff
@@ -104,11 +111,13 @@ fun PreviewScreen(
...
             onStartVideoRecording = viewModel::startVideoRecording,
             onStopVideoRecording = viewModel::stopVideoRecording,
-            onToastShown = viewModel::onToastShown
+            onToastShown = viewModel::onToastShown,
+            onSnackBarResult = viewModel::onSnackBarResult
         )
     }
 }
 
+@SuppressLint("UnusedMaterial3ScaffoldPaddingParameter")
```
temcguir, line 120: Are we sure we aren't supposed to be using the padding parameter here?

temcguir (approved):
A couple nits but otherwise LGTM.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewScreen.kt
davidjiagoogle: "Scaffold padding values define padding that will be applied along the inside edges of a box.

The value will only be provided if the scaffold view’s bottomBar is set. Otherwise, it will only return 0.0 dp."

Since we don't have a bottomBar here, I think we have no use for the padding value

PR #156, opened 2024-03-29
Update versions and version naming

JolandaVerhoef:


feature/preview/src/test/java/com/google/jetpackcamera/feature/preview/workaround/ComposableCaptureToImage.kt
```diff
@@ -13,8 +13,11 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+@file:SuppressLint("UseSdkSuppress")
```
temcguir, line 16: No action required; this is probably ok since this file is a temporary workaround, but the lint warnings here seem like a bug. The `@RequiresApi` calls are not on tests, they're on utility functions. `@RequiresApi` seems like a more appropriate annotation than `@SdkSuppress` in this case. Should we file an issue?

temcguir (approved):
Thank you, Jolanda!

temcguir (approved):

PR #162, opened 2024-03-29
Smooth rotation on orientation change

JolandaVerhoef:
Allow for a smooth rotation while running the app. Replaced the quick-settings lazy grid with a FlowRow as the grid was depending on screenWidth instead of constraints.

Open issues:
- Surface aspect ratio breaks when in non-initial orientation.
- small flash sometimes when doing orientation change


feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/RotationUtil.kt
```diff
@@ -0,0 +1,118 @@
...
+    return this then Modifier.fillMaxSize().layout { measurable, constraints ->
+        val height = maxOf(constraints.maxWidth, constraints.maxHeight)
+        val width = minOf(constraints.maxWidth, constraints.maxHeight)
+        val placeable = measurable.measure(
+            Constraints.fixed(width, height)
+        )
+        layout(placeable.width, placeable.height) {
+            placeable.placeWithLayer(0, 0) {
+                if (constraints.maxWidth > constraints.maxHeight) {
+                    rotationZ = -currentDegrees
```
alexvanyo, line 107: I think this logic assumes that a `Display.rotation` of `0` implies a portrait orientation. On the Pixel Fold inner screen, `Display.rotation` is `0` when in landscape, so this will layout the content with width <= height, but then not rotate it by any amount.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/RotationUtil.kt
```diff
@@ -0,0 +1,118 @@
...
+ * that width <= height, and rotated afterwards.
+ */
+@Composable
+fun Modifier.rotatedLayout(): Modifier {
+    var currentOrientation by remember { mutableIntStateOf(Surface.ROTATION_0) }
+    val currentDegrees = currentOrientation * 90f
+    val newOrientation = LocalConfiguration.current.orientation
+    val display = LocalView.current.display
+    LaunchedEffect(newOrientation, display) {
+        val newRotation = display.rotation
```
alexvanyo, line 93: The display rotation can change without `LocalConfiguration.current` changing. I think we could register a [display listener](https://developer.android.com/media/camera/camerax/orientation-rotation#displayListener) to track that?

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/RotationUtil.kt
```diff
@@ -0,0 +1,118 @@
...
+            systemBarsBehavior =
+                WindowInsetsControllerCompat.BEHAVIOR_SHOW_TRANSIENT_BARS_BY_SWIPE
+            hide(WindowInsetsCompat.Type.systemBars())
+        }
+    }
+    onDispose {
+        context.getActivity()?.window?.let { window ->
+            if (currentRotationAnimation != null) {
+                window.attributes = window.attributes.apply {
+                    rotationAnimation = WindowManager.LayoutParams.ROTATION_ANIMATION_ROTATE
```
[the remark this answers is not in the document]
temcguir: I've modified this to now restore the rotationAnimation to the old rotation animation before the DisposableEffect was applied. This seems to work pretty well, but I am not quite sure using a DisposableEffect is the right thing to do here. Based on [this documentation](https://developer.android.com/develop/ui/compose/layouts/insets#insets-and-phases) it's possible it may not apply correctly since we are getting the window during composition.

I'll continue to look at this to see if there is a better way using Compose's existing inset modifiers.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
```diff
@@ -145,28 +147,46 @@ fun PreviewDisplay(
                             currentOnFlipCamera()
                         }
                     )
-                },
+                }
+                .layout { measurable, constraints ->
```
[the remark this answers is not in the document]
temcguir: Unfortunately, no. The `aspectRatio` argument is actually an enum that can represent an aspect ratio and its reciprocal, depending on the orientation of the screen. What this layout modifier is doing is calculating the correct aspect ratio that maximizes the screen space used. I'm still not sure if it's the correct thing to do for all screens (specifically foldables), so I'll look into it more, but this does eliminate a recomposition that would be required by changing the `aspectRatio` argument to its reciprocal.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/RotationUtil.kt
temcguir: I was able to key off configuration changes to get the updated display rotation. This eliminated an unnecessary recomposition due to the LaunchedEffect modifying MutableState, so we no longer get a flicker in the UI on rotation changes.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/RotationUtil.kt
temcguir: Good catch. I've changed the code to base the rotation on what the layout should look like in the device natural orientation.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/quicksettings/QuickSettingsScreen.kt
```diff
@@ -109,31 +117,47 @@ fun QuickSettingsScreenOverlay(
...
-            )
+
+        // Only needed to initialize
+        val currentDisplayRotation = LocalView.current.display.rotation
+        var newSurfaceRotation by remember { mutableIntStateOf(currentDisplayRotation) }
+        var prevSurfaceRotation by remember { mutableIntStateOf(currentDisplayRotation) }
+        key(LocalConfiguration.current) {
+            prevSurfaceRotation = newSurfaceRotation
+            newSurfaceRotation = LocalView.current.display.rotation
+        }
```
alexvanyo, line 128: I think this is assuming that the rotation of the display can only change when the `Configuration` also changes, but I don't think that's true in the general case. In free-form windowing, the device (and display) can rotate without the configuration changing.

temcguir: Can you recommend a way to test this? I'll try to address it.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
```diff
@@ -535,7 +558,7 @@ fun ToggleButton(
                 } else {
                     Modifier
                 }
-            ),
+            ).then(modifier),
```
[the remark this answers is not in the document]
temcguir: Good catch. I think this was added accidentally. I'll remove it.

Issue #157, opened 2024-03-29
Improve external storage handling for external intent Image Capture


JolandaVerhoef:
`AndroidManifest.xml` gives a lint warning for this code:

```
    <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE"/>
    <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE" />
```

The warning is:

```
... is deprecated (and is not granted) when targeting Android 13+. If you need to query or interact 
with MediaStore or media files on the shared storage, you should instead use one or more new 
storage permissions: READ_MEDIA_IMAGES, READ_MEDIA_VIDEO or READ_MEDIA_AUDIO. 
```

Anshu-Bijarnia:
Hi, I can try to fix this issue. Can you assign me this issue ?

temcguir:
Oops I just changed this code in #154 , but please take a look to see if there are any improvements that could be made. There were a lot of changes made around this permission and scoped storage between APIs 28-30, so it may be worth testing on emulators running on 27-31 to see if it is working correctly to take pictures / videos.

We can discuss on this issue if you have any questions.

PR #163, opened 2024-03-29
Improved external storage permission handling #157

Anshu-Bijarnia:
This pull request provides a good solution for the external storage handling based on different android versions. Closing -> Improve external storage handling for external intent Image Capture #157

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -1,5 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
-<!--
+<?xml version="1.0" encoding="utf-8"?><!--
   ~ Copyright (C) 2023 The Android Open Source Project
   ~
   ~ Licensed under the Apache License, Version 2.0 (the "License");
@@ -25,10 +24,17 @@
         android:required="false" />
 
     <uses-permission android:name="android.permission.CAMERA" />
-    <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE"
-        android:maxSdkVersion="28"
-        tools:ignore="ScopedStorage" />
 
+    <!-- Devices running Android 12L (API level 32) or lower  -->
+    <uses-permission
+        android:name="android.permission.READ_EXTERNAL_STORAGE"
+        android:maxSdkVersion="32" />
+    <!-- Devices running Android 13 (API level 33) or higher -->
+    <uses-permission android:name="android.permission.READ_MEDIA_IMAGES" />
+    <uses-permission android:name="android.permission.READ_MEDIA_VIDEO" />
+    <!-- To handle the reselection within the app on devices running Android 14
+         or higher if your app targets Android 14 (API level 34) or higher.  -->
+    <uses-permission android:name="android.permission.READ_MEDIA_VISUAL_USER_SELECTED" />
 
     <application
         android:name=".JetpackCameraApplication"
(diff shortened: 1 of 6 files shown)


Anshu-Bijarnia:
@davidjiagoogle can you guide me regarding what i should do to make these tests pass ?

app/src/main/AndroidManifest.xml
```diff
@@ -25,10 +24,17 @@
         android:required="false" />
 
     <uses-permission android:name="android.permission.CAMERA" />
-    <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE"
```
davidjiagoogle, line 28: Hi Anshu, you removed writing permission and replaced it with reading permissions. Won't this cause a problem with image capturing and recording where we save an image or video file onto disk?

davidjiagoogle: More on this: Image capture is tested in ImageCaptureDeviceTest, where the WRITE_EXTERNAL_STORAGE permission is specified again in the test. I believe if you change AppTestUtil.APP_REQUIRED_PERMISSIONS to reflect your changes here, the test will fail, as the app wouldn't be able to save onto storage without WRITE permission

davidjiagoogle:
@Anshu-Bijarnia Have you tried running the tests locally with a connected device or emulator?

Anshu-Bijarnia:
> @Anshu-Bijarnia Have you tried running the tests locally with a connected device or emulator?

I will try fixing the errors and test the application locally.

Anshu-Bijarnia:
> @Anshu-Bijarnia Have you tried running the tests locally with a connected device or emulator?

Can you help me with this, even adding the write permission. The ImageCaptureDeviceTest fails. Write permission is only required till SDK - 28, correct? Higher than that the device already has access to scoped storage by default. Guide me on where i am going wrong.

davidjiagoogle:
@Anshu-Bijarnia  

I can't be sure the exact problem you had after adding these permissions to AppTestUtil.APP_REQUIRED_PERMISSION the PR is not updated with them, but I'm guessing you ran into this error:
"Permission: android.permission.[one of the permissions] cannot be granted!
 junit.framework.AssertionFailedError: Failed to grant permissions"

This could be because you are trying to grant a permission passed a build version on which it can be granted. For example, if you try to grant "READ_EXTERNAL_STORAGE" on 32 and above, this error would show.

Did you version gate the permissions like you did in JcaApp when adding them to AppTestUtil?

Could you update the PR with the latest change and let's see what the errors are?

app/src/main/AndroidManifest.xml
```diff
@@ -26,9 +25,18 @@
...
+    <!-- Devices running Android 12L (API level 32) or lower  -->
+    <uses-permission
+        android:name="android.permission.READ_EXTERNAL_STORAGE"
+        android:maxSdkVersion="32" />
+    <!-- Devices running Android 13 (API level 33) or higher -->
+    <uses-permission android:name="android.permission.READ_MEDIA_IMAGES" />
+    <uses-permission android:name="android.permission.READ_MEDIA_VIDEO" />
+    <!-- To handle the reselection within the app on devices running Android 14
+         or higher if your app targets Android 14 (API level 34) or higher.  -->
+    <uses-permission android:name="android.permission.READ_MEDIA_VISUAL_USER_SELECTED" />
```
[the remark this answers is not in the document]
Anshu-Bijarnia: I will do that and make a PR

PR #165, opened 2024-04-03
Convert UiDevice to ComposeTestRule in ImageCaptureDeviceTest

davidjiagoogle:
I believe the reason why toast.isDisplay fails is that the flow goes too fast and the activity is exited before the composeTestRule can detect the toast being displayed. To fix that, the lines that verify the toast were removed, and only the result is asserted.

I also discovered a bug where in image_capture_external_illegal_uri(), the test is waiting for a RESULT_CANCEL even though the MainActivity never returns the result code in the case of failure. I remember we made it this way intentionally. Fixed this test to verify the failure toast instead of waiting for the ActivityResult

### DIFF
--- a/app/src/androidTest/java/com/google/jetpackcamera/ImageCaptureDeviceTest.kt
+++ b/app/src/androidTest/java/com/google/jetpackcamera/ImageCaptureDeviceTest.kt
@@ -38,12 +38,12 @@ import androidx.test.rule.GrantPermissionRule
 import com.google.jetpackcamera.feature.preview.ui.CAPTURE_BUTTON
 import com.google.jetpackcamera.feature.preview.ui.IMAGE_CAPTURE_FAIL_TOAST
 import com.google.jetpackcamera.feature.preview.ui.IMAGE_CAPTURE_SUCCESS_TOAST
+import java.io.File
+import java.net.URLConnection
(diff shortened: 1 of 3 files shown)


app/src/androidTest/java/com/google/jetpackcamera/ImageCaptureDeviceTest.kt
```diff
@@ -95,25 +97,33 @@ internal class ImageCaptureDeviceTest {
...
-            uiDevice.pressBack()
-            activityScenario!!.result
-        }.register("key_illegal_uri", TEST_CONTRACT) { result ->
-            assert(!result)
-        }.launch(inputUri)
-        deleteFilesInDirAfterTimestamp(timeStamp)
+        )
+
+        intent.putExtra(MediaStore.EXTRA_OUTPUT, uri)
+        intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
```
temcguir, line 125: I don't think we want to use this flag. We should ensure the app is closed from the previous test before running the next test instead. This should be possible by ensuring `ActivityScenario.close()` is called at the end of each test. This can be accomplished using `use`, which is what `runScenarioTest` does in our `UiTestUtil.kt` file.

PR #168, opened 2024-04-08
Add per-lens and system constraints for HDR video

temcguir:
This adds system constraints (including per-lens CameraConstraints) which are initialized once when the CameraUseCase is initialized. This pulls out all of the "supported" fields from `CameraAppSettings`, since those are not actually camera settings, but rather constraints. Constraints are now published separately from their own Repository.

To accomodate this new structure, some changes to the UiState were also made since the constraints did not need to be part of the UiState when the constraints weren't yet ready.

Some renaming occurred in `SettingsRepository` to reinforce that the camera app settings there are actually just default settings, and do not represent the current state of the camera.

- Remove constraints from CameraAppSettings
- Add constraints classes
- Add constraints for dynamic range
- Add ConstraintsRepository to data layer
- Add constraints to SettingsScreen
- Add constraints to preview screen / quick settings
- Add constraints to all tests and use new UiState classes


data/settings/src/main/proto/com/google/jetpackcamera/settings/jca_settings.proto
```diff
@@ -31,19 +31,19 @@ option java_multiple_files = true;
 
 message JcaSettings {
   bool default_front_camera = 2 [deprecated = true];
-  bool front_camera_available = 3;
-  bool back_camera_available = 4;
+  bool front_camera_available = 3 [deprecated = true];
```
yasith, line 34: Do we want to keep these as deprecated? Since we're pre-release, would it be better to get rid of these values?

temcguir: Yeah I'll remove them. 

After release, we will need to keep track and make sure that we don't remove/renumber whenever we update the build code of the app.

yasith (approved):

PR #169, opened 2024-04-10
Bump CameraX version for temporal noise fix

temcguir:
This bumps the snapshot version to the build that includes
 aosp/3029366.


aaa72 (approved):
Thank you!

PR #170, opened 2024-04-10
Ultra HDR image capture in quick settings

heytsai:
This adds Ultra HDR image capture as an option in quicksettings:
  

VideoCapture is currently disabled when Ultra HDR is enabled for following reasons:
1. For single stream mode, CameraEffect does not support HDR related processing.
2. For multiple stream mode, "Preview + Ultra HDR ImageCapture + VideoCapture" will trigger. StreamSharing, which conflict with make it not exactly multiple-stream.
3. Allowing Ultra HDR with HDR video would complicate the app design.

Also for the above reasons, the Ultra HDR, HDR Video and Single Stream Mode features are set to be mutually exclusive. An alter message will be shown when trying to enable two of them:

When Ultra HDR is enabled, the preview will also be set to HDR to match the WYSIWYG concept.


heytsai:
Close this PR, The new approach is in PR#206.

Issue #175, opened 2024-04-12
Image capture blink does not line up with capture started event


temcguir:
The blink on the screen when capturing an image does not line up with the start of the image capture. This is especially aparent when using flash or screen flash: the blink happens well before the flash fires.

We should be using the new `ImageCapture.takePicture` extension function and hook into the `onCaptureStarted` callback to trigger the blink.

See: https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:camera/camera-core/src/main/java/androidx/camera/core/ImageCaptureExt.kt;l=48;drc=a9a0a0a3e75e13ea21708d70e4d5ebfe59025ef9

PR #176, opened 2024-04-15
Use new ImageCapture.takePicture with onCaptureStarted

davidjiagoogle:
Addresses https://github.com/google/jetpack-camera-app/issues/175.

call BlinkState.play() in onCaptureStarted instead of in captureButton pressed


feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewViewModel.kt
```diff
@@ -158,12 +159,12 @@ class PreviewViewModel @Inject constructor(
         }
     }
 
-    fun captureImage() {
+    fun captureImage(blinkState: BlinkState?) {
```
temcguir, line 162: I think this is breaking UDF. The UI should be reacting to state rather than passing it down to the ViewModel.

One way I can think to reverse this is to add a capture counter integer to PreviewUiState that gets incremented every time the onCaptureStarted callback is called. In the UI, we can have a LaunchedEffect keyed on this integer that runs the BlinkState animation.

This will also allow us to push BlinkState into  a `remember` variable inside the `PreviewDisplay` composable and we won't need to pass it around to other components.

We're doing something similar with the zoom UI here: https://github.com/google/jetpack-camera-app/blob/17c50b55f1dcd19bc2752a7d765a815895cc2558/feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/CameraControlsOverlay.kt#L92

temcguir (changes requested):

PR #188, opened 2024-04-18
SnackBar queueing fix

davidjiagoogle:
https://github.com/google/jetpack-camera-app/assets/125502967/10bad22f-f28f-43ad-8904-7d08312b4c28

Everytime a new snackbar shows, the old one is dismissed. Making it always a queue of 1 and reflecting the latest capture


feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
```diff
@@ -127,6 +130,10 @@ fun TestableSnackBar(
     ) {
         val context = LocalContext.current
         scope.launch {
```
temcguir, line 132: We shouldn't be using scope.launch within a `@Composable` (the IDE actually complains about this). It's fine for now, but let's follow up to make it more compose idiomatic.

temcguir (approved):

PR #192, opened 2024-04-23
Use new ImageCapture.takePicture with onCaptureStarted

davidjiagoogle:
Addresses https://github.com/google/jetpack-camera-app/issues/175.

Removed BlinkState and added PreviewUiState.blinkTimeStamp


domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/test/FakeCameraUseCase.kt
```diff
@@ -93,7 +93,7 @@ class FakeCameraUseCase(
             }
     }
 
-    override suspend fun takePicture() {
+    override suspend fun takePicture(onCaptureStarted: (() -> Unit)?) {
```
temcguir, line 96: nit: Why make this nullable and not just `= {}`?

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
```diff
@@ -197,13 +199,31 @@ fun PreviewDisplay(
...
+                    durationMillis = (BLINK_TIME / 2).toInt(),
+                    easing = LinearEasing
+                ),
+                label = "Blink Animation"
+            )
+
+            LaunchedEffect(previewUiState.blinkTimeStamp) {
+                if (previewUiState.blinkTimeStamp != 0L) {
+                    imageVisible = false
+                    delay(BLINK_TIME)
```
temcguir, line 216: Should this also be `BLINK_TIME / 2`?

temcguir (approved):

PR #193, opened 2024-04-24
Updated Permissions Request Flow

Kimblebee:
Cartoon lions asking for permission

⭐ TLDR; Updates the permission screen request flow to be scalable for multiple permission. This PR includes a new optional permission to record audio 🎤 _(PR #194 implements audio in video recording)_
* New permissions can be optional or mandatory (revoked mandatory perms prevent PreviewScreen from loading)
* Granting camera permission,  minimizing the app and setting it to "ask every time" OR revoking the permission will cause it to be prompted when the app is reopened.
* Granting audio/optional permission, minimizing the app, and setting it to "ask every time" will cause it to be prompted when the app is reopened

**The permissions screen will only display if:**
* permissions are set to "ask every time"
* camera permission is not granted

**_Interesting changes:_**
* `permissions` module -- excised all permissions components from `JcaApp` and slapped them into their own ✨ _new_ ✨ module. (I think they like it there 😄 )
* ❗❗ `AppTestUtil.kt` -- added `RECORD_AUDIO` to list of required permissions. **Make sure tests grant or revoke newly added permissions or they will FAIL** (preview screen will be unable to load)
* `JcaApp.kt` --  moved permission screen into the navhost. permissions screen will automatically navigate to preview once complete. Preview will automatically navigate to permissions screen if the permission has been revoked while the app was minimized. pops back stack so you can't navigate back after entering and leaving permissions screen
* `PermissionsEnums.kt` -- closely mirrors `QuickSettingsEnums.kt` in purpose and implementation. It stores UI information relevant for a permission's screen (i.e. Icon, description, accessibility text)
* `PermissionsUi.kt` -- No major visual changes. Refactored to support `PermissionEnum`. Completely removed `CameraPermission` component. `PermissionTemplate` now used for all permissions in `PermissionsScreens.kt`
* `PermissionsViewModel.kt` -- viewmodel for... you guessed it: `PermissionsScreen`. The most important thing it does is hold onto the queue of `PermissionEnums` to be displayed on the permission screen and pops them off as you respond to permission requests.
* `PermissionsScreen.kt` -- Takes in a PermissionsViewModel and displays the first value from the queue. The function to request a required permission changes its behavior if it has been revoked by the user. The user will be taken to the app settings screen to manually grant the permission.  (Since this is rather abrupt, a future PR will prompt a dialog to politely suggest the user to the app settings to grant the permission.)

Demo of the multiple permissions screen

https://github.com/google/jetpack-camera-app/assets/20013168/ba2fbe8c-dcd9-46a3-9496-15139dd9ca35


app/src/main/java/com/google/jetpackcamera/ui/PermissionsEnums.kt
```diff
@@ -0,0 +1,106 @@
...
+import com.google.jetpackcamera.R
+
+const val CAMERA_PERMISSION = "android.permission.CAMERA"
+
+const val AUDIO_RECORD_PERMISSION = "android.permission.RECORD_AUDIO"
+
+/**
+ * Helper class storing a permission's relevant UI information
+ */
+interface PermissionInfoProvider {
```
yasith, line 34: I think a sealed class or an interface is more appropriate here

app/src/main/java/com/google/jetpackcamera/ui/PermissionsUi.kt
yasith, line 1: Probably best to split this into, "PermissionsScreen.kt" and "PermissionsUi.kt", could also be in "ui/permissions"

app/src/main/java/com/google/jetpackcamera/ui/PermissionsUi.kt
```diff
@@ -38,40 +40,112 @@ import androidx.compose.ui.Alignment
...
+ */
 @OptIn(ExperimentalPermissionsApi::class)
 @Composable
-fun CameraPermission(cameraPermissionState: PermissionState, modifier: Modifier = Modifier) {
+fun PermissionsScreen(
+    modifier: Modifier = Modifier,
+    permissionEnums: Set<PermissionEnum>,
+    openAppSettings: () -> Unit
+) {
+    val permissionsViewModel =
```
yasith, line 65: You should't create your ViewModel inside of the PermissionsScreen composable. You can do something similar to what we do in PreviewScreen with the PreviewViewmodel. Or use the `viewModel()` function : https://developer.android.com/reference/kotlin/androidx/lifecycle/viewmodel/compose/package-summary#viewmodel

app/src/main/java/com/google/jetpackcamera/ui/JcaApp.kt
```diff
@@ -38,30 +41,63 @@ import com.google.jetpackcamera.ui.Routes.SETTINGS_ROUTE
...
             modifier = modifier.fillMaxSize(),
-            cameraPermissionState = permissionState
+            permissionEnums = getUnGrantedPermissions(permissionStates),
+            openAppSettings = openAppSettings
         )
     }
 }
 
+@OptIn(ExperimentalPermissionsApi::class)
+private fun getUnGrantedPermissions(
```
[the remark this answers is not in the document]
Kimblebee: yeah that was a placeholder. can't use the term revoked or refused because it actually avoids revoked permissions. 
it only grabs permissions that are required and haven't been granted (camera)
Or Permissions that have been neither denied nor granted (first time the app is opened, or if the permission is set to ask every time)

I may just call it `getRequestablePermissions` and describe the rationale in a kdocs

app/src/main/java/com/google/jetpackcamera/MainActivity.kt
```diff
@@ -200,3 +204,13 @@ private fun Int.toColorModeString(): String {
         else -> "<Unknown>"
     }
 }
+
+/**
+ * Open the app settings when necessary to enable permissions
+ */
+fun Activity.openAppSettings() {
```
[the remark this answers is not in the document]
Kimblebee: yes because it needs to be able to use `startActivity` to open the app settings screen. There may be another way to do this, I just wasn't sure

app/src/main/AndroidManifest.xml
```diff
@@ -25,6 +25,8 @@
         android:required="false" />
 
     <uses-permission android:name="android.permission.CAMERA" />
+    <uses-permission android:name="android.permission.RECORD_AUDIO" />
+
```
davidjiagoogle, line 29: nit: remove extra line?

app/src/main/java/com/google/jetpackcamera/ui/permissions/PermissionsEnums.kt
```diff
@@ -0,0 +1,112 @@
...
+    },
+
+    RECORD_AUDIO {
+        override fun getPermission(): String = AUDIO_RECORD_PERMISSION
+
+        override fun isOptional(): Boolean = false
+
+        override fun getDrawableResId(): Int = R.drawable.audio_microphone
+
+        override fun getImageVector(): ImageVector? = null
```
temcguir, line 102: Instead of using the drawable, you should use material3-icon's `Icons.Outlined.Mic` here.

app/src/main/java/com/google/jetpackcamera/ui/permissions/PermissionsEnums.kt
```diff
@@ -0,0 +1,112 @@
...
+enum class PermissionEnum : PermissionInfoProvider {
+
+    CAMERA {
+        override fun getPermission(): String = CAMERA_PERMISSION
+
+        override fun isOptional(): Boolean = false
+
+        override fun getDrawableResId(): Int = R.drawable.photo_camera
+
+        override fun getImageVector(): ImageVector? = null
```
temcguir, line 84: You can replace the drawable here with `Icons.Outline.PhotoCamera`

app/src/main/res/drawable/audio_microphone.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8"?>
```
temcguir, line 1: As mentioned in the Enums file, you shouldn't need to add this if you use the material3 icons library.

app/src/main/res/values/strings.xml
```diff
@@ -16,11 +16,16 @@
...
     <string name="camera_not_available">Camera not available</string>
     <string name="external_capture_uri_not_supplied">No parcelable URI is supplied for external capture.</string>
+
+    <string name="camera_permission_screen_title">Enable Camera</string>
+    <string name="camera_permission_required_rationale">Please provide permission to access to the camera. It is necessary for this app to function.</string>
+    <string name="camera_permission_accessibility_text">An icon representing a camera</string>
+    <string name="request_permission">Allow Access</string>
+
+    <string name="microphone_permission_screen_title">Enable Audio Recording</string>
+    <string name="microphone_permission_required_rationale">Enabling this is necessary for audio in video recordings.</string>
```
temcguir, line 29: Should we also mention explicitly that it's possible to record without audio?

app/src/main/AndroidManifest.xml
```diff
@@ -25,6 +25,8 @@
         android:required="false" />
```
temcguir, line 26: From [this list](https://developer.android.com/guide/topics/manifest/uses-feature-element#permissions), we should also add:
```
<uses-feature
    android:name="android.hardware.microphone"
    android:required="false" />
```

app/src/main/java/com/google/jetpackcamera/ui/permissions/PermissionsViewModel.kt
```diff
@@ -0,0 +1,37 @@
...
+import androidx.lifecycle.ViewModel
+
+/**
+ * A [ViewModel] for [PermissionsScreen]]
+ */
+class PermissionsViewModel(
+    val permissionEnums: Set<PermissionEnum>
+) : ViewModel() {
+
+    val visiblePermissionDialogQueue = mutableStateListOf<PermissionEnum>()
```
temcguir, line 28: For consistency, let's keep `compose` classes out of the `ViewModel`. Instead, you can replace this with a `MutableStateFlow<PermissionsUiState>` where `PermissionsUiState` looks something like:
```
sealed interface PermissionsUiState {
  data class PermissionsNeeded(val currentPermission: PermissionEnum) : PermissionsUiState 
  object AllPermissionsGranted : PermissionsUiState
}
```

You will need to move some of the bookkeeping from JcaApp to PermissionsViewModel, but that's probably the right thing to do anyways.

app/src/main/java/com/google/jetpackcamera/ui/JcaApp.kt
```diff
@@ -34,34 +38,79 @@ import com.google.jetpackcamera.settings.SettingsScreen
...
     } else {
-        CameraPermission(
+        val permissionsViewModel =
+            viewModel { PermissionsViewModel(getRequestablePermissions(permissionStates)) }
+        // you'll have the option to go through camera and all other optional permissions
+        PermissionsScreen(
             modifier = modifier.fillMaxSize(),
-            cameraPermissionState = permissionState
+            viewModel = permissionsViewModel,
+            openAppSettings = openAppSettings
```
temcguir, line 82: It would be preferable to get this into JetpackCameraNavHost so we're more consistent with how we handle our screens.

feature/permissions/src/main/java/com/google/jetpackcamera/permissions/PermissionsViewModel.kt
```diff
@@ -0,0 +1,102 @@
...
+        permissionQueue.addAll(getRequestablePermissions(permissionStates))
+    }
+
+    private val _permissionsUiState: MutableStateFlow<PermissionsUiState> =
+        MutableStateFlow(getCurrentPermission())
+    val permissionsUiState: StateFlow<PermissionsUiState> = _permissionsUiState.asStateFlow()
+
+    private fun getCurrentPermission(): PermissionsUiState {
+        if (permissionQueue.isEmpty()) {
+            return PermissionsUiState.AllPermissionsGranted
```
temcguir, line 59: Since both these branches return, you can just say:
```
return if (...) ... else ...
```

feature/permissions/src/main/java/com/google/jetpackcamera/permissions/PermissionsViewModel.kt
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    fun dismissPermission() {
+        if (!permissionQueue.isEmpty()) {
+            permissionQueue.removeFirst()
+        }
+
+        viewModelScope.launch {
+            _permissionsUiState.emit(getCurrentPermission())
+        }
```
temcguir, line 72: Since `_permissionsUiState` is a `StateFlow`, you can skip the `viewModelScope.launch` and just do `_permissionsUiState.update { }`.

feature/permissions/src/main/java/com/google/jetpackcamera/permissions/PermissionsViewModel.kt
```diff
@@ -0,0 +1,102 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.jetpackcamera.permissions
+
+import android.Manifest
+import androidx.compose.runtime.mutableStateListOf
```
temcguir, line 19: Let's try to avoid using compose classes in ViewModels. It kind of locks us in to using compose if we want to write unit tests against the ViewModels themselves, which shouldn't strictly be necessary.

feature/permissions/src/main/java/com/google/jetpackcamera/permissions/ui/PermissionsScreenComponents.kt
```diff
@@ -37,41 +37,67 @@ import androidx.compose.runtime.Composable
...
         modifier = modifier,
-        permissionState = cameraPermissionState,
-        painter = painterResource(id = R.drawable.photo_camera),
-        iconAccessibilityText = stringResource(id = R.string.camera_permission_accessibility_text),
-        title = stringResource(id = R.string.camera_permission_screen_title),
-        bodyText = stringResource(id = R.string.camera_permission_required_rationale),
-        requestButtonText = stringResource(R.string.request_permission)
+        onRequestPermission = {
+            // if declined by user, must navigate to system app settings to enable permission
+            // todo(b/337338326): open a dialog that tells user they must go to device's app settings to enable permissions
```
temcguir, line 72: Can we just change the message in the `PermissionTemplate`'s rationale text to re-iterate that the camera permission is required and change the button text to say something like "Go to app settings"?

feature/permissions/src/main/java/com/google/jetpackcamera/permissions/ui/PermissionsScreenComponents.kt
```diff
@@ -37,41 +37,67 @@ import androidx.compose.runtime.Composable
...
+            if (permissionState.status.shouldShowRationale) {
```
temcguir, line 73: Does this take into account that the developer might have purposefully denied a permission? When I deny the audio permission, I'm still being asked to go to the app settings.

Kimblebee: this is only used when a permission is flagged as required (which is just camera permission). If a user declines a permission flagged as optional (microphone), the screen closes after declining and it will not be prompted again when the app is reopened.
> When I deny the audio permission, I'm still being asked to go to the app settings.

I didn't have this issue on my devices, I'll check again on some other devices and to make sure the logic is right.

temcguir: It's still happening on my S23 Ultra. Except now it crashes on line 87 when I deny the permission with a `NullPointerException` because the audio permission doesn't have a rationale body text.

temcguir: After checking out the latest code, it no longer crashes. I was able to get into a state where the spinner was going forever saying the camera was loading, but that may be unrelated to this change. We can tackle that as a separate bug.

feature/permissions/src/test/java/com/google/jetpackcamera/permissions/ExampleUnitTest.kt
temcguir, line 1: Please remove.

temcguir (approved):

PR #194, opened 2024-04-24
enable audio in video recording

Kimblebee:
Adds audio to video if the permission has been granted.

**Currently, the video will always have audio _if_ the permission is granted.** This can be merged into main on its own, but it needs  #193 to be functional

Demo Recording (rebased on top of #193):

https://github.com/google/jetpack-camera-app/assets/20013168/b1c5c1b1-e04f-4771-8fb7-3838ba11e46d


domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
```diff
@@ -415,10 +419,14 @@ constructor(
         return eligibleContentValues
     }
 
+    @SuppressLint("MissingPermission")
```
yasith, line 422: Why do we need the MissingPermission suppress lint here? Is it because the other PR isn't merged yet?

If so, you can rebase this PR on top of the other one.

domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
```diff
@@ -415,10 +419,14 @@ constructor(
...
     override suspend fun startVideoRecording(
         onVideoRecord: (CameraUseCase.OnVideoRecordEvent) -> Unit
     ) {
         Log.d(TAG, "recordVideo")
+        //todo: add setting to enable or disable audio when permission is granted
```
yasith, line 427: nit: create an issue for the TODO, and add the issue id here.

yasith (approved):

PR #197, opened 2024-04-29
David/camera provider await instance

davidjiagoogle:
Use the new API in ProcessCameraProviderExt


domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
```diff
@@ -124,7 +124,7 @@ constructor(
     private val currentSettings = MutableStateFlow<CameraAppSettings?>(null)
 
     override suspend fun initialize() {
-        cameraProvider = ProcessCameraProvider.getInstance(application).await()
+        cameraProvider = ProcessCameraProvider.awaitInstance(application.applicationContext)
```
temcguir, line 127: Since `Application` is a `Context`, you should be able to get rid of the extra `.applicationContext`.

temcguir (approved):

PR #199, opened 2024-05-01
Disable external video capture

davidjiagoogle:
Move PreviewMode into PreviewUiState via assisted injection. I
n PreviewVideModel::startVideoRecording(), check previewMode. 
If previewMode is ExternalImageCaptureMode, no-op and display the appropriate snackbar.
Device test added.


domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
```diff
@@ -400,6 +403,10 @@ constructor(
     override suspend fun startVideoRecording(
         onVideoRecord: (CameraUseCase.OnVideoRecordEvent) -> Unit
     ) {
+        if (videoCaptureUseCase == null) {
+            Log.e(TAG, "attempted video recording with null videoCapture use case")
+            return
```
temcguir, line 408: This should throw an exception or return an error in the `onVideoRecord` callback. Failing silently could cause issues.

temcguir (approved):
Overall this looks good. We can expand on it later when we need to do things like disable image capture when HDR video is enabled.

It would also be good to get a test for this in `ImageCaptureDeviceTest`.

PR #200, opened 2024-05-02
Update to CameraX viewfinder-compose snapshot with API changes

temcguir:
- Now uses EMBEDDED and EXTERNAL instead of PERFORMANCE and COMPATIBLE

 - Also includes fixes to display EMBEDDED correctly & proper surface
   release logic.


JolandaVerhoef (approved):
LGTM!

PR #201, opened 2024-05-02
audio visualizer

Kimblebee:
Added a visualizer for audio input while recording video.
It only displays while video is recording
Displays muted icon when there is ZERO audio input
Screenshot 2024-05-01 at 5 18 26 PM

here is a quick demo🐻

https://github.com/google/jetpack-camera-app/assets/20013168/4763a6a7-5845-4289-b198-01bd6c3d1659


feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
```diff
@@ -74,6 +78,59 @@ import kotlinx.coroutines.delay
...
+                                .toDouble()
+                            )
+                )
+            ).dp,
+        label = "AudioAnimation"
+    )
+    Box(modifier = modifier) {
+        Canvas(
+            modifier = Modifier
+                .size(animatedSize)
```
JolandaVerhoef, line 103: Instead of updating the size here, could you use a fixed max size and draw a circle of the needed size inside the Canvas onDraw? That way you won't need a new layout pass every time the value change (which will be every frame I guess), Compose will be smart enough to only re-draw.

Kimblebee: Good idea! it hadn't crossed my mind

temcguir (approved):

PR #202, opened 2024-05-08
Add tap to focus feature using the coordinate transformer in viewfinder

yasith:
It looks to me as if there's some jank when starting focus and metering. Let me know what you think

https://github.com/google/jetpack-camera-app/assets/417355/bbc0817e-741e-4859-a1c7-706d5147c0d5

### DIFF
--- a/domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraEvent.kt
+++ b/domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraEvent.kt
@@ -1,0 +1,29 @@
+/*
+ * Copyright (C) 2024 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.jetpackcamera.domain.camera
+
+import androidx.camera.core.MeteringPoint
+
+/**
+ * An event that can be sent to the camera coroutine.
+ */
+sealed interface CameraEvent {
+
+    /**
+     * Represents a focus metering event, that the camera can act on.
+     */
+    class FocusMeteringEvent(val meteringPoint: MeteringPoint) : CameraEvent
+}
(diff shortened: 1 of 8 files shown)


domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraUseCase.kt
```diff
@@ -86,7 +87,7 @@ interface CameraUseCase {
 
     suspend fun setLensFacing(lensFacing: LensFacing)
 
-    fun tapToFocus(display: Display, surfaceWidth: Int, surfaceHeight: Int, x: Float, y: Float)
+    suspend fun tapToFocus(meteringPoint: MeteringPoint)
```
temcguir, line 90: Let's try to avoid adding another CameraX class to the `CameraUseCase` interface. Make this a point (x and y) and we can keep the `MeteringPointFactory` inside of `CameraXCameraUseCase`. We have access to the `SurfaceRequest` there before we send it, so we can get the buffer dimensions there.

domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraEvent.kt
```diff
@@ -0,0 +1,23 @@
...
+ * limitations under the License.
+ */
+
+package com.google.jetpackcamera.domain.camera
+
+import androidx.camera.core.MeteringPoint
+
+sealed interface CameraEvent
+
+class FocusMeteringEvent(val meteringPoint: MeteringPoint) : CameraEvent
```
temcguir, line 23: nit: We've been putting our sealed interface implementations nested in the sealed interface itself. For consistency, can we change this to:
```
sealed interface CameraEvent {
  class FocusMetering(...) : CameraEvent
}
```

temcguir (changes requested):

domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
```diff
@@ -287,6 +290,20 @@ constructor(
                 var prevTransientSettings = initialTransientSettings
                 cameraProvider.runWith(sessionSettings.cameraSelector, useCaseGroup) { camera ->
                     Log.d(TAG, "Camera session started")
+
+                    launch {
+                        cameraEvents.consumeAsFlow().collect {
```
temcguir, line 295: This is probably OK as is for now, but I would recommend wrapping this in a `getSurfaceRequests().filterNotNull().map<MeteringPointFactory> { ... }.collectLatest { }` rather than creating a new `MeteringPointFactory` for every event.  This allows you to make `tapToFocus` no longer need to be a `suspend fun` (though we may need to do that later anyways for a response).

You would also want to break out the `FocusMeteringEvent` into its own channel so you can have a conflated channel that only delivers the latest event. 

I'm fine with whichever way you want to go.

temcguir (approved):

PR #203, opened 2024-05-09
Add ToggleButton ui component

heytsai:
This adds a ui component that can be used to choose between image capture and video capture when HDR is enabled:

Screenshot from 2024-05-09 20-30-50

The button can be disabled for the case that only HDR video is supported

The animation for transition between different states is also implemented for completeness.

### DIFF
--- a/feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
+++ b/feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
@@ -44,7 +44,6 @@ import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.material.icons.Icons
 import androidx.compose.material.icons.filled.CameraAlt
 import androidx.compose.material.icons.filled.FlipCameraAndroid
-import androidx.compose.material.icons.filled.Refresh
 import androidx.compose.material.icons.filled.Settings
 import androidx.compose.material.icons.filled.VideoStable
 import androidx.compose.material.icons.filled.Videocam
--- a/feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
+++ b/feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
@@ -15,6 +15,7 @@
  */
 package com.google.jetpackcamera.feature.preview.ui
 
+import android.content.res.Configuration
 import android.util.Log
 import android.view.Display
 import android.widget.Toast
@@ -33,9 +34,10 @@ import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.BoxWithConstraints
 import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.aspectRatio
+import androidx.compose.foundation.layout.fillMaxHeight
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.height
-import androidx.compose.foundation.layout.offset
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.size
 import androidx.compose.foundation.layout.width
@@ -59,7 +61,6 @@ import androidx.compose.material3.SuggestionChip
 import androidx.compose.material3.Surface
 import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
-import androidx.compose.runtime.CompositionLocalProvider
 import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
@@ -75,22 +76,20 @@ import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.graphics.painter.Painter
 import androidx.compose.ui.graphics.vector.rememberVectorPainter
 import androidx.compose.ui.input.pointer.pointerInput
-import androidx.compose.ui.layout.onSizeChanged
+import androidx.compose.ui.layout.layout
 import androidx.compose.ui.platform.LocalContext
-import androidx.compose.ui.platform.LocalDensity
 import androidx.compose.ui.platform.testTag
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.Dp
-import androidx.compose.ui.unit.IntOffset
 import androidx.compose.ui.unit.dp
 import androidx.compose.ui.unit.sp
 import com.google.jetpackcamera.feature.preview.PreviewUiState
 import com.google.jetpackcamera.feature.preview.R
 import com.google.jetpackcamera.feature.preview.VideoRecordingState
+import com.google.jetpackcamera.feature.preview.ui.theme.PreviewPreviewTheme
 import com.google.jetpackcamera.settings.model.AspectRatio
 import com.google.jetpackcamera.settings.model.Stabilization
-import kotlin.math.roundToInt
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.launch
 
@@ -401,40 +400,30 @@ fun ToggleButton(
     modifier: Modifier = Modifier.width(64.dp).height(32.dp),
     initialState: ToggleState = ToggleState.Left,
     onToggleStateChanged: (newState: ToggleState) -> Unit = {},
-    isDisabled: Boolean = false,
+    enabled: Boolean = true,
     iconPadding: Dp = 8.dp
 ) {
-    val backgroundColor = MaterialTheme.colorScheme.surface
+    val backgroundColor = MaterialTheme.colorScheme.surfaceContainerHighest
     val disableColor = MaterialTheme.colorScheme.onSurface
     val iconSelectionColor = MaterialTheme.colorScheme.onPrimary
     val iconUnSelectionColor = MaterialTheme.colorScheme.primary
     val circleSelectionColor = MaterialTheme.colorScheme.primary
-    val circleColor = if (isDisabled) disableColor.copy(alpha = 0.12f) else circleSelectionColor
-    val density = LocalDensity.current
-    var width by remember { mutableStateOf(0.dp) }
-    var height by remember { mutableStateOf(0.dp) }
+    val circleColor = if (enabled) circleSelectionColor else disableColor.copy(alpha = 0.12f)
     var toggleState by remember { mutableStateOf(initialState) }
     val animatedTogglePosition by animateFloatAsState(
         when (toggleState) {
             ToggleState.Left -> 0f
             ToggleState.Right -> 1f
         },
-        label = "anchor"
+        label = "togglePosition"
     )
     val scope = rememberCoroutineScope()
 
     Surface(
         modifier = modifier
-            .onSizeChanged {
-                width = with(density) { it.width.toDp() }
-                height = with(density) { it.height.toDp() }
-            }
             .clip(shape = RoundedCornerShape(50))
-            .background(backgroundColor)
             .then(
-                if (isDisabled) {
-                    Modifier
-                } else {
+                if (enabled) {
                     Modifier.clickable {
                         scope.launch {
                             toggleState = when (toggleState) {
@@ -444,8 +433,11 @@ fun ToggleButton(
                             onToggleStateChanged(toggleState)
                         }
                     }
+                } else {
+                    Modifier
                 }
-            )
+            ),
+        color = backgroundColor
     ) {
         Box {
             Row(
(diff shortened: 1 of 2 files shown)


temcguir (commented):
This is looking good. We just need to add some polish.

Generally, we want to use the material design guidelines to design our components. This toggle button seems to be a combination of [IconButton](https://m3.material.io/components/icon-buttons/overview)s on a track.

See [this spec](https://m3.material.io/components/icon-buttons/specs#dbe9395c-f772-408e-8dd3-b8cf34be0656) for how we should treat the different states of the icons. The selected icon should use a `Filled` variants whereas the unselected icon should use an `Outlined` variant. 

Note that [disabled states have a transparency applied](https://m3.material.io/foundations/interaction/states/applying-states#3c3032e8-b07a-42ac-a508-a32f573cc7e1). For those states, the color can be modified with:
```
color.copy(alpha = 0.38f)
```

With these changes, I would expect the toggle button to look something like this:
Screenshot 2024-05-10 at 4 14 40 PM
(Note the colors here all come from `MaterialTheme.colorScheme`.)

JolandaVerhoef:
UX question - is there a reason you're showing the video on the left and still image on the right?  The Pixel system camera has them the other way around (Camera icon on the left, video on the right)

feature/preview/src/main/res/drawable/hdr_image.xml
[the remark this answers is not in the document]
heytsai: Thank you, updated.

feature/preview/src/main/res/drawable/hdr_video.xml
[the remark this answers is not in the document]
heytsai: Thank you, updated.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
```diff
@@ -365,3 +387,150 @@ fun CaptureButton(
...
+    }
+}
+
+@Preview(backgroundColor = 0xFF000000, showBackground = true)
+@Composable
+private fun Preview_ToggleButton_Selecting_Left() {
+    CompositionLocalProvider(LocalContentColor provides Color.White) {
+        ToggleButton(
+            leftIcon = painterResource(id = R.drawable.hdr_video),
+            rightIcon = painterResource(id = R.drawable.hdr_image),
```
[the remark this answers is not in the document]
heytsai: Thank you, done. Also, add some codes in the preview that switches icons between filled and outlined based on selection state.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
```diff
@@ -365,3 +387,150 @@ fun CaptureButton(
...
+@Composable
+fun ToggleButton(
+    leftIcon: Painter,
+    rightIcon: Painter,
+    modifier: Modifier = Modifier,
+    initialState: ToggleState = ToggleState.Left,
+    onToggleStateChanged: (newState: ToggleState) -> Unit = {},
+    isDisabled: Boolean = false,
+    width: Dp = 64.dp,
+    height: Dp = 32.dp,
```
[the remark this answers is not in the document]
heytsai: Thank you, done.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
```diff
@@ -365,3 +387,150 @@ fun CaptureButton(
...
+    iconPadding: Dp = 8.dp,
+    selectionColor: Color = Color(0xFF1ABC9C),
+    backgroundColor: Color = Color.LightGray
```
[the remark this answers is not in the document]
heytsai: Thank you. Since the color definition becomes a little complicated, I moved them inside the method. PTAL.

heytsai:
> UX question - is there a reason you're showing the video on the left and still image on the right? The Pixel system camera has them the other way around (Camera icon on the left, video on the right)

Thank you, updated it to put the camera icon on the left.

There is no strong reason, just because the checked item will always be video when the toggle is disabled (if a device supports Ultra HDR, it will also support HDR video), and I feel more natural to put it on the left side.

heytsai:
The ui component after fixes would be like follows:

Screenshot from 2024-05-13 15-39-49

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
```diff
@@ -365,3 +388,169 @@ fun CaptureButton(
...
+        label = "anchor"
+    )
+    val scope = rememberCoroutineScope()
+
+    Surface(
+        modifier = modifier
+            .onSizeChanged {
+                width = with(density) { it.width.toDp() }
+                height = with(density) { it.height.toDp() }
+            }
```
temcguir, line 431: This is going to cause an extra unnecessary recomposition. We should be able to remove the rememberd width/height variables if we switch to applying the position during the layout phase (see comment below).

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
```diff
@@ -365,3 +388,169 @@ fun CaptureButton(
...
+                    Modifier
+                        .offset {
+                            IntOffset(
+                                x = with(density) {
+                                    animatedTogglePosition * (width - height).toPx()
+                                }.roundToInt(),
+                                y = 0
+                            )
+                        }
+                        .size(height)
```
temcguir, line 465: This can be replaced with the following code to eliminate the need for width/height variables:
```
                        .layout { measurable, constraints ->
                            val placeable = measurable.measure(constraints)
                            layout(placeable.width, placeable.height) {
                                val xPos = animatedTogglePosition *
                                        (constraints.maxWidth - placeable.width)
                                placeable.placeRelative(xPos.toInt(), 0)
                            }
                        }
                        .fillMaxHeight()
                        .aspectRatio(1f)
```

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
```diff
@@ -365,3 +388,169 @@ fun CaptureButton(
...
+fun ToggleButton(
+    leftIcon: Painter,
+    rightIcon: Painter,
+    modifier: Modifier = Modifier.width(64.dp).height(32.dp),
+    initialState: ToggleState = ToggleState.Left,
+    onToggleStateChanged: (newState: ToggleState) -> Unit = {},
+    isDisabled: Boolean = false,
+    iconPadding: Dp = 8.dp
+) {
+    val backgroundColor = MaterialTheme.colorScheme.surface
```
temcguir, line 407: This should be `MaterialTheme.colorScheme.surfaceContainerHighest` according to the [spec](https://m3.material.io/components/icon-buttons/specs#5309610a-4515-44f9-830d-880e2a2240a2)
image

When it's disabled, it should also be `disableColor` with an alpha of `0.12`:
image

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
```diff
@@ -365,3 +388,169 @@ fun CaptureButton(
...
+}
+
+@Preview(backgroundColor = 0xFF000000, showBackground = true)
+@Composable
+private fun Preview_ToggleButton_Selecting_Left() {
+    val initialState = ToggleState.Left
+    var toggleState by remember {
+        mutableStateOf(initialState)
+    }
+    CompositionLocalProvider(LocalContentColor provides Color.White) {
```
temcguir, line 513: Rather than using this `CompositionLocalProvider`, you can instead wrap this with:
```
    MaterialTheme(colorScheme = darkColorScheme()) { }
```

Since we're likely to only want to use the dark color scheme for now. Alternatively, if you want to show both dark and light color schemes, you can replace the `colorScheme` argument with:
```
if (isSystemInDarkTheme()) darkColorScheme() else lightColorScheme()
```
and use multiple `@Preview` annotations on top of each of these functions:
```
@Preview(name = "Light Mode", uiMode = Configuration.UI_MODE_NIGHT_NO)
@Preview(name = "Dark Mode", uiMode = Configuration.UI_MODE_NIGHT_YES)
```
(it also helps to wrap the `ToggleButton` in a `Surface {}` block so the background adapts to the color scheme as well.)

temcguir (changes requested):

heytsai:
Thank you Trevor. The preview after the fixed is as follows (the first one is Light Mode):

Screenshot from 2024-05-15 11-00-12

Also, I changed the parameter "isDisabled" to "enabled" to align with other components (e.g., Button).

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
```diff
@@ -365,3 +388,169 @@ fun CaptureButton(
...
+    val scope = rememberCoroutineScope()
+
+    Surface(
+        modifier = modifier
+            .onSizeChanged {
+                width = with(density) { it.width.toDp() }
+                height = with(density) { it.height.toDp() }
+            }
+            .clip(shape = RoundedCornerShape(50))
+            .background(backgroundColor)
```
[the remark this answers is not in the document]
heytsai: Thank you, updated.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
heytsai: Thank you, updated.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
heytsai: Thank you, updated! It looks more straight forward now.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
heytsai: Thank you, updated the background color to "surfaceContainerHighest". Since the disable color is already "onSurface", using 0.12 alpha for container and 0.38 for the icon, I left them there.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
heytsai: Thank you. Since I saw there is a method called "SettingsPreviewTheme" in the settings artifact, which has complete considerations for preview theme, I copied it and renamed it as "PreviewPreviewTheme" for reusing through component preview.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
```diff
@@ -365,3 +387,164 @@ fun CaptureButton(
...
+            Row(
+                modifier = Modifier.matchParentSize().then(
+                    if (enabled) Modifier else Modifier.alpha(0.38f)
+                ),
+                verticalAlignment = Alignment.CenterVertically,
+                horizontalArrangement = Arrangement.SpaceBetween
+            ) {
+                Icon(
+                    painter = leftIcon,
+                    contentDescription = "leftIcon",
```
temcguir, line 472: Add this content discription as an argument to the composable with the default of `"leftIcon"`. We should be able to change it to be an accurate description of the icon.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
```diff
@@ -365,3 +387,164 @@ fun CaptureButton(
...
+                    tint = if (!enabled) {
+                        disableColor
+                    } else if (toggleState == ToggleState.Left) {
+                        iconSelectionColor
+                    } else {
+                        iconUnSelectionColor
+                    }
+                )
+                Icon(
+                    painter = rightIcon,
```
temcguir, line 483: Add this content discription as an argument to the composable with the default of `"rightIcon"`. We should be able to change it to be an accurate description of the icon.

temcguir (approved):
Thank you for following up with all the small changes :)

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
heytsai: Thank you, updated.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
heytsai: Thank you, done.

Issue #173, opened 2024-04-11
Single-stream crashes on some devices when HDR is enabled


temcguir:
When HDR is enabled, on the Galaxy S23 Ultra we get the following crash when switching to single-stream capture:
```
FATAL EXCEPTION: ShaderCopy
Process: com.google.jetpackcamera, PID: 9166
java.lang.IllegalStateException: glDrawArrays: GL error 0x502
	at com.google.jetpackcamera.domain.camera.effects.ShaderCopy.checkGlErrorOrThrow(ShaderCopy.kt:334)
	at com.google.jetpackcamera.domain.camera.effects.ShaderCopy.access$checkGlErrorOrThrow(ShaderCopy.kt:35)
	at com.google.jetpackcamera.domain.camera.effects.ShaderCopy$drawFrame$1.invoke(ShaderCopy.kt:159)
	at com.google.jetpackcamera.domain.camera.effects.ShaderCopy$drawFrame$1.invoke(ShaderCopy.kt:123)
	at com.google.jetpackcamera.domain.camera.effects.CopyingSurfaceProcessor$renderWithSurfaceRequest$2$2$1$outputRenderTarget$1.onDrawFrame(CopyingSurfaceProcessor.kt:221)
	at androidx.graphics.opengl.GLThread.requestRenderInternal(GLThread.kt:304)
	at androidx.graphics.opengl.GLThread.requestRender$lambda$19$lambda$18(GLThread.kt:193)
	at androidx.graphics.opengl.GLThread.$r8$lambda$886mosJXDbxspS18XIztAZcQfOA(Unknown Source:0)
	at androidx.graphics.opengl.GLThread$$ExternalSyntheticLambda1.run(D8$$SyntheticClass:0)
	at android.os.Handler.handleCallback(Handler.java:958)
	at android.os.Handler.dispatchMessage(Handler.java:99)
	at android.os.Looper.loopOnce(Looper.java:230)
	at android.os.Looper.loop(Looper.java:319)
	at android.os.HandlerThread.run(HandlerThread.java:67)
```

This device may have a graphic driver issue that needs a workaround. In the worst case, we can disable single-stream for this device when HDR is enabled.

PR #206, opened 2024-05-21
Support Ultra HDR image capture

heytsai:
This makes Ultra HDR image capture became available when HDR is enabled in quick settings if the device supports Ultra HDR:

  

VideoCapture is currently disabled when Ultra HDR is enabled for following reasons:
1. For single stream mode, CameraEffect does not support HDR related processing.
2. For multiple stream mode, "Preview + Ultra HDR ImageCapture + VideoCapture" will trigger. StreamSharing, which makes it not exactly multiple-stream.

Also for the above reasons, Ultra HDR and Single Stream Mode features are set to be mutually exclusive. An error message will be shown when trying to enable both of them:


domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
```diff
@@ -592,8 +616,8 @@ constructor(
         }
     }
 
-    override suspend fun setCaptureMode(captureMode: CaptureMode) {
-        currentSettings.update { old ->
+    override suspend fun setCaptureMode(captureMode: CaptureMode): Boolean {
+        return currentSettings.tryUpdate { old ->
```
temcguir, line 620: Calling `setCaptureMode` should change other settings besides `captureMode` if the combination isn't supported. So if the `captureMode` argument passed in is SINGLE_STREAM, then it should change the `imageFormat` to `JPEG` rather than failing.

Similarly, if you call `setImageFormat(JPEG_ULTRA_HDR)`, then it should change the `captureMode` in the settings to `MULTI_STREAM`.

temcguir (commented):
Can you make the text under the HDR icon in quicksettings just say "HDR" rather than "HLG10"?

I also don't have a device that supports UltraHDR and HLG10 video, so I can't test that case locally. But it seems to do the right thing for a device that supports only HLG10 video.

data/settings/src/main/java/com/google/jetpackcamera/settings/model/CameraAppSettings.kt
```diff
@@ -30,8 +30,16 @@ data class CameraAppSettings(
...
+    val targetFrameRate: Int = TARGET_FPS_AUTO,
+    val imageFormat: ImageOutputFormat = ImageOutputFormat.JPEG
+) {
+    fun isCombinationSupported(): Boolean {
+        // Makes Ultra HDR and Single Stream Mode features mutually exclusive, since single-stream
+        // mode uses CameraEffect, which does not support HDR-related functions.
+        return imageFormat != ImageOutputFormat.JPEG_ULTRA_HDR ||
+            captureMode != CaptureMode.SINGLE_STREAM
+    }
+}
```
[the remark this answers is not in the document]
heytsai: Thank you, updated with the following changes:
1. exposed the mutually exclusive information by using `supportedImageFormatsMap` (`Map<CaptureMode, Set<ImageOutputFormat>>`) instead of `supportedImageFormats` in `CameraConstraints`.
2. change mode to `SINGLE_STREAM` will reset `imageFormat` to `JPEG` instead of failing.
 
Since HDR video may still be supported in single stream mode, the HDR button may not be deactivated, the UI control approach is slightly different, disabling the toggle button instead of the quick settings button:
1. For capture mode switching: check image format and reset it if needed.
2. For image format switching: will not affect capture mode. Since the toggle button enable/disable is based on the `supportedImageFormatsMap`, the image format cannot be set to Ultra HDR in single-stream mode (`supportedImageFormats` contains only `JPEG`).
 
Screenshot of `Single-stream Mode + HDR`:
  

P.S.: Since the testing device (S24) crashes when enabling HDR video in single stream mode ([issue#173](https://github.com/google/jetpack-camera-app/issues/173)), the check is done with VideoCapture not actually bound.

PTAL.

domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
heytsai: Thank you, updated. Please refer to the previous reply.

heytsai:
Also, displays HDR icon text in QuickSettings as HDR instead of HLG10.

temcguir (approved):
Thanks for implementing this!

The only piece I feel is missing is disabling the CaptureMode toggle in quick settings when the ImageOutputFormat is set to UltraHDR. Might not be a big deal though; the current UX may feel fine. I just don't have a device to test it with.

heytsai:
Thank you Trevor!

IIRC, lab has devices that support Ultra HDR, but you would need some setups and also connect to them. I will share the information in chat in case you need.

heytsai:
Updated for fixing merge conlicts.

heytsai:
Sorry, found code and string resource that are not used, so I add an update to remove them.

PR #208, opened 2024-05-22
Toggle mute/unmute while recording

Kimblebee:
This PR turns the audio visualizer into a button to toggle off/on the incoming audio for an in-progress video recording.

**_some little notes on this PR_**
- `withAudioEnabled(true)` will **always** be applied as long as the Record Audio permission has been granted.
- `CameraXCameraUseCase` 's `recording` must call `.mute(true)` **after** starting for a recording to start off muted
- The muted/open mic icon on the visualizer is based on whether the incoming audio amplitude is zero or nonzero, not a some `on/off` uIState.

I was intentional about avoiding the terms "audioEnabled/audioDisabled" because it is not the same as muted audio. 
my thought went something like this:
* `withAudioEnabled(true)` the recorder is handed a mic with the choice of turning it on or off. 
* If the audio isn't enabled, the recorder was never handed a mic; there's nothing to turn on or off.


feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewViewModel.kt
```diff
@@ -163,6 +163,24 @@ class PreviewViewModel @AssistedInject constructor(
             cameraUseCase.setLensFacing(newLensFacing)
         }
     }
+    fun toggleAudioMuted() {
+        setAudioMuted(
+            !(previewUiState.value as PreviewUiState.Ready)
+                .currentCameraSettings
+                .audioMuted
+        )
+    }
```
temcguir, line 172: It looks like this isn't used. You can remove it.

temcguir (approved):

domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
```diff
@@ -512,6 +524,7 @@ constructor(
                         }
                     }
                 }
+        recording?.mute(currentSettings.value!!.audioMuted)
```
Jaehwa-Noh, line 527: I just wonder, does force unwrap `currentSettings.value!!` fine to use?

Kimblebee: Good point.

It should be safe to unwrap at this point as the currentSettings should have been initialized, i think it'd still be a good idea to use a safe call operator regardless.

thanks!

PR #210, opened 2024-05-23
Set ResolutionSelector for ImageCapture and Preview

davidjiagoogle:
VideoCapture.setResolutionSelector() is library-restricted and therefore isn't called


yasith (approved):

domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
```diff
@@ -692,13 +701,26 @@ constructor(
...
 
+    private fun getResolutionSelector(aspectRatio: AspectRatio): ResolutionSelector {
+        val aspectRatioStrategy = when(aspectRatio) {
+            AspectRatio.THREE_FOUR -> AspectRatioStrategy.RATIO_16_9_FALLBACK_AUTO_STRATEGY
+            AspectRatio.NINE_SIXTEEN -> AspectRatioStrategy.RATIO_16_9_FALLBACK_AUTO_STRATEGY
+            else -> AspectRatioStrategy.RATIO_16_9_FALLBACK_AUTO_STRATEGY
+        }
+        return ResolutionSelector.Builder().setAspectRatioStrategy(aspectRatioStrategy).build()
+    }
+
```
temcguir, line 723: For VideoCapture, we set the aspect ratio using [Recorder.Builder#setAspectRatio](https://developer.android.com/reference/androidx/camera/video/Recorder.Builder#setAspectRatio(int)).

That should be set as well.

temcguir (changes requested):

temcguir (approved):

PR #214, opened 2024-05-30
Aspect Ratio buttons test tags

davidjiagoogle:
Adding res Ids to the buttons so the IQ test may use them


feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/quicksettings/ui/TestTags.kt
```diff
@@ -20,5 +20,7 @@ const val QUICK_SETTINGS_DROP_DOWN = "QuickSettingsDropDown"
 const val QUICK_SETTINGS_HDR_BUTTON = "QuickSettingsHdrButton"
 const val QUICK_SETTINGS_FLASH_BUTTON = "QuickSettingsFlashButton"
 const val QUICK_SETTINGS_FLIP_CAMERA_BUTTON = "QuickSettingsFlipCameraButton"
+const val QUICK_SETTINGS_RATIO_3_4_BUTTON = "QuickSettingsAspectRatio3:4Button"
+const val QUICK_SETTINGS_RATIO_9_16_BUTTON = "QuickSettingsAspectRatio9:16Button"
 const val QUICK_SETTINGS_RATIO_1_1_BUTTON = "QuickSettingsAspectRatio1:1Button"
```
temcguir, line 25: Just realized these should all be "QuickSettingsRatioX:XButton" instead of "QuickSettingsAspectRatioX:XButton", in order to match the field names.

davidjiagoogle: Done.

temcguir (approved):

PR #215, opened 2024-05-31
permissions flow test

Kimblebee:
Tests written for permissions UI flow

Also added some reusable utilities for interacting with permissions dialogs and finding system UI components by text.

Test orchestration is required for each permissions test. Each test must be isolated to its own instrumentation since permissions are being granted and reset during the tests. resetting or revorking permissions between tests isn't supported for testing in a single instrumentation -- it can cause unexpected, flaky behavior when tests are run in sequence.


app/src/androidTest/java/com/google/jetpackcamera/UiTestUtil.kt
```diff
@@ -76,3 +87,98 @@ fun ComposeTestRule.getCurrentLensFacing(): LensFacing {
...
+    if (Build.VERSION.SDK_INT >= 23) {
+        println("sdk version ${Build.VERSION.SDK_INT}")
+        val denyPermission = findObjectByText(
+            uiDevice = uiDevice,
+            text = when {
+                // todo fix this to be "Don't allow"... lol
+                // for some reason.... cannot find when the string has an apostrophe...
+                // i tried using \'...
+
+                Build.VERSION.SDK_INT >= 30 -> "Don"
```
temcguir, line 155: Similar to above, we should try to get these strings by res ID rather than relying on the text.

(btw, one way you might try escaping the apostrophe is with a unicode char: `Don\u2019t allow`)

app/build.gradle.kts
```diff
@@ -82,6 +87,14 @@ android {
...
             }
         }
+
+        unitTests.all {
+            it.testLogging {
+                events("failed", "standardOut", "standardError") // Log events
+                exceptionFormat = org.gradle.api.tasks.testing.logging.TestExceptionFormat.FULL // Show full stack traces
+                showStackTraces = true
+            }
+        }
```
temcguir, line 97: AFAIK, `unitTests` shouldn't affect instrumentation tests. So this shouldn't have any effect. Are you actually seeing this do something?

.idea/androidTestResultsUserPreferences.xml
temcguir, line 1: Delete

.idea/deploymentTargetSelector.xml
temcguir, line 1: Delete

feature/permissions/src/test/java/com/google/jetpackcamera/permissions/ExampleUnitTest.kt
temcguir, line 1: This file can be deleted.

feature/permissions/src/main/java/com/google/jetpackcamera/permissions/ui/TestTags.kt
```diff
@@ -0,0 +1,20 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.jetpackcamera.permissions.ui
+
+const val REQUEST_PERMISSION_BUTTON_TAG = "RequestPermissionTag"
+const val CAMERA_PERMISSION_TAG = "CameraPermissionTag"
+const val AUDIO_RECORD_PERMISSION_TAG = "RecordAudioPermissionTag"
```
temcguir, line 20: We're trying to keep these constant names in sync with the string literals:
`REQUEST_PERMISSION_BUTTON` -> `"RequestPermissionButton"`

app/src/androidTest/java/com/google/jetpackcamera/UiTestUtil.kt
```diff
@@ -76,3 +87,98 @@ fun ComposeTestRule.getCurrentLensFacing(): LensFacing {
...
+ *  Clicks ALLOW option on an open permission dialog
+ */
+fun grantPermissionDialog(uiDevice: UiDevice) {
+    if (Build.VERSION.SDK_INT >= 23) {
+        val allowPermission = findObjectByText(
+            uiDevice = uiDevice,
+            text = when {
+                Build.VERSION.SDK_INT == 23 -> "Allow"
+                Build.VERSION.SDK_INT <= 28 -> "ALLOW"
+                Build.VERSION.SDK_INT == 29 -> "Allow only while using the app"
```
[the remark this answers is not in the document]
Kimblebee: Sounds good.

is it possible for this resource id value to be different depending on the device manufacturer? It seems it has changed once for allow and deny after api 30.

So far the deny and allow once id's work. but uiautomator still can't find the `Allow` button. Not sure why that is but i'll take another minute to look into it.

app/src/androidTest/java/com/google/jetpackcamera/utils/UiTestUtil.kt
```diff
@@ -77,3 +90,129 @@ fun ComposeTestRule.getCurrentLensFacing(): LensFacing {
...
+                return wrappedRule.apply(base, description)
+            }
+        }
+        // If no match, return the base statement without granting permissions
+        return base
+    }
+}
+
+// functions for interacting with system permission dialog
+@SdkSuppress(minSdkVersion = 30)
```
temcguir, line 120: nit: `@SdkSuppress` is only needed on methods annotation with `@Test`. This one isn't needed.

app/src/androidTest/java/com/google/jetpackcamera/utils/UiTestUtil.kt
```diff
@@ -77,3 +90,129 @@ fun ComposeTestRule.getCurrentLensFacing(): LensFacing {
...
+fun grantPermissionDialog(uiDevice: UiDevice) {
+    if (Build.VERSION.SDK_INT >= 23) {
+        Log.d(TAG, "Searching for Allow Button...")
+
+        val allowPermission = findObjectById(
+            uiDevice = uiDevice,
+            resId = when {
+                Build.VERSION.SDK_INT <= 29 ->
+                    "com.android.packageinstaller:id/permission_allow_button"
+                else -> "com.android.permissioncontroller:id/permission_allow_button"
```
temcguir, line 148: It's possible on some devices it could be `permission_allow_foreground_only_button`, which corresponds to "Allow while the app is running".

app/src/androidTest/java/com/google/jetpackcamera/utils/UiTestUtil.kt
```diff
@@ -77,3 +90,129 @@ fun ComposeTestRule.getCurrentLensFacing(): LensFacing {
...
+
+        denyPermission!!.click()
+    }
+}
+
+/**
+ * Finds a system button by its resource ID.
+ * fails if not found
+ */
+fun findObjectById(
```
yasith, line 182: nit: private

app/src/androidTest/java/com/google/jetpackcamera/utils/UiTestUtil.kt
```diff
@@ -77,3 +90,129 @@ fun ComposeTestRule.getCurrentLensFacing(): LensFacing {
...
+    } else {
+        uiDevice.findObject(selector)
+    }
+}
+
+/**
+ * Finds a system button by its string value.
+ * fails if not found
+ */
+fun findObjectByText(
```
yasith, line 203: nit: private

yasith (approved):

app/src/androidTest/java/com/google/jetpackcamera/utils/UiTestUtil.kt
Kimblebee: unused, so removed it. finding by text can cause issues if the device is in another language, so best to stick with the id's

app/src/androidTest/java/com/google/jetpackcamera/PermissionsTest.kt
```diff
@@ -0,0 +1,177 @@
...
+    fun noPermissionsScreenWhenAlreadyGranted() {
+        runScenarioTest<MainActivity> {
+            composeTestRule.waitUntil(timeoutMillis = APP_START_TIMEOUT_MILLIS) {
+                composeTestRule.onNodeWithTag(CAPTURE_BUTTON).isDisplayed()
+            }
+        }
+    }
+
+    @Test
+    fun grantedCameraPermissionClosesPage() = runScenarioTest<MainActivity> {
```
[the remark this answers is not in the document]
Kimblebee: I'm specifically testing for the camera permission closing the page. I guess the last line was just an extra verification check, so i've removed it.

 If future permissions (i.e. location for metadata) were to be added, the order of what permissions comes after camera could change.

feature/permissions/src/main/java/com/google/jetpackcamera/permissions/ui/TestTags.kt
```diff
@@ -0,0 +1,20 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.jetpackcamera.permissions.ui
+
+const val REQUEST_PERMISSION_BUTTON = "RequestPermissionButton"
+const val CAMERA_PERMISSION_BUTTON = "CameraPermissionButton"
+const val AUDIO_RECORD_PERMISSION_BUTTON = "RecordAudioPermissionButton"
```
Jaehwa-Noh, line 20: I think this need to sync constant names and string literals also.

PR #216, opened 2024-06-03
Add a new product flavor for using preview SDKs.

yasith:
Also changed it so we only build the stable debug targets in CI


.github/workflows/PullRequestWorkflow.yaml
temcguir, line 1: Do we need any changes to MergeToMainWorkflow or are we OK with the current `assemble` output?

temcguir (approved):

PR #218, opened 2024-06-03
Video audio test

Kimblebee:
first test for video audio based on main branch.

The test just looks for the presence of the audio visualizer's open-mic icon (tagged as `AMPLITUDE_HOT_TAG`) while recording.

This test **must** run on a physical device. The emulators don't currently mock audio input. 🎙️ 

The open-mic icon's presence is determined by the incoming audio amplitude. _So as long as long as the device isn't in some kind of sensory deprivation chamber or the dead of space ~ it should pass._ 👍🏾 

⭐ *Also moved Utils files into its own subfolder, so test classes depending on it have updated imports*


feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/TestTags.kt
```diff
@@ -23,3 +23,5 @@ const val VIDEO_CAPTURE_EXTERNAL_UNSUPPORTED_TAG = "ImageCaptureExternalUnsuppor
 const val PREVIEW_DISPLAY = "PreviewDisplay"
 const val SCREEN_FLASH_OVERLAY = "ScreenFlashOverlay"
 const val SETTINGS_BUTTON = "SettingsButton"
+const val AMPLITUDE_NONE_TAG = "AudioAmplitudeNoneTag"
```
temcguir, line 26: nit: change string to match tag: "AmplitudeNoneTag".

app/src/androidTest/java/com/google/jetpackcamera/VideoAudioTest.kt
```diff
@@ -0,0 +1,83 @@
...
+                5000
+            )
+        }
+    }
+}
+
+// todo assert device has audio settings enabled
+// todo mute while recording
+// todo unmute while recording
+// todo start recording muted
```
temcguir, line 83: If you can't implement these in this PR, Instead of leaving these TODOs here, file an issue so we don't lose track.

app/src/androidTest/java/com/google/jetpackcamera/VideoAudioTest.kt
```diff
@@ -0,0 +1,83 @@
...
+    @Test
+    fun audioIncomingWhenEnabled() {
+        runScenarioTest<MainActivity> {
+            // check audio visualizer composable for muted/unmuted icon.
+            // icon will only be unmuted if audio is nonzero
+            composeTestRule.waitUntil(timeoutMillis = APP_START_TIMEOUT_MILLIS) {
+                composeTestRule.onNodeWithTag(CAPTURE_BUTTON).isDisplayed()
+            }
+
+            // record video
```
temcguir, line 67: We're eventually going to need to clean up these videos so they don't fill up the available space on test devices. Can you file an issue for that?

temcguir (approved):

PR #219, opened 2024-06-04
Fix ImageCaptureDeviceTest

davidjiagoogle:
Convert some of the uiDevice.findObject() and uiDevice.wait() to composeTestRule. In external intent launches, the using composeTestRule.waitUtil() for the snackBar causes a strange behavior, so that one remains using uiDevice.wait().

Instead of the convoluted implementation before, added runScenarioTestForResult() for simpler code and better readability.

For image_capture() failure we were experiencing in presubmit, removed the assertion on deleteFileInDireAfterTimestamp() and assert the lastModifiedTime of the folder. I cannot repro the failure on my device, so I hope this may fix it.

### DIFF
--- a/app/src/androidTest/java/com/google/jetpackcamera/ImageCaptureDeviceTest.kt
+++ b/app/src/androidTest/java/com/google/jetpackcamera/ImageCaptureDeviceTest.kt
@@ -25,13 +25,10 @@ import androidx.compose.ui.test.isDisplayed
 import androidx.compose.ui.test.junit4.createEmptyComposeRule
 import androidx.compose.ui.test.onNodeWithTag
 import androidx.compose.ui.test.performClick
-import androidx.test.core.app.ActivityScenario
 import androidx.test.ext.junit.runners.AndroidJUnit4
 import androidx.test.platform.app.InstrumentationRegistry
 import androidx.test.rule.GrantPermissionRule
-import androidx.test.uiautomator.By
 import androidx.test.uiautomator.UiDevice
-import androidx.test.uiautomator.Until
 import com.google.jetpackcamera.feature.preview.ui.CAPTURE_BUTTON
 import com.google.jetpackcamera.feature.preview.ui.IMAGE_CAPTURE_FAILURE_TAG
 import com.google.jetpackcamera.feature.preview.ui.IMAGE_CAPTURE_SUCCESS_TAG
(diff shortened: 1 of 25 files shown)


app/src/androidTest/java/com/google/jetpackcamera/UiTestUtil.kt
```diff
@@ -37,6 +38,15 @@ inline fun <reified T : Activity> runScenarioTest(
     }
 }
 
+inline fun <reified T : Activity> runScenarioTestForResult(
+    intent: Intent,
+    crossinline block: ActivityScenario<T>.() -> Unit
+): ActivityScenario<T> {
```
temcguir, line 44: Could you just have this return `ActivityResult` instead of `ActivityScenario`? Just put in a `return scenario.result` after `scenario.apply(block)`.

app/src/androidTest/java/com/google/jetpackcamera/ImageCaptureDeviceTest.kt
```diff
@@ -51,71 +49,81 @@ internal class ImageCaptureDeviceTest {
     val permissionsRule: GrantPermissionRule =
         GrantPermissionRule.grant(*(APP_REQUIRED_PERMISSIONS).toTypedArray())
 
+    @get:Rule
+    val composeTestRule = createEmptyComposeRule()
+
     private val instrumentation = InstrumentationRegistry.getInstrumentation()
     private var activityScenario: ActivityScenario<MainActivity>? = null
```
temcguir, line 56: Does this need to be an instance variable or could it just be a local variable in each test?

davidjiagoogle: I just followed the convention in other test where this is an instance variable like this

temcguir: Sorry I'm seeing how Github is displaying my comment and it looks confusing. I'm talking about the `activityScenario` instance variable. It doesn't look like it's needed as an instance variable (or maybe at all).

app/src/androidTest/java/com/google/jetpackcamera/ImageCaptureDeviceTest.kt
```diff
@@ -51,71 +49,81 @@ internal class ImageCaptureDeviceTest {
...
+
+                composeTestRule.onNodeWithTag(CAPTURE_BUTTON)
+                    .assertExists()
+                    .performClick()
+                /* This line needs to be uiDevice.wait() instead of composeTestRule.waitUntil()
+                 * because the latter causes a strange behavior where the node is not found. */
+                uiDevice.wait(
+                    Until.findObject(By.res(IMAGE_CAPTURE_FAILURE_TAG)),
+                    IMAGE_CAPTURE_TIMEOUT_MILLIS
+                )
```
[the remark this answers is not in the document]
davidjiagoogle: Nevermind this. It turns out to be some confusion.

PR #223, opened 2024-06-10
Switch to low light boost mode, when enabled

yasith:


domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
```diff
@@ -769,9 +777,25 @@ constructor(
...
     override suspend fun setLowLightBoost(lowLightBoost: LowLightBoost) {
-        currentSettings.update { old ->
-            old?.copy(lowLightBoost = lowLightBoost)
+        when (lowLightBoost) {
+            LowLightBoost.DISABLED -> {
+                currentSettings.update { old ->
+                    old?.copy(
+                        lowLightBoost = lowLightBoost,
+                        captureMode = CaptureMode.MULTI_STREAM,
+                        flashMode = FlashMode.AUTO
```
Jaehwa-Noh, line 786: Is low light boost need to adjust for flash mode?

PR #231, opened 2024-06-17
Implement time limit to recordings

Kimblebee:
Implementation and setting to add a duration limit to recordings.
_Currently there is no UI indication of elapsed time or duration limit prior to recording._

Original Setting screen PR in #255 

`setDurationLimitMillis` [API](https://developer.android.com/reference/androidx/camera/video/MediaStoreOutputOptions.Builder#setDurationLimitMillis(long)) automatically stops camera recording when it reaches the time limit.

`RecordingElapsedTimeNanos` is piped up and ready to be used by the UI.

❗ `RecordingElapsedTimeNanos` cannot be reliably be used to update the `VideoRecordingState`, as the value may be slightly (few ms) smaller than when the camera actually stops recording. That means the VideoRecordingState will never update, as `RecordingElapsedTimeNanos` may never be >= the duration limit.
* This has called attention to flaws in `VideoRecordingState`, and will need to be refactored in a separate PR.

📷 **Screenshots/Demo**

Screenshot 2024-09-17 at 12 07 53 PM

Screenshot 2024-09-17 at 11 50 04 AM

https://github.com/user-attachments/assets/d7cfdeac-e1a0-4c71-ae79-9ac02e771e88


feature/settings/src/main/java/com/google/jetpackcamera/settings/ui/SettingsComponents.kt
```diff
@@ -274,48 +331,68 @@ fun CaptureModeSetting(
...
-                        onClick = { setTargetFps(fpsOption) },
-                        enabled = supportedFps.contains(fpsOption)
+                        text = stringResource(id = R.string.fps_selector_auto),
+                        selected = fpsUiState.currentSelection == FPS_AUTO,
+                        onClick = { setTargetFps(FPS_AUTO) },
+                        enabled = fpsUiState.fpsAutoState is SingleSelectableState.Selectable
                     )
+                    listOf(FPS_15, FPS_30, FPS_60).forEach { fpsOption ->
+                        SingleChoiceSelector(
+                            text = "%d".format(fpsOption),
```
Jaehwa-Noh, line 376: Just a question, when using merely simple `%d` like this code, is better than using formatting string?
https://developer.android.com/guide/topics/resources/string-resource#formatting-strings

Kimblebee: nope, i should be using a formatted string here. thanks for the catch!

Jaehwa-Noh: I've given a review on similar format in here https://github.com/google/jetpack-camera-app/pull/257#discussion_r1728031052
and got an answer.

Then I'd searched it for a while on internet and found this.
https://developer.android.com/reference/java/lang/String.html#format(java.lang.String,%20java.lang.Object[])
> Returns a formatted string using the specified format string and arguments.
>
>The locale always used is the one returned by [Locale.getDefault(Locale.Category)](https://developer.android.com/reference/java/util/Locale#getDefault(java.util.Locale.Category)) with [FORMAT](https://developer.android.com/reference/java/util/Locale.Category#FORMAT) category specified.

String.format had been always get default local for localization.
So It seems to be okay to use `"%d".format()`.

Thank you for answering.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraUseCase.kt
```diff
@@ -744,7 +747,12 @@ constructor(
                 MediaStoreOutputOptions.Builder(
                     application.contentResolver,
                     MediaStore.Video.Media.EXTERNAL_CONTENT_URI
-                )
+                ).apply {
+                    // apply duration limit if applicable
+                    if (currentSettings.value?.maxVideoDurationMillis != -1L) {
```
temcguir, line 752: nit: if `currentSettings.value` is `null`, then it won't be equal to `-1L`, which will trigger a `NullPointerException` where it calls `currentSettings.value!!` below.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraUseCase.kt
```diff
@@ -744,7 +747,12 @@ constructor(
...
+                        setDurationLimitMillis(currentSettings.value!!.maxVideoDurationMillis)
```
temcguir, line 753: Instead of reading from `currentSettings` here, let's move the `maxDurationMillis` to be an argument to this function. It should be sent as a property in the `VideoCaptureControlEvent.StartVideoRecording` event so that it is closer associated with calling `CameraUseCase.startVideoRecording`.

feature/settings/src/main/java/com/google/jetpackcamera/settings/ui/SettingsComponents.kt
```diff
@@ -189,6 +189,7 @@ fun DarkModeSetting(
     )
 }
 
+// todo uistate for flash
```
temcguir, line 192: Unneeded comment?

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/CameraControlsOverlay.kt
```diff
@@ -324,6 +326,11 @@ private fun CaptureButton(
 ) {
     val multipleEventsCutter = remember { MultipleEventsCutter() }
     val context = LocalContext.current
+
+    // TODO: b/361133784 - [JCA] Update VideoRecordingState when recording reaches time limit
+    // recording actually stops a few ms before its reflected in recordingElapsedTimeNanos, so a
+    // launchedEffect cannot depend on it to trigger the UI change
+
```
temcguir, line 333: Can this be removed?

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/CameraControlsOverlay.kt
```diff
@@ -307,6 +308,7 @@ private fun CaptureButton(
     isQuickSettingsOpen: Boolean,
     videoRecordingState: VideoRecordingState,
     modifier: Modifier = Modifier,
+    maxVideoDurationMillis: Long,
```
temcguir, line 311: It looks like this isn't needed anymore.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/CameraControlsOverlay.kt
```diff
@@ -269,6 +269,7 @@ private fun ControlsBottom(
                 previewUiState = previewUiState,
                 isQuickSettingsOpen = isQuickSettingsOpen,
                 videoRecordingState = videoRecordingState,
+                maxVideoDurationMillis = currentCameraSettings.maxVideoDurationMillis,
```
temcguir, line 272: It looks like this isn't needed anymore.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewViewModel.kt
```diff
@@ -662,6 +665,10 @@ class PreviewViewModel @AssistedInject constructor(
                     videoRecordingState = VideoRecordingState.INACTIVE
                 ) ?: old
             }
+            Log.d(
+                TAG,
+                (previewUiState.value as PreviewUiState.Ready).videoRecordingState.toString()
+            )
```
temcguir, line 671: Can this debug logging be removed? It looks like it could cause log spam.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraUseCase.kt
```diff
@@ -129,7 +129,10 @@ interface CameraUseCase {
     sealed interface OnVideoRecordEvent {
         data class OnVideoRecorded(val savedUri: Uri) : OnVideoRecordEvent
 
-        data class OnVideoRecordStatus(val audioAmplitude: Double) : OnVideoRecordEvent
+        data class OnVideoRecordStatus(
+            val audioAmplitude: Double,
+            val timeStamp: Long
```
temcguir, line 134: nit: `timeStampNanos`.

.idea/other.xml
temcguir, line 1: Looks like this file isnt' needed.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraUseCase.kt
Kimblebee: I am a bit confused why it would need to be in a `VideoCaptureControlEvent.StartVideoRecording` event  when `transientSettings` is available as in `RunVideoRecording()`

temcguir: It's a matter of organizing concepts and keeping things more deterministic. 

Since events are queued in the `videoCaptureControlEvents` Channel, if we put the duration limit in `TransientSessionSettings`, it's possible that the duration limit could change in the time between sending the `StartVideoRecording` event and when the event is actually serviced, making it less deterministic.

When you start a recording, you should know what the duration limit will be. By putting the duration limit in the event, when you read the code, you can clearly see "this event will start a recording with a limit of 5 seconds" rather than "this event will start a recording that may or may not be limited in duration".  This will reduce cognitive load of those who need to maintain code. They won't have to go searching for what the duration will be in the code.

.idea/other.xml
temcguir, line 1: No need to check this in.

.idea/runConfigurations.xml
temcguir, line 1: No need to check this in.

temcguir (approved):

PR #233, opened 2024-06-18
Hdr Ui Affordance Improvements

davidjiagoogle:
In this PR, two UI components are changed. They are HdrQuickSet and HdrCaptureModeToggleButton.

1. The differentiation between HdrDynamicRange and HdrImageOutputFormat is introduced. Before this change, we assume if HLG10 dynamic range is supported, JPEG_Ultra_HDR is also supported. While in practice this may be true, in theory the app should check each one individually. This means it's in theory possible for the ToggleButton to be disabled but set to image capture because HLG10 is not supported but JPEG/R is.

2. Add logics for ExternalImageCapture mode. If the app is in ExternalImageCapture mode, we check if JPEG/R is supported to determine whether HDR can be turned on. This is because in this mode, videoCapture is disable and therefore irrelevant. If HDR is supported and turned on, the app switch to JPEG/R and the toggle is set to ImageCapture and disabled.

3. When in ExternalImageCapture, the toggle is always display regardless of HDR is turned on or not, and the toggle is always on Image Capture and disabled. This is to show that video capture is disabled

4. When the toggle is disabled and user tries to switch, the appropriate toast error message is displayed

Behavior during external image capture:
https://github.com/google/jetpack-camera-app/assets/125502967/2c1b9ba6-8f72-4f83-99fe-1703514ec9bb


feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewViewModel.kt
```diff
@@ -259,6 +262,50 @@ class PreviewViewModel @AssistedInject constructor(
...
+            stringRes = R.string.toast_video_capture_external_unsupported
+            testTag = VIDEO_CAPTURE_EXTERNAL_UNSUPPORTED_TAG
+        } else if (!hdrDynamicRangeSupported) {
+            stringRes = R.string.toast_hdr_video_unsupported
+            testTag = HDR_VIDEO_UNSUPPORTED_TAG
+
+        } else {
+            stringRes = R.string.toast_hdr_photo_unsupported
+            testTag = HDR_IMAGE_UNSUPPORTED_TAG
+        }
```
Jaehwa-Noh, line 294: Is there any way to convert `if` into `when`?

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewScreen.kt
```diff
@@ -111,6 +111,7 @@ fun PreviewScreen(
             onChangeDynamicRange = viewModel::setDynamicRange,
             onLowLightBoost = viewModel::setLowLightBoost,
             onChangeImageFormat = viewModel::setImageFormat,
+            onToggleWhenDisabled = viewModel::showToastForDisabledHdrToggle,
```
temcguir, line 114: Rather than calling back into the ViewModel, we can just create a model class in the UI layer which the ViewModel can populate. Something like:

```
data class CaptureModeToggleUiState (
  val enabled : Boolean,
  val disabledReason: CaptureModeUnsupportedReason?
  val currentMode: ...
   // All other fields defining the state of the button
) {
  enum CaptureModeUnsupportedReason (val testTag: String, val reasonTextResId: Int) {
    // Enumerations for unsupported reasons
  }
}
```

We can then calculate all the fields needed in `PreviewViewModel` and add `CaptureModeToggleUiState` as a field to `PreviewUiState`.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/CaptureModeToggleUiState.kt
```diff
@@ -0,0 +1,41 @@
...
+        ),
+        HDR_VIDEO_UNSUPPORTED(HDR_VIDEO_UNSUPPORTED_TAG, R.string.toast_hdr_video_unsupported),
+        HDR_IMAGE_UNSUPPORTED(HDR_IMAGE_UNSUPPORTED_TAG, R.string.toast_hdr_photo_unsupported)
+    }
+
+    enum class CaptureToggleMode {
+        CAPTURE_TOGGLE_IMAGE,
+        CAPTURE_TOGGLE_VIDEO
+    }
+}
```
temcguir, line 41: Let's take this one step further to be in line with the advice of https://developer.android.com/topic/architecture/recommendations#viewmodel that says that you should use sealed classes if states are exclusive. This could be rewritten as:
```
sealed interface CaptureModeToggleUiState {
    data object Invisible : CaptureModeToggleUiState

    sealed interface Visible : CaptureModeToggleUiState {
        val currentMode: ToggleMode
    }

    data class Enabled(override val currentMode: ToggleMode) : Visible

    data class Disabled(
        override val currentMode: ToggleMode,
        val disabledReason: DisabledReason
    ) : Visible

    enum class DisabledReason(val testTag: String, val reasonTextResId: Int) {
        VIDEO_CAPTURE_EXTERNAL_UNSUPPORTED(
            VIDEO_CAPTURE_EXTERNAL_UNSUPPORTED_TAG,
            R.string.toast_video_capture_external_unsupported
        ),
        HDR_VIDEO_UNSUPPORTED(HDR_VIDEO_UNSUPPORTED_TAG, R.string.toast_hdr_video_unsupported),
        HDR_IMAGE_UNSUPPORTED(HDR_IMAGE_UNSUPPORTED_TAG, R.string.toast_hdr_photo_unsupported)
    }

    enum class ToggleMode {
        IMAGE,
        VIDEO
    }
}
```
(I've also shortened some of the names to make it a little less verbose)

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/CameraControlsOverlay.kt
```diff
@@ -335,41 +333,42 @@ private fun CaptureButton(
 }
 
 @Composable
-private fun HdrCaptureModeToggleButton(
-    initialImageFormat: ImageOutputFormat = ImageOutputFormat.JPEG,
-    supportedImageFormats: Set<ImageOutputFormat>? = null,
-    onChangeImageFormat: (ImageOutputFormat) -> Unit = {}
+private fun CaptureModeToggleButton(
+    uiState: CaptureModeToggleUiState,
```
temcguir, line 337: With the sealed interface change, this will become:
`uiState: CaptureModeToggleUiState.Visible`

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewViewModel.kt
```diff
@@ -259,6 +323,26 @@ class PreviewViewModel @AssistedInject constructor(
         }
     }
 
+    fun showToastForDisabledHdrToggle() {
+        val disabledReason =
+            (previewUiState.value as? PreviewUiState.Ready)?.captureModeToggleUiState
+                ?.disabledReason ?: throw RuntimeException(
+                "showing Hdr toggle disabled toast without disabled reason."
+            )
```
temcguir, line 331: Now you can have the `disabledReason` passed in as an argument to `showToastForDisabledHdrToggle` rather than polling the UiState.

That way we can avoid the `as?` (or any of the `?` operators) returning null in case the previewUiState has changed between tapping the disabled capture toggle and calling this method.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewViewModel.kt
```diff
@@ -259,6 +323,26 @@ class PreviewViewModel @AssistedInject constructor(
         }
     }
 
+    fun showToastForDisabledHdrToggle() {
```
temcguir, line 326: nit: rename to `showSnackBarForDisabledCaptureToggle`

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewUiState.kt
```diff
@@ -40,7 +40,8 @@ sealed interface PreviewUiState {
         val toastMessageToShow: ToastMessage? = null,
         val snackBarToShow: SnackbarData? = null,
         val lastBlinkTimeStamp: Long = 0,
-        val previewMode: PreviewMode
+        val previewMode: PreviewMode,
+        val captureModeToggleUiState: CaptureModeToggleUiState
```
temcguir, line 44: nit: Set this to default to `CaptureModeToggleUiState.Invisible` so you don't need to add it as an argument to all of the compose Preview functions.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/quicksettings/QuickSettingsScreen.kt
```diff
@@ -215,18 +220,26 @@ private fun ExpandedQuickSettingsUi(
...
-                                    PreviewMode.ExternalImageCaptureMode &&
-                                    previewUiState.systemConstraints.forCurrentLens(
-                                        currentCameraSettings
-                                    )
-                                        ?.let { it.supportedDynamicRanges.size > 1 } ?: false
+                                hdrImageFormat = currentCameraSettings.defaultHdrImageOutputFormat,
+                                hdrDynamicRangeSupported = cameraConstraints?.let
+                                    { it.supportedDynamicRanges.size > 1 } ?: false,
+                                hdrImageFormatSupported = cameraConstraints?.let {
+                                    it.supportedImageFormatsMap[CaptureMode.MULTI_STREAM]!!.size > 1
```
temcguir, line 240: Shouldn't this be `it.supportedImageFormatsMap[currentCameraSettings.captureMode]?.let { it.size > 1 }`?

SINGLE_STREAM mode will have only a single supported ImageCaptureFormat currently.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewViewModel.kt
```diff
@@ -110,14 +114,75 @@ class PreviewViewModel @AssistedInject constructor(
...
+        cameraAppSettings: CameraAppSettings
+    ): CaptureModeToggleUiState {
+        val cameraConstraints: CameraConstraints? = systemConstraints.forCurrentLens(
+            cameraAppSettings
+        )
+        val hdrDynamicRangeSupported = cameraConstraints?.let {
+            it.supportedDynamicRanges.size > 1
+        } ?: false
+        val hdrImageFormatSupported = cameraConstraints?.let {
+            it.supportedImageFormatsMap[CaptureMode.MULTI_STREAM]!!.size > 1
```
temcguir, line 140: Shouldn't this be `it.supportedImageFormatsMap[cameraAppSettings.captureMode]?.let { it.size > 1 }`?

SINGLE_STREAM mode will have only a single supported ImageCaptureFormat currently.

feature/preview/src/main/res/values/strings.xml
```diff
@@ -29,6 +29,8 @@
     <string name="toast_video_capture_external_unsupported">External video capture not supported</string>
     <string name="stabilization_icon_description_preview_and_video">Preview is Stabilized</string>
     <string name="stabilization_icon_description_video_only">Only Video is Stabilized</string>
+    <string name="toast_hdr_photo_unsupported">HDR photo not supported</string>
+    <string name="toast_hdr_video_unsupported">HDR video not supported</string>
```
temcguir, line 33: These don't really give much information beyond what the toggle button UI already shows. Can we add more detailed information like:
* "UltraHDR photos not supported on this device" (when neither lens supports > 1 image capture mode for both single-stream and multi-stream)
* "UltraHDR photos not supported by current lens" (when both multi-stream and single-stream capture modes do not support > 1 image capture mode)
* "Multi-stream mode does not support UltraHDR photo capture for current lens" 
* "Single-stream mode does not support UltraHDR photo capture for current lens"
* "Video not supported while app is in image-only capture mode" (for this one you can just update the text in `toast_video_capture_external_unsupported`)
* "HDR video not supported on this device" (when neither lens supports HDR video)
* "HDR video not supported by current lens" (when only the current lens doesn't support HDR video)

All of this information should be available in the `SystemConstraints` and `CameraAppSettings` passed into `getCaptureToggleUiState`.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewViewModel.kt
```diff
@@ -259,6 +326,21 @@ class PreviewViewModel @AssistedInject constructor(
         }
     }
 
+    fun showSnackBarForDisabledHdrToggle(disabledReason: CaptureModeToggleUiState.DisabledReason) {
+        viewModelScope.launch {
+            _previewUiState.update { old ->
+                (old as? PreviewUiState.Ready)?.copy(
+                    snackBarToShow = SnackbarData(
+                        cookie = "DisabledHdrToggle",
```
temcguir, line 334: nit: this cookie should be unique (see how it is done for `captureImageInternal`). You could rename `imageCaptureCalledCount` to `snackbarCount` and reuse that if you want.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewUiState.kt
davidjiagoogle: I don't think this is true. If the previewMode is ExternalImageCapture, we'd still have to set this to Visible in PreviewViewModel's initialization.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewViewModel.kt
```diff
@@ -110,14 +116,116 @@ class PreviewViewModel @AssistedInject constructor(
...
+                    if (cameraConstraints?.let {
+                            it.supportedImageFormatsMap[captureMode]!!.size > 1
+                        } == true
+                    ) {
+                        if (lens == currentLensFacing) {
+                            Preconditions.checkState(captureMode != currentCaptureMode)
+                            when (captureMode) {
+                                CaptureMode.MULTI_STREAM ->
+                                    CaptureModeToggleUiState.DisabledReason
+                                        .HDR_IMAGE_UNSUPPORTED_ON_SINGLE_STREAM
```
temcguir, line 200: This is never actually returned since it is in a `for`-loop.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewViewModel.kt
```diff
@@ -110,14 +116,116 @@ class PreviewViewModel @AssistedInject constructor(
...
+            }
+            CaptureModeToggleUiState.DisabledReason.HDR_IMAGE_UNSUPPORTED_ON_DEVICE
+        } else if (!hdrDynamicRangeSupported) {
+            for (lens in LensFacing.values()) {
+                if (lens == currentLensFacing) {
+                    continue
+                }
+                val cameraConstraints = systemConstraints.perLensConstraints[lens]
+                if (cameraConstraints?.let { it.supportedDynamicRanges.size > 1 } == true) {
+                    CaptureModeToggleUiState.DisabledReason.HDR_VIDEO_UNSUPPORTED_ON_LENS
```
temcguir, line 220: This is also never returned since it's in a `for`-loop.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewViewModel.kt
```diff
@@ -110,14 +116,116 @@ class PreviewViewModel @AssistedInject constructor(
...
+        if (previewMode is PreviewMode.ExternalImageCaptureMode) {
+            return CaptureModeToggleUiState.DisabledReason.VIDEO_CAPTURE_EXTERNAL_UNSUPPORTED
+        }
+        return if (!hdrImageFormatSupported) {
+            for (lens in LensFacing.values()) {
+                val cameraConstraints = systemConstraints.perLensConstraints[lens]
+                for (captureMode in CaptureMode.values()) {
+                    if (cameraConstraints?.let {
+                            it.supportedImageFormatsMap[captureMode]!!.size > 1
+                        } == true
```
temcguir, line 193: You should avoid the `!!` operator (see go/kotlin-practices/null#avoid-not-null-assertion-operator).

You can replace this with:
```
if (cameraConstraints?.supportedImageFormatsMap?.get(captureMode)?.let { it.size > 1 } == true)
```

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewViewModel.kt
```diff
@@ -110,14 +116,116 @@ class PreviewViewModel @AssistedInject constructor(
...
+        cameraAppSettings: CameraAppSettings
+    ): CaptureModeToggleUiState {
+        val cameraConstraints: CameraConstraints? = systemConstraints.forCurrentLens(
+            cameraAppSettings
+        )
+        val hdrDynamicRangeSupported = cameraConstraints?.let {
+            it.supportedDynamicRanges.size > 1
+        } ?: false
+        val hdrImageFormatSupported = cameraConstraints?.let {
+            it.supportedImageFormatsMap[cameraAppSettings.captureMode]!!.size > 1
```
temcguir, line 142: You should avoid the `!!` operator (see go/kotlin-practices/null#avoid-not-null-assertion-operator).

You can replace this with:
```
val hdrImageFormatSupported = cameraConstraints?.supportedImageFormatsMap?.get(cameraAppSettings.captureMode)?.let { it.size > 1 } ?: false
```

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewViewModel.kt
```diff
@@ -110,14 +116,116 @@ class PreviewViewModel @AssistedInject constructor(
...
+        currentLensFacing: LensFacing,
+        currentCaptureMode: CaptureMode
+    ): CaptureModeToggleUiState.DisabledReason {
+        if (previewMode is PreviewMode.ExternalImageCaptureMode) {
+            return CaptureModeToggleUiState.DisabledReason.VIDEO_CAPTURE_EXTERNAL_UNSUPPORTED
+        }
+        return if (!hdrImageFormatSupported) {
+            for (lens in LensFacing.values()) {
+                val cameraConstraints = systemConstraints.perLensConstraints[lens]
+                for (captureMode in CaptureMode.values()) {
```
temcguir, line 190: Enum.values()' is recommended to be replaced by 'Enum.entries' since Kotlin 1.9

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewViewModel.kt
```diff
@@ -110,14 +116,116 @@ class PreviewViewModel @AssistedInject constructor(
...
+        hdrImageFormatSupported: Boolean,
+        systemConstraints: SystemConstraints,
+        currentLensFacing: LensFacing,
+        currentCaptureMode: CaptureMode
+    ): CaptureModeToggleUiState.DisabledReason {
+        if (previewMode is PreviewMode.ExternalImageCaptureMode) {
+            return CaptureModeToggleUiState.DisabledReason.VIDEO_CAPTURE_EXTERNAL_UNSUPPORTED
+        }
+        return if (!hdrImageFormatSupported) {
+            for (lens in LensFacing.values()) {
```
temcguir, line 188: Enum.values()' is recommended to be replaced by 'Enum.entries' since Kotlin 1.9

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewViewModel.kt
```diff
@@ -110,14 +116,116 @@ class PreviewViewModel @AssistedInject constructor(
...
+                            }
+                        } else {
+                            CaptureModeToggleUiState.DisabledReason.HDR_IMAGE_UNSUPPORTED_ON_LENS
+                        }
+                    }
+                }
+            }
+            CaptureModeToggleUiState.DisabledReason.HDR_IMAGE_UNSUPPORTED_ON_DEVICE
+        } else if (!hdrDynamicRangeSupported) {
+            for (lens in LensFacing.values()) {
```
temcguir, line 214: Enum.values()' is recommended to be replaced by 'Enum.entries' since Kotlin 1.9

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewViewModel.kt
```diff
@@ -110,14 +116,116 @@ class PreviewViewModel @AssistedInject constructor(
...
+        return if (!hdrImageFormatSupported) {
+            for (lens in LensFacing.values()) {
+                val cameraConstraints = systemConstraints.perLensConstraints[lens]
+                for (captureMode in CaptureMode.values()) {
+                    if (cameraConstraints?.let {
+                            it.supportedImageFormatsMap[captureMode]!!.size > 1
+                        } == true
+                    ) {
+                        if (lens == currentLensFacing) {
+                            Preconditions.checkState(captureMode != currentCaptureMode)
```
temcguir, line 196: It looks like we're iterating through all of the CaptureModes, so won't this crash eventually?

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewViewModel.kt
temcguir: Actually, instead of iterating through all of the `LensFacing` values, just use the entries from the systemConstraints.perLensConstraints` map. Something like 
```
systemConstraints.perLensConstraints.forEach { lens, constraints -> 
...
}
```

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewViewModel.kt
davidjiagoogle: Ah I forgot to remove this in the commit

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewUiState.kt
davidjiagoogle: I'm in favor of not setting a default value to make us think about the toggle state each time PreviewUiState is made since the logic for displaying/hiding has a lot of corner cases, and the fact that it's going to need another condition added to it when externalVideoCapture is implemented

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewViewModel.kt
```diff
@@ -360,7 +505,17 @@ class PreviewViewModel @AssistedInject constructor(
...
+                _previewUiState.update { old ->
+                    (old as? PreviewUiState.Ready)?.copy(
+                        captureModeToggleUiState =
+                        getCaptureToggleUiState(
+                            old.systemConstraints,
+                            cameraUseCase.getCurrentSettings().value!!
+                        )
+                    ) ?: old
+                }
+            }
```
temcguir, line 518: This `also` block shouldn't be needed. The updates from the CameraUseCase should be sufficient.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewViewModel.kt
```diff
@@ -110,14 +115,116 @@ class PreviewViewModel @AssistedInject constructor(
...
+                        cameraAppSettings.cameraLensFacing,
+                    )
+                )
+            }
+        } else {
+            CaptureModeToggleUiState.Invisible
+        }
+    }
+
+    @SuppressLint("RestrictedApi")
```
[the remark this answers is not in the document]
davidjiagoogle: This is no longer needed. It was added because a Preconditions check that was removed

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewViewModel.kt
```diff
@@ -154,6 +261,15 @@ class PreviewViewModel @AssistedInject constructor(
...
             cameraUseCase.setCaptureMode(captureMode)
+        }.also {
+            _previewUiState.update { old ->
+                (old as? PreviewUiState.Ready)?.copy(
+                    captureModeToggleUiState = getCaptureToggleUiState(
+                        old.systemConstraints,
+                        cameraUseCase.getCurrentSettings().value!!
+                    )
+                ) ?: old
+            }
```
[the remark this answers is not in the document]
davidjiagoogle: If user updates dynamicRange, imageOutputFormat, or CaptureMode, currentSettings is updated by CameraXCameraUseCase, but CaptureToggleUiState isn't. But changing any of the three could result in CaptureToggleUiState changing, therefore shouldn't we also update it here?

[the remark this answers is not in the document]
davidjiagoogle: Otherwise CaptureToggleUiState is only updated in the init block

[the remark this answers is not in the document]
temcguir: The init block is basically a `while` loop (with suspending wait). It listens for all changes from `CameraXCameraUseCase` and will update `CaptureToggleUiState` whenever any of the other settings change.

We probably want to restructure the init block a little bit since it also recalculates `CaptureToggleUiState` whenever the zoom level changes, but that can happen in a different PR.

temcguir (commented):
See existing inline comment. We should try to avoid updating the `PreviewUiState` directly in `PreviewViewModel` setters if it can be derived from `CameraAppSettings` / `Constraints`. That will happen in the `Flow` in the `init` block. This allows us to use a single source of truth. So for instance, if you tried to set the capture mode but the camera has other constraints that prevent it, we won't accidentally put the `PreviewUiState` into a state that doesn't actually represent what the camera is doing.

temcguir (approved):

PR #236, opened 2024-06-25
Implement LowLightBoost Capture

yasith:
This PR contains the following.

- Low Light Boost(LLB) feature toggle is enabled, if the device supports LLB
- When LLB toggle is turned on, switch to multi-stream mode, and enable LLB capture with Camera2Interop. Show the outlined LLB indicator with 50% opacity
- Query for the LLB active status `CONTROL_LOW_LIGHT_BOOST_STATE_ACTIVE`, if active change the outlined LLB indicator to a filled indicator with full opacity

### DIFF
--- a/domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraUseCase.kt
+++ b/domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraUseCase.kt
@@ -73,6 +73,8 @@ interface CameraUseCase {
 
     fun getZoomScale(): StateFlow<Float>
 
+    fun getLowLightBoostActiveStatus(): StateFlow<Boolean>
+
     fun getSurfaceRequest(): StateFlow<SurfaceRequest?>
 
     fun getScreenFlashEvents(): SharedFlow<ScreenFlashEvent>
--- a/domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
+++ b/domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
@@ -20,8 +20,13 @@ import android.app.Application
 import android.content.ContentResolver
 import android.content.ContentValues
 import android.content.pm.PackageManager
+import android.hardware.camera2.CameraCaptureSession
+import android.hardware.camera2.CameraCaptureSession.CaptureCallback
 import android.hardware.camera2.CameraCharacteristics
 import android.hardware.camera2.CameraMetadata
+import android.hardware.camera2.CaptureRequest
+import android.hardware.camera2.CaptureResult
+import android.hardware.camera2.TotalCaptureResult
 import android.net.Uri
 import android.os.Build
 import android.os.Environment
@@ -29,6 +34,7 @@ import android.provider.MediaStore
 import android.util.Log
 import android.util.Range
 import androidx.camera.camera2.interop.Camera2CameraInfo
+import androidx.camera.camera2.interop.Camera2Interop
 import androidx.camera.camera2.interop.ExperimentalCamera2Interop
 import androidx.camera.core.AspectRatio.RATIO_16_9
 import androidx.camera.core.AspectRatio.RATIO_4_3
@@ -712,6 +718,7 @@ constructor(
         supportedStabilizationModes: Set<SupportedStabilizationMode>,
         effect: CameraEffect? = null
     ): UseCaseGroup {
+        Log.d(TAG, "createUseCaseGroup")
         val previewUseCase = createPreviewUseCase(sessionSettings, supportedStabilizationModes)
         imageCaptureUseCase = createImageUseCase(sessionSettings)
         if (!disableVideoCapture) {
@@ -780,9 +787,25 @@ constructor(
     }
 
     override suspend fun setLowLightBoost(lowLightBoost: LowLightBoost) {
-        currentSettings.update { old ->
-            old?.copy(lowLightBoost = lowLightBoost)
+        when (lowLightBoost) {
+            LowLightBoost.DISABLED -> {
+                currentSettings.update { old ->
+                    old?.copy(
+                        lowLightBoost = lowLightBoost,
+                        captureMode = CaptureMode.MULTI_STREAM,
+                        flashMode = FlashMode.AUTO
+                    )
+                }
+            }
+            LowLightBoost.ENABLED -> currentSettings.update { old ->
+                old?.copy(
+                    lowLightBoost = lowLightBoost,
+                    captureMode = CaptureMode.SINGLE_STREAM,
+                    flashMode = FlashMode.OFF,
+                )
+            }
         }
+
     }
 
     override suspend fun setAudioMuted(isAudioMuted: Boolean) {
@@ -840,10 +863,12 @@ constructor(
                 )
     }
 
+    @androidx.annotation.OptIn(ExperimentalCamera2Interop::class)
     private fun createPreviewUseCase(
         sessionSettings: PerpetualSessionSettings,
         supportedStabilizationModes: Set<SupportedStabilizationMode>
     ): Preview {
+        Log.d(TAG, "createPreviewUseCase")
         val previewUseCaseBuilder = Preview.Builder()
         // set preview stabilization
         if (shouldPreviewBeStabilized(sessionSettings, supportedStabilizationModes)) {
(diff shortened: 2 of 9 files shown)


domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
```diff
@@ -212,6 +225,20 @@ constructor(
...
+            Camera2CameraInfo
+                .from(cameraInfo)
+                .getCameraCharacteristic(CameraCharacteristics.CONTROL_AE_AVAILABLE_MODES)
+                ?.contains(
+                    CameraMetadata.CONTROL_AE_MODE_ON_LOW_LIGHT_BOOST_BRIGHTNESS_PRIORITY
+                )
+        }.any()
+        false -> false
+    }
+
```
temcguir, line 241: We should be checking for support on a per-lens basis. This can be rewritten as an extension property to make it easier:

```
private val CameraInfo.isLowLightBoostSupported: Boolean
    @androidx.annotation.OptIn(ExperimentalCamera2Interop::class)
    @OptIn(BuildCompat.PrereleaseSdkCheck::class)
    get() = BuildCompat.isAtLeastV() &&
        Camera2CameraInfo
            .from(this)
            .getCameraCharacteristic(CameraCharacteristics.CONTROL_AE_AVAILABLE_MODES)
            ?.contains(
                CameraMetadata.CONTROL_AE_MODE_ON_LOW_LIGHT_BOOST_BRIGHTNESS_PRIORITY
            ) ?: false
```

Then in the `CameraConstraints` (line 197), we can say:
```
lowLightBoostSupport = camInfo.isLowLightBoostSupported
```

domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
```diff
@@ -760,8 +792,23 @@ constructor(
...
 
     override suspend fun setLowLightBoost(lowLightBoost: LowLightBoost) {
-        currentSettings.update { old ->
-            old?.copy(lowLightBoost = lowLightBoost)
+        when (lowLightBoost) {
+            LowLightBoost.DISABLED -> {
+                currentSettings.update { old ->
+                    old?.copy(
+                        lowLightBoost = lowLightBoost,
+                        captureMode = CaptureMode.MULTI_STREAM,
```
temcguir, line 800: What if the `CaptureMode` was `SINGLE_STREAM` before disabling low-light boost?

If we can only support LLB in `SINGLE_STREAM` mode, then we should do something similar to what we did with `CameraConstraints.supportedImageFormatsMap` where we have a `Map<CaptureMode, Boolean>`. We can then determine in the UI whether we should allow users to enable LLB depending on which capture mode is enabled (otherwise we disable the button).

domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
```diff
@@ -834,6 +883,34 @@ constructor(
...
+                        val boostState = result.get(CaptureResult.CONTROL_LOW_LIGHT_BOOST_STATE)
+                        if (boostState == CameraMetadata.CONTROL_LOW_LIGHT_BOOST_STATE_ACTIVE) {
+                            Log.d(TAG, "LLB Active")
+                            _lowLightBoostActiveStatus.value = true
+                        } else {
+                            Log.d(TAG, "LLB Inactive")
+                            _lowLightBoostActiveStatus.value = false
+                        }
+                    }
+                })
```
temcguir, line 910: We should add this callback regardless of whether LLB is enabled. Technically the `CONTROL_AE_MODE` can be set at any time (without needing to rebind use cases). So it could go from the `DISABLED` -> `ENABLED` state without needing to recreate / rebind `Preview`.

domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
```diff
@@ -834,6 +883,34 @@ constructor(
...
         )
 
+        if (currentSettings.value?.lowLightBoost == LowLightBoost.ENABLED) {
+            Log.d(TAG, "Adding Control AE_MODE_ON_LOW_LIGHT_BOOST_BRIGHTNESS_PRIORITY")
+            val camera2Interop = Camera2Interop.Extender(previewUseCaseBuilder)
+            if (Build.VERSION.SDK_INT >= 35) {
+                camera2Interop.setCaptureRequestOption(
+                    CaptureRequest.CONTROL_AE_MODE,
+                    CameraMetadata.CONTROL_AE_MODE_ON_LOW_LIGHT_BOOST_BRIGHTNESS_PRIORITY
+                )
```
temcguir, line 893: This should be done using `Camera2CameraControl` rather than adding it to the `Preview.Builder`. It's similar to the `CameraControl.setZoomRatio` or `setFlashModeInternal` that we set in the `cameraProvider.runWith {}` block.

domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
```diff
@@ -834,6 +883,34 @@ constructor(
             getResolutionSelector(sessionSettings.aspectRatio)
         )
 
+        if (currentSettings.value?.lowLightBoost == LowLightBoost.ENABLED) {
```
temcguir, line 886: Rather than reading this from `currentSettings`, `lowLightBoost` should be added to `TransientSessionSettings` and pulled out of settings in the the `currentSettings` Flow in `runCamera()`. That way we know we have a valid snapshot whenever we rebind the camera.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewViewModel.kt
```diff
@@ -93,15 +93,17 @@ class PreviewViewModel @AssistedInject constructor(
             combine(
                 cameraUseCase.getCurrentSettings().filterNotNull(),
                 constraintsRepository.systemConstraints.filterNotNull(),
-                cameraUseCase.getZoomScale()
-            ) { cameraAppSettings, systemConstraints, zoomScale ->
+                cameraUseCase.getZoomScale(),
+                cameraUseCase.getLowLightBoostActiveStatus()
```
temcguir, line 97: We're going to start to run out of the ability to combine these flows if we keep adding more. `combine` only has overloads for up to 5 flows (with unique types). Since zoomScale and LLBActiveStatus are both descriptive camera state, maybe we can combine them into a single data class called `DescriptiveCameraState`? Then we can have a single flow for both from `CameraUseCase`.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
```diff
@@ -358,14 +358,18 @@ fun StabilizationIcon(
...
+) {
     when (lowLightBoost) {
         LowLightBoost.ENABLED -> {
             Icon(
-                imageVector = Icons.Outlined.Nightlight,
+                imageVector = if (active) Icons.Filled.Nightlight else Icons.Outlined.Nightlight,
                 contentDescription =
                 stringResource(id = R.string.quick_settings_lowlightboost_enabled),
-                modifier = modifier.alpha(0.5f)
+                modifier = modifier.alpha(if (active) 1.0f else 0.5f)
```
temcguir, line 372: Disabled states should use an alpha of `0.38f`. See https://m3.material.io/foundations/interaction/states/applying-states#4aff9c51-d20f-4580-a510-862d2e25e931

temcguir (changes requested):

domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraUseCase.kt
```diff
@@ -73,6 +73,8 @@ interface CameraUseCase {
 
     fun getZoomScale(): StateFlow<Float>
 
+    fun getLowLightBoostActiveStatus(): StateFlow<Boolean>
```
jsaund, line 76: Have you considered defining an enum for this? Maybe `UNSUPPORTED, ACTIVE, INACTIVE` ?
That way the client only needs to observe this and not `isLowLightBoostSupport` and the default can be `UNSUPPORTED` rather than `true`

domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
```diff
@@ -541,6 +566,10 @@ constructor(
     private val _zoomScale = MutableStateFlow(1f)
     override fun getZoomScale(): StateFlow<Float> = _zoomScale.asStateFlow()
 
+    private val _lowLightBoostActiveStatus = MutableStateFlow(true)
```
jsaund, line 569: This would mean that low light is active?
a) maybe the enum approach I suggested above might be better?
b) maybe starting as `inactive` might be better?
c) if I got the active / inactive wrong then it might be better to type it to an enum to avoid confusion

domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
```diff
@@ -834,6 +881,34 @@ constructor(
             getResolutionSelector(sessionSettings.aspectRatio)
         )
 
+        if (currentSettings.value?.lowLightBoost == LowLightBoost.ENABLED) {
+            Log.d(TAG, "Adding Control AE_MODE_ON_LOW_LIGHT_BOOST_BRIGHTNESS_PRIORITY")
+            val camera2Interop = Camera2Interop.Extender(previewUseCaseBuilder)
+            if (Build.VERSION.SDK_INT >= 35) {
```
jsaund, line 887: nit: should we use `BuildCompat.isAtLeastV()` instead?

domain/camera/src/main/java/com/google/jetpackcamera/domain/camera/CameraXCameraUseCase.kt
```diff
@@ -834,6 +881,34 @@ constructor(
...
+                    CameraMetadata.CONTROL_AE_MODE_ON_LOW_LIGHT_BOOST_BRIGHTNESS_PRIORITY
+                )
+                camera2Interop.setSessionCaptureCallback(object : CaptureCallback() {
+                    override fun onCaptureCompleted(
+                        session: CameraCaptureSession,
+                        request: CaptureRequest,
+                        result: TotalCaptureResult
+                    ) {
+                        super.onCaptureCompleted(session, request, result)
+                        val boostState = result.get(CaptureResult.CONTROL_LOW_LIGHT_BOOST_STATE)
```
jsaund, line 899: Not sure if this will change anything for your use case but sharing for context:

`LOW_LIGHT_BOOST_STATE` is only active / inactive if `CaptureResult.CONTROL_AE_MODE` is `AE_ON_LOW_LIGHT_BOOST_BRIGHTNESS_PRIORITY`

So if the current camera feature combination wasn't supported when LLB was requested, the result may switch back to AE_MODE_ON

Or if the AE_MODE was not LLB, then maybe hide the moon icon? 

Unless if the UI keeps the moon icon there but in an INACTIVE state? When pressed it will turn on LLB? But that might not be intuitive because the scene could be well lit and still be inactive so it wouldn't be clear to the user when AE_MODE is LLB or auto.

PR #247, opened 2024-07-10
settings screen constraints

Kimblebee:
Enable and disable default settings based on lens constraints and other selected settings.

⚡ **Big changes**: 
* Each setting now has its own individual UiState located in `SettingsUiState`.
* `SettingsViewModel` sets each UiState based on device capabilities and other selected default settings

✏️ **Note:**  The UI constraints currently disregard the current camera settings (i.e. settings changed in quick settings screen) and only focus on the state of the default settings


feature/settings/src/main/java/com/google/jetpackcamera/settings/SettingEnabledState.kt
```diff
@@ -0,0 +1,64 @@
...
+    override val settingEnabledState: SettingEnabledState,
+    val optionsStates: Map<Int, SettingEnabledState>
+) : SettingUiState
+
+class FlipLensUiState(override val settingEnabledState: SettingEnabledState) : SettingUiState
+class StabilizationUiState(
+    override val settingEnabledState: SettingEnabledState,
+    val previewStabilizationState: SettingEnabledState,
+    val videoStabilizationState: SettingEnabledState
+) : SettingUiState
```
temcguir, line 54: One thing we want to avoid in a sealed class heirarchy is properties that are not used when we are in a certain state. So in this case, it doesn't make sense to have `isEnabled` as a property when we already have sealed interface subclasses for `Enabled` and `Disabled`. Additionally, in `FpsUiState`, we have `optionsStates` which may not be necessary when `settingsEnabledState` is `Disabled`.

Consider a hierarchy like this:
```
sealed interface SettingUiState {
    data object Disabled : SettingsUiState
    data class Enabled(
        val defaultFrontLensUiState: DefaultFrontLensUiState,
        val fpsUiState: FpsUiState,
        val stabilizationUiState: StabilizationUiState
       // ... Additional properties for each setting
    ) : SettingUiState
}

sealed interface SingleSelectableState {
    data object Selectable : SingleSelectableState
    data class Disabled(val disabledRationale: DisabledRationale) : SingleSelectableState
}

// Doesn't really need to be a sealed interface since even when disabled,
// we still need the radio button to either be on or off.
data class DefaultLensFacingUiState(
    val useFrontLens: Boolean,
    val selectable: SingleSelectableState
)

sealed interface FpsUiState {
    data class Enabled(
        val currentSelection: Int,
        val fpsAutoState: SingleSelectableState,
        val fpsFifteenState: SingleSelectableState,
        val fpsThirtyState: SingleSelectableState,
        val fpsSixtyState: SingleSelectableState,
        // Contains text like "Selected FPS only supported by rear lens"
        val additionalContext: String = "",
    ) : FpsUiState

    // FPS selection completely disabled. Cannot open dialog.
    data class Disabled(val disabledRationale: DisabledRationale) : FpsUiState
}

sealed interface StabilizationUiState {
    data class Enabled(
        val currentSelection: Int,
        val stabilizationOnState: SingleSelectableState,
        val stabilizationHighQualityState: SingleSelectableState,
        // Contains text like "Selected stabilization mode only supported by rear lens"
        val additionalContext: String = ""
    ) : StabilizationUiState

    // Stabilization selection completely disabled. Cannot open dialog.
    data class Disabled(val disabledRationale: DisabledRationale) : StabilizationUiState
}

data class DisabledRationale(val testTag: String, val reasonText: String)
```

If we really wanted to, we could reduce duplication by turning `FpsUiState` and `StabilizationState` into a shared generic sealed interface such as `RadioButtonSettingUiState`, but then we would also need to replace all of the properties with something like a `Map`. As it is, it's a bit more readable and is more flexible if we need to add special behavior to the individual settings.

feature/settings/src/main/java/com/google/jetpackcamera/settings/SettingEnabledState.kt
```diff
@@ -0,0 +1,64 @@
...
+
+enum class DisabledRationale(val testTag: String, val reasonTextResId: Int) {
+    DEVICE_UNSUPPORTED(
+        DEVICE_UNSUPPORTED_TAG,
+        R.string.device_unsupported
+    ),
+    FPS(FPS_UNSUPPORTED_TAG, R.string.fps_unsupported),
+    STABILIZATION_UNSUPPORTED(STABILIZATION_UNSUPPORTED_TAG, R.string.stabilization_unsupported),
+    LENS_UNSUPPORTED(LENS_UNSUPPORTED_TAG, R.string.lens_unsupported)
+}
```
temcguir, line 64: This would probably be better as a data class. We may want the reason text to change depending on constraints. "The front lens doesn't support high quality stabilization". We can construct a string like this in the ViewModel using [formattable String resources](https://developer.android.com/guide/topics/resources/string-resource#formatting-strings), but if we use enum constants, then we are stuck with generic strings.

feature/settings/src/main/java/com/google/jetpackcamera/settings/SettingsViewModel.kt
```diff
@@ -51,14 +59,279 @@ class SettingsViewModel @Inject constructor(
         ) { updatedSettings, constraints ->
             SettingsUiState.Enabled(
                 cameraAppSettings = updatedSettings,
-                systemConstraints = constraints
+                systemConstraints = constraints,
```
temcguir, line 62: If we thoroughly define our `SettingsUiState`, we should no longer need the `CameraAppSettings` or `SystemConstraints` in `SettingsUiState`. The entire state of every setting will be calculated in the `ViewModel`, and the `SettingsUiState` will define exactly what should be shown in the UI.

Kimblebee: yup, i'm working on just that! It'll be in the updated the PR when I finish addressing your other comments and some other cleanup.

feature/settings/src/main/java/com/google/jetpackcamera/settings/SettingEnabledState.kt
Kimblebee: that sounds like a good idea

feature/settings/src/main/java/com/google/jetpackcamera/settings/ui/SettingsComponents.kt
```diff
@@ -361,40 +400,28 @@ private fun getStabilizationStringRes(
...
     // entire setting disabled when no available fps or target fps = 60
     // stabilization is unsupported >30 fps
     BasicPopupSetting(
         modifier = modifier,
         title = stringResource(R.string.video_stabilization_title),
         leadingIcon = null,
-        enabled = stabilizationUiState.settingEnabledState,
-        description = if (stabilizationUiState.settingEnabledState.isEnabled) {
+        enabled = stabilizationUiState is StabilizationUiState.Enabled,
+        description = if (stabilizationUiState is StabilizationUiState.Enabled) {
```
temcguir, line 414: nit: use `when (stabilizationUiState)` here so you get implicit type conversion and don't need to use explicit casting.

temcguir: This applies here and to the rest of this file.

feature/settings/src/main/java/com/google/jetpackcamera/settings/SettingEnabledState.kt
```diff
@@ -0,0 +1,130 @@
...
+import com.google.jetpackcamera.settings.model.FlashMode
+import com.google.jetpackcamera.settings.model.Stabilization
+import com.google.jetpackcamera.settings.ui.DEVICE_UNSUPPORTED_TAG
+import com.google.jetpackcamera.settings.ui.FPS_UNSUPPORTED_TAG
+import com.google.jetpackcamera.settings.ui.LENS_UNSUPPORTED_TAG
+import com.google.jetpackcamera.settings.ui.STABILIZATION_UNSUPPORTED_TAG
+
+sealed interface SingleSelectableState {
+    data object Selectable : SingleSelectableState
+    data class Disabled(val disabledRationale: Set<DisabledRationale>) : SingleSelectableState {
```
temcguir, line 30: Why use a `Set<DisabledRationale>` here instead of just `DisabledRationale`?

temcguir (commented):
Some quick comments. I may need to take another look.

feature/settings/src/main/java/com/google/jetpackcamera/settings/SettingEnabledState.kt
Kimblebee: there could be multiple factors that prevent an option from being enabled

for example, lets say 60 fps is supported by the rear, but not the front lens on a device.
if stabilization is on the front camera, the 60fps would be disabled because of the stabilization mode and the selected lens.
I figured leaving all the constraints it as a set would leave flexibility to display on the UI or debug.

I also wasn't entirely sure how i should prioritize which single DisabledRationale to display when there are multiple

temcguir: Just keep in mind this is UI state; it shouldn't necessarily be making the decision about which is the most important to show, the ViewModel should.

I can see the need to possibly want to show multiple reasons, but it's also possible that those reasons could be collated into a single reason in the ViewModel. So for instance instead of:
```
Disabled(
    setOf(
        DisabledRationale("The lens doesn't support 60 fps"),
        DisabledRationale("The current stabilization mode does not allow 60 fps")
    )
)
```

It could be:
```
Disabled(
    DisabledRationale("The lens and stabilization mode do not allow for 60fps")
)
```

Or, the ViewModel can just decide to show one of the DisabledRationale, and then when that rationale is no longer relevant, it can update the `Disabled` state to show the next DisabledRationale.

feature/settings/src/main/res/values/strings.xml
```diff
@@ -101,6 +101,11 @@
     <string name="fps_stabilization_disclaimer">*Available stabilization modes may change due to selected frame rate.</string>
     <string name="lens_stabilization_disclaimer">*Some devices may not support stabilization on both lens.</string>
 
+    <!-- disabled reason strings-->
+    <string name="device_unsupported">%1$s is unsupported by the device</string>
+    <string name="fps_unsupported"> %1$s is unsupported at %1$d fps</string>
```
temcguir, line 106: I think this should be `%1$d` -> `%2$d`. I get a format conversion exception otherwise.

feature/settings/src/main/res/values/strings.xml
```diff
@@ -101,6 +101,11 @@
...
+    <string name="stabilization_unsupported">%$1s is Unsupported by current stabilization</string>
```
temcguir, line 107: nit: Unsupported doesn't need to be capitalized.

feature/settings/src/main/java/com/google/jetpackcamera/settings/SettingsUiState.kt
```diff
@@ -15,16 +15,193 @@
...
+import com.google.jetpackcamera.settings.model.LensFacing
+import com.google.jetpackcamera.settings.model.Stabilization
+import com.google.jetpackcamera.settings.ui.DEVICE_UNSUPPORTED_TAG
+import com.google.jetpackcamera.settings.ui.FPS_60
+import com.google.jetpackcamera.settings.ui.FPS_UNSUPPORTED_TAG
+import com.google.jetpackcamera.settings.ui.LENS_UNSUPPORTED_TAG
+import com.google.jetpackcamera.settings.ui.STABILIZATION_UNSUPPORTED_TAG
+
+const val STABILIZATION_SETTING_PREFIX = "Stabilization"
+const val LENS_SETTING_PREFIX = "Lens flip"
```
temcguir, line 33: Instead of saying "Lens flip is unsupported ..." can we say "Front camera does not support..." or "Rear camera does not support...".

feature/settings/src/main/java/com/google/jetpackcamera/settings/ui/SettingsComponents.kt
```diff
@@ -634,6 +752,34 @@ fun SingleChoiceSelector(
...
+            disabledRationale.affectedSettingName
+        )
+
+        is DisabledRationale.StabilizationUnsupportedRationale -> context.getString(
+            disabledRationale.reasonTextResId,
+            disabledRationale.affectedSettingName
+        )
+    }
+}
+
```
temcguir, line 782: nit: since this is a `@Composable`, you don't need to use `LocalContext.current`. You can just use `stringResource`:
```
@Composable
@ReadOnlyComposable
fun disabledRationaleString(disabledRationale: DisabledRationale): String =
    when (disabledRationale) {
        is DisabledRationale.DeviceUnsupportedRationale -> stringResource(
            disabledRationale.reasonTextResId,
            disabledRationale.affectedSettingName
        )

        is DisabledRationale.FpsUnsupportedRationale -> stringResource(
                disabledRationale.reasonTextResId,
                disabledRationale.affectedSettingName,
                disabledRationale.currentFps
        )

        is DisabledRationale.LensUnsupportedRationale -> stringResource(
            disabledRationale.reasonTextResId,
            disabledRationale.affectedSettingName
        )

        is DisabledRationale.StabilizationUnsupportedRationale -> stringResource(
            disabledRationale.reasonTextResId,
            disabledRationale.affectedSettingName
        )
    }
```

feature/settings/src/main/java/com/google/jetpackcamera/settings/SettingsViewModel.kt
```diff
@@ -50,15 +62,311 @@ class SettingsViewModel @Inject constructor(
...
+    private fun getFpsUiState(
+        systemConstraints: SystemConstraints,
+        cameraAppSettings: CameraAppSettings
+    ): FpsUiState {
+        val optionConstraintRationale: MutableMap<Int, SingleSelectableState> = mutableMapOf()
+
+        val deviceFrameRates: Set<Int> =
+            systemConstraints.perLensConstraints.values.fold(emptySet()) { union, constraints ->
+                union + constraints.supportedFixedFrameRates
+            }
```
temcguir, line 292: Since we now seem to be disabling settings based on the lens, I don't think we should do a union of all the lenses anymore. We should get the `deviceFrameRates` from the current `cameraLensFacing`.

feature/settings/src/main/java/com/google/jetpackcamera/settings/SettingsViewModel.kt
```diff
@@ -50,15 +62,311 @@ class SettingsViewModel @Inject constructor(
...
         )
 
+    private fun getStabilizationUiState(
+        systemConstraints: SystemConstraints,
+        cameraAppSettings: CameraAppSettings
+    ): StabilizationUiState {
+        val deviceStabilizations: Set<SupportedStabilizationMode> =
+            systemConstraints.perLensConstraints.values.fold(emptySet()) { union, constraints ->
+                union + constraints.supportedStabilizationModes
+            }
```
temcguir, line 87: Since we now seem to be disabling settings based on the lens, I don't think we should do a union of all the lenses anymore. We should get the `deviceStabilizations` from the current `cameraLensFacing`.

feature/settings/src/main/java/com/google/jetpackcamera/settings/SettingsUiState.kt
```diff
@@ -15,16 +15,218 @@
...
+    data class StabilizationUnsupportedRationale(override val affectedSettingNameResId: Int) :
+        DisabledRationale {
+        override val reasonTextResId = R.string.stabilization_unsupported
+        override val testTag = STABILIZATION_UNSUPPORTED_TAG
+    }
+
+    sealed interface LensUnsupportedRationale : DisabledRationale {
+        data class FrontLensUnsupportedRationale(override val affectedSettingNameResId: Int) :
+            LensUnsupportedRationale {
+            override val reasonTextResId: Int = R.string.current_lens_unsupported
```
temcguir, line 92: Is this supposed to be current or front?

feature/settings/src/main/java/com/google/jetpackcamera/settings/SettingsUiState.kt
```diff
@@ -15,16 +15,218 @@
...
+            override val testTag = LENS_UNSUPPORTED_TAG
+        }
+
+        data class RearLensUnsupportedRationale(override val affectedSettingNameResId: Int) :
+            LensUnsupportedRationale {
+            override val reasonTextResId: Int = R.string.current_lens_unsupported
```
temcguir, line 98: Is this supposed to be current or rear?

feature/settings/src/main/java/com/google/jetpackcamera/settings/ui/SettingsComponents.kt
```diff
@@ -634,6 +751,34 @@ fun SingleChoiceSelector(
     }
 }
 
+@Composable
+@ReadOnlyComposable
+fun disabledRationaleString(disabledRationale: DisabledRationale): String {
+    return when (disabledRationale) {
+        is DisabledRationale.DeviceUnsupportedRationale -> stringResource(
+
+            disabledRationale.reasonTextResId,
```
temcguir, line 760: This should be `stringResource(disabledRationale.reasonTextResId)`

feature/settings/src/main/java/com/google/jetpackcamera/settings/ui/SettingsComponents.kt
```diff
@@ -634,6 +751,34 @@ fun SingleChoiceSelector(
...
+            disabledRationale.affectedSettingNameResId
+        )
+
+        is DisabledRationale.FpsUnsupportedRationale -> stringResource(
+            disabledRationale.reasonTextResId,
+            disabledRationale.affectedSettingNameResId,
```
temcguir, line 766: This should be `stringResource(disabledRationale.reasonTextResId)`

feature/settings/src/main/java/com/google/jetpackcamera/settings/ui/SettingsComponents.kt
```diff
@@ -634,6 +751,34 @@ fun SingleChoiceSelector(
...
+            disabledRationale.currentFps
+        )
+
+        is DisabledRationale.LensUnsupportedRationale -> stringResource(
+            disabledRationale.reasonTextResId,
+            disabledRationale.affectedSettingNameResId
```
temcguir, line 772: This should be `stringResource(disabledRationale.reasonTextResId)`

feature/settings/src/main/java/com/google/jetpackcamera/settings/ui/SettingsComponents.kt
```diff
@@ -634,6 +751,34 @@ fun SingleChoiceSelector(
...
+        )
+
+        is DisabledRationale.StabilizationUnsupportedRationale -> stringResource(
+            disabledRationale.reasonTextResId,
+            disabledRationale.affectedSettingNameResId
```
temcguir, line 777: This should be `stringResource(disabledRationale.reasonTextResId)`

feature/settings/src/main/java/com/google/jetpackcamera/settings/SettingEnabledState.kt
temcguir: Which option are you referring to by "the latter"?

My suggestion is to remove the setOf. If you have multiple disabled rationales, just use the first one. If that disabled rationale goes away, then the SettingsUiState will be updated with the next disabled rationale.

Currently you're always using the first disabled rationale in the set, but if there is a reason in the future to use multiple, you can add an overload that takes a set.

Kimblebee: woops, by latter I was referring to the first snippet 
```
Disabled(
    setOf(
        DisabledRationale("The lens doesn't support 60 fps"),
        DisabledRationale("The current stabilization mode does not allow 60 fps")
    )
)
```
It makes sense to just add an overload down the line should it be necessary

temcguir (approved):

PR #248, opened 2024-07-18
External Video Capture

davidjiagoogle:
Adds an additional PreviewMode for External Video Capture. This PR adds the following features

- External Video Capture mode is entered when JCA is launched via MediaStore.ACTION_VIDEO_CAPTURE
- User can't capture images in this mode (although they can still change settings related to image capture such as screen flash)
- Video capturing in this mode with no uri supplied results in exception
- Video capture in this mode saves the captured video to the uri location and exits JCA with RESULT_OK
- The toggle ui would have the same behaviors as in external image capture mode except the toggle position and toast texts are for video
- The default aspect ratio is not changed unlike in External Image Capture


temcguir (commented):
One quick comment but still need to take another look.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraUseCase.kt
```diff
@@ -526,65 +536,75 @@ constructor(
...
-                    run {
-                        Log.d(TAG, onVideoRecordEvent.toString())
-                        when (onVideoRecordEvent) {
-                            is VideoRecordEvent.Finalize -> {
-                                when (onVideoRecordEvent.error) {
-                                    ERROR_NONE ->
-                                        onVideoRecord(
-                                            CameraUseCase.OnVideoRecordEvent.OnVideoRecorded
+            pendingRecord.start(callbackExecutor) { onVideoRecordEvent ->
+                run {
```
temcguir, line 577: I think this `run` block is unnecessary.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraUseCase.kt
```diff
@@ -248,7 +250,7 @@ constructor(
         currentSettings.value =
             cameraAppSettings
                 .tryApplyDynamicRangeConstraints()
-                .tryApplyAspectRatioForExternalCapture(externalImageCapture)
+                .tryApplyAspectRatioForExternalCapture(disableVideoCapture)
```
temcguir, line 253: We should also disable image capture when we're in external intent video capture mode, and handle it accordingly here. Maybe the `disableVideoCapture` boolean should be changed to a `BitSet` of use cases that are enabled / disabled.

Also, the aspect ratio in this method should be set to `SIXTEEN_NINE` explicitly in that case.

feature/preview/src/test/java/com/google/jetpackcamera/feature/preview/PreviewViewModelTest.kt
```diff
@@ -89,15 +89,15 @@ class PreviewViewModelTest {
     @Test
     fun startVideoRecording() = runTest(StandardTestDispatcher()) {
         previewViewModel.startCameraUntilRunning()
-        previewViewModel.startVideoRecording()
+        previewViewModel.startVideoRecording(null, true) {}
```
temcguir, line 92: Can we add a separate test for External Intent Video Capture mode, like we have for image capture?

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraUseCase.kt
```diff
@@ -123,11 +127,11 @@ interface CameraUseCase {
      * Represents the events for video recording.
      */
     sealed interface OnVideoRecordEvent {
-        object OnVideoRecorded : OnVideoRecordEvent
+        data class OnVideoRecorded(val savedUri: Uri) : OnVideoRecordEvent
```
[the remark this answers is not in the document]
aaa72: Thanks. I will align to this commit by "savedUri".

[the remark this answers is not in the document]
davidjiagoogle: Looks like they are the same now

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraUseCase.kt
davidjiagoogle: I tried that but not attaching imageCapture use case causes other problems that increases the complexity of this class by a lot.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraUseCase.kt
```diff
@@ -526,65 +536,75 @@ constructor(
...
-            MediaStoreOutputOptions.Builder(
-                application.contentResolver,
-                MediaStore.Video.Media.EXTERNAL_CONTENT_URI
-            )
-                .setContentValues(contentValues)
-                .build()
 
+        val pendingRecord = if (!ignoreUri) {
+            val fileOutputOptions = FileOutputOptions.Builder(
+                File(videoCaptureUri!!.getPath())
```
Kimblebee, line 542: should we force unwrap here? is it possible at all that `videoCaptureUri` is null?

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraUseCase.kt
```diff
@@ -67,7 +67,11 @@ interface CameraUseCase {
         ignoreUri: Boolean = false
     ): ImageCapture.OutputFileResults
 
-    suspend fun startVideoRecording(onVideoRecord: (OnVideoRecordEvent) -> Unit)
+    suspend fun startVideoRecording(
+        videoCaptureUri: Uri?,
+        ignoreUri: Boolean = false,
```
Kimblebee, line 72: nit: every reference to ignoreUri is `!ignoreUri`. Feels a little weird because i read it in my head like "not not this".
Would it be better to just name it something that starts positive like`shouldUseUri`and adjust the references appropriately?

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraUseCase.kt
davidjiagoogle: Instead of BitSet I used an enum with STANDARD, VIDEO_ONLY, and IMAGE_ONLY, since the three modes are mutually exclusive. If we introduce more variants in the future, we could consider using bitset

feature/preview/src/test/java/com/google/jetpackcamera/feature/preview/PreviewViewModelTest.kt
davidjiagoogle: Expanded VideoRecordingDeviceTest

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraUseCase.kt
davidjiagoogle: The problems have been resolved.

null check at line 521. videoCaptureUri can't be null here

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraUseCase.kt
davidjiagoogle: makes sense. Renamed and reversed the checks

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewViewModel.kt
```diff
@@ -232,9 +243,11 @@ class PreviewViewModel @AssistedInject constructor(
                 it.size > 1
             } ?: false
         val isShown = previewMode is PreviewMode.ExternalImageCaptureMode ||
+            previewMode is PreviewMode.ExternalVideoCaptureMode ||
             cameraAppSettings.imageFormat == ImageOutputFormat.JPEG_ULTRA_HDR ||
             cameraAppSettings.dynamicRange == DynamicRange.HLG10
         val enabled = previewMode !is PreviewMode.ExternalImageCaptureMode &&
+            previewMode !is PreviewMode.ExternalVideoCaptureMode &&
```
Jaehwa-Noh, line 250: I think
```
previewMode !is PreviewMode.ExternalImageCaptureMode &&
            previewMode !is PreviewMode.ExternalVideoCaptureMode
```
can replace to `previewMode is PreviewMode.StandardMode`.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/CameraControlsOverlay.kt
```diff
@@ -319,19 +331,45 @@ private fun CaptureButton(
...
             if (isQuickSettingsOpen) {
                 onToggleQuickSettings()
             }
         },
         onLongPress = {
-            onStartVideoRecording()
+            when (previewUiState.previewMode) {
+                is PreviewMode.StandardMode -> {
+                    onStartVideoRecording(null, false) {}
+                }
```
Jaehwa-Noh, line 352: This condition can remove because same logic with `else`.

app/src/androidTest/java/com/google/jetpackcamera/VideoRecordingDeviceTest.kt
```diff
@@ -104,6 +155,23 @@ internal class VideoRecordingDeviceTest {
...
+                    }
+                }
+                hasDeletedFile = true
+            }
+        }
+        return hasDeletedFile
+    }
+
     companion object {
         private val TEST_CONTRACT = object : ActivityResultContracts.TakePicture() {
```
temcguir, line 176: Can you rename this to IMAGE_TEST_CONTRACT?

app/src/androidTest/java/com/google/jetpackcamera/VideoRecordingDeviceTest.kt
```diff
@@ -70,6 +73,54 @@ internal class VideoRecordingDeviceTest {
             activityScenario!!.result
         }.register("key", TEST_CONTRACT) { result ->
             assert(!result)
+            assert(activityScenario!!.result.resultCode == Activity.RESULT_CANCELED)
+        }.launch(uri)
+    }
+
+    @Test
+    fun video_capture_external_intent() = run {
```
temcguir, line 81: The `= run` is unnecessary here.

app/src/androidTest/java/com/google/jetpackcamera/VideoRecordingDeviceTest.kt
```diff
@@ -70,6 +73,54 @@ internal class VideoRecordingDeviceTest {
...
+            )
+            activityScenario!!.result
+        }.register("key", VIDEO_TEST_CONTRACT) { _ ->
+            assert(activityScenario!!.result.resultCode == Activity.RESULT_OK)
+            deleteFilesInDirAfterTimestamp(timeStamp)
+        }.launch(uri)
+    }
+
+    @Test
+    fun image_capture_external_with_video_capture_intent() = run {
```
temcguir, line 104: The `= run` is unnecessary here.

app/src/androidTest/java/com/google/jetpackcamera/VideoRecordingDeviceTest.kt
```diff
@@ -104,6 +155,23 @@ internal class VideoRecordingDeviceTest {
         )
     }
 
+    private fun deleteFilesInDirAfterTimestamp(timeStamp: Long): Boolean {
+        var hasDeletedFile = false
+        for (file in File(ImageCaptureDeviceTest.DIR_PATH).listFiles()) {
```
temcguir, line 160: Shouldn't this use a DIR_PATH that uses `getExternalStoragePublicDirectory(Environment.DIRECTORY_MOVIES)`?

app/src/androidTest/java/com/google/jetpackcamera/VideoRecordingDeviceTest.kt
```diff
@@ -70,6 +73,54 @@ internal class VideoRecordingDeviceTest {
...
+            deleteFilesInDirAfterTimestamp(timeStamp)
+        }.launch(uri)
+    }
+
+    @Test
+    fun image_capture_external_with_video_capture_intent() = run {
+        val timeStamp = System.currentTimeMillis()
+        val uri = getTestUri(timeStamp)
+        getTestRegistry {
+            activityScenario = ActivityScenario.launchActivityForResult(it)
```
temcguir, line 108: Can you use `runScenarioTestForResult` like in ImageCaptureDeviceTest?

app/src/androidTest/java/com/google/jetpackcamera/VideoRecordingDeviceTest.kt
```diff
@@ -70,6 +73,54 @@ internal class VideoRecordingDeviceTest {
...
+            uiDevice.findObject(By.res(CAPTURE_BUTTON)).click()
+            uiDevice.wait(
+                Until.findObject(By.res(IMAGE_CAPTURE_EXTERNAL_UNSUPPORTED_TAG)),
+                5000
+            )
+            uiDevice.pressBack()
+            activityScenario!!.result
+        }.register("key", VIDEO_TEST_CONTRACT) { result ->
+            assert(!result)
+            assert(activityScenario!!.result.resultCode == Activity.RESULT_CANCELED)
```
temcguir, line 123: Use:
```
assertThat(result).isFalse()
assertThat(resultCode).isEqualTo(Activity.RESULT_CANCELLED)
```

app/src/androidTest/java/com/google/jetpackcamera/VideoRecordingDeviceTest.kt
```diff
@@ -70,6 +73,54 @@ internal class VideoRecordingDeviceTest {
...
+                5000
+            )
+            uiDevice.findObject(By.res(CAPTURE_BUTTON)).longClick()
+            uiDevice.wait(
+                Until.findObject(By.res(VIDEO_CAPTURE_SUCCESS_TAG)),
+                5000
+            )
+            activityScenario!!.result
+        }.register("key", VIDEO_TEST_CONTRACT) { _ ->
+            assert(activityScenario!!.result.resultCode == Activity.RESULT_OK)
```
temcguir, line 98: Use `assertThat(resultCode).isEqualTo(Activity.RESULT_OK)`

app/src/androidTest/java/com/google/jetpackcamera/VideoRecordingDeviceTest.kt
```diff
@@ -70,6 +73,54 @@ internal class VideoRecordingDeviceTest {
             activityScenario!!.result
         }.register("key", TEST_CONTRACT) { result ->
             assert(!result)
+            assert(activityScenario!!.result.resultCode == Activity.RESULT_CANCELED)
```
temcguir, line 76: Use:
```
assertThat(result).isFalse()
assertThat(resultCode).isEqualTo(Activity.RESULT_CANCELLED)
```

app/src/androidTest/java/com/google/jetpackcamera/VideoRecordingDeviceTest.kt
```diff
@@ -70,6 +73,54 @@ internal class VideoRecordingDeviceTest {
...
+            )
+            uiDevice.findObject(By.res(CAPTURE_BUTTON)).longClick()
+            uiDevice.wait(
+                Until.findObject(By.res(VIDEO_CAPTURE_SUCCESS_TAG)),
+                5000
+            )
+            activityScenario!!.result
+        }.register("key", VIDEO_TEST_CONTRACT) { _ ->
+            assert(activityScenario!!.result.resultCode == Activity.RESULT_OK)
+            deleteFilesInDirAfterTimestamp(timeStamp)
```
temcguir, line 99: Can we just delete the `uri` directly rather than relying on timestamp?

app/src/androidTest/java/com/google/jetpackcamera/VideoRecordingDeviceTest.kt
```diff
@@ -70,6 +73,54 @@ internal class VideoRecordingDeviceTest {
...
+        val timeStamp = System.currentTimeMillis()
+        val uri = getTestUri(timeStamp)
+        getTestRegistry {
+            activityScenario = ActivityScenario.launchActivityForResult(it)
+            // Wait for the capture button to be displayed
+            uiDevice.wait(
+                Until.findObject(By.res(CAPTURE_BUTTON)),
+                5000
+            )
+            uiDevice.findObject(By.res(CAPTURE_BUTTON)).longClick()
```
temcguir, line 91: How long is a `longClick()`? If we don't have a sufficiently long click, then the recording may fail due to not having enough data. Maybe we should use a different UiAutomator API to perform the long press?

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewViewModel.kt
```diff
@@ -159,6 +162,14 @@ class PreviewViewModel @AssistedInject constructor(
         }
     }
 
+    private fun getUseCaseMode(previewMode: PreviewMode): CameraUseCase.UseCaseMode {
```
temcguir, line 165: nit: When you have an exhaustive conversion function like this, you can make it an extension function:

```
private fun PreviewMode.toUseCaseMode() = when (this) {
    is PreviewMode.ExternalImageCaptureMode -> CameraUseCase.UseCaseMode.IMAGE_ONLY
    is PreviewMode.ExternalVideoCaptureMode -> CameraUseCase.UseCaseMode.VIDEO_ONLY
    is PreviewMode.StandardMode -> CameraUseCase.UseCaseMode.STANDARD
}
```

app/src/androidTest/java/com/google/jetpackcamera/VideoRecordingDeviceTest.kt
davidjiagoogle: No longer valid

app/src/androidTest/java/com/google/jetpackcamera/VideoRecordingDeviceTest.kt
davidjiagoogle: No longer valid

app/src/androidTest/java/com/google/jetpackcamera/VideoRecordingDeviceTest.kt
davidjiagoogle: No longer valid

app/src/androidTest/java/com/google/jetpackcamera/VideoRecordingDeviceTest.kt
```diff
@@ -47,73 +56,119 @@ internal class VideoRecordingDeviceTest {
...
+                up()
+            }
+    }
+
+    private fun idleForVideoDuration() {
+        try {
+            composeTestRule.waitUntil(timeoutMillis = VIDEO_DURATION_MILLIS) {
+                composeTestRule.onNodeWithTag("dummyTagForLongPress").isDisplayed()
+            }
+        } catch (e: ComposeTimeoutException) {
```
temcguir, line 166: We should eventually replace this with a check for the timestamp UI of the video recording duration. Could you add a TODO here?

temcguir (approved):

PR #250, opened 2024-07-23
Enable torch when video recording and flash mode is ON

aaa72:
* The torch will be enabled when receiving VideoRecordEvent.Start and disabled when receiving VideoRecordEvent.Finalize.
 * Add CameraXCamerauseCaseTest.

### DIFF
--- a/core/camera/build.gradle.kts
+++ b/core/camera/build.gradle.kts
@@ -94,9 +94,15 @@ dependencies {
     testImplementation(libs.mockito.core)
     androidTestImplementation(libs.androidx.junit)
     androidTestImplementation(libs.androidx.espresso.core)
+    androidTestImplementation(libs.kotlinx.coroutines.test)
+    androidTestImplementation(libs.rules)
+    androidTestImplementation(libs.truth)
 
     // Futures
     implementation(libs.futures.ktx)
+
+    // LiveData
+    implementation(libs.androidx.lifecycle.livedata)
 
     // CameraX
     implementation(libs.camera.core)
--- a/core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraUseCase.kt
+++ b/core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraUseCase.kt
@@ -36,6 +36,7 @@ import androidx.camera.camera2.interop.Camera2Interop
 import androidx.camera.camera2.interop.ExperimentalCamera2Interop
 import androidx.camera.core.AspectRatio.RATIO_16_9
 import androidx.camera.core.AspectRatio.RATIO_4_3
+import androidx.camera.core.Camera
 import androidx.camera.core.CameraEffect
 import androidx.camera.core.CameraInfo
 import androidx.camera.core.CameraSelector
@@ -49,6 +50,8 @@ import androidx.camera.core.ImageCaptureException
 import androidx.camera.core.Preview
 import androidx.camera.core.SurfaceOrientedMeteringPointFactory
 import androidx.camera.core.SurfaceRequest
+import androidx.camera.core.TorchState
+import androidx.camera.core.UseCase
 import androidx.camera.core.UseCaseGroup
 import androidx.camera.core.ViewPort
 import androidx.camera.core.resolutionselector.AspectRatioStrategy
@@ -62,8 +65,10 @@ import androidx.camera.video.Recording
 import androidx.camera.video.VideoCapture
 import androidx.camera.video.VideoRecordEvent
 import androidx.camera.video.VideoRecordEvent.Finalize.ERROR_NONE
+import androidx.concurrent.futures.await
 import androidx.core.content.ContextCompat
 import androidx.core.content.ContextCompat.checkSelfPermission
+import androidx.lifecycle.asFlow
 import com.google.jetpackcamera.core.camera.CameraUseCase.ScreenFlashEvent.Type
 import com.google.jetpackcamera.core.camera.effects.SingleSurfaceForcingEffect
 import com.google.jetpackcamera.settings.SettableConstraintsRepository
@@ -92,6 +97,7 @@ import kotlin.coroutines.ContinuationInterceptor
 import kotlin.math.abs
 import kotlin.properties.Delegates
 import kotlinx.atomicfu.atomic
+import kotlinx.coroutines.CompletableDeferred
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.asExecutor
@@ -168,8 +174,6 @@ constructor(
         imageCaptureExtender.setSessionCaptureCallback(captureCallback)
     }
 
-    private var videoCaptureUseCase: VideoCapture<Recorder>? = null
-    private var recording: Recording? = null
     private lateinit var captureMode: CaptureMode
     private lateinit var systemConstraints: SystemConstraints
     private var disableVideoCapture by Delegates.notNull<Boolean>()
@@ -178,6 +182,7 @@ constructor(
         MutableSharedFlow()
     private val focusMeteringEvents =
         Channel<CameraEvent.FocusMeteringEvent>(capacity = Channel.CONFLATED)
+    private var videoCaptureControlEvents: Channel<VideoCaptureControlEvent>? = null
 
     private val currentSettings = MutableStateFlow<CameraAppSettings?>(null)
 
@@ -222,6 +227,7 @@ constructor(
 
                         val supportedFixedFrameRates = getSupportedFrameRates(camInfo)
                         val supportedImageFormats = getSupportedImageFormats(camInfo)
+                        val hasFlashUnit = camInfo.hasFlashUnit()
 
                         put(
                             lensFacing,
@@ -235,7 +241,8 @@ constructor(
                                     // Ultra HDR now.
                                     Pair(CaptureMode.SINGLE_STREAM, setOf(ImageOutputFormat.JPEG)),
                                     Pair(CaptureMode.MULTI_STREAM, supportedImageFormats)
-                                )
+                                ),
+                                hasFlashUnit = hasFlashUnit
                             )
                         )
                     }
@@ -354,6 +361,39 @@ constructor(
                                 FocusMeteringAction.Builder(it.meteringPoint).build()
                             Log.d(TAG, "Starting focus and metering")
                             camera.cameraControl.startFocusAndMetering(focusMeteringAction)
+                        }
+                    }
+
+                    val channelToClose = videoCaptureControlEvents
+                    videoCaptureControlEvents = Channel()
+                    channelToClose?.close()
+                    launch {
+                        val videoCapture = useCaseGroup.getVideoCapture()
+                        var recording: Recording? = null
+                        videoCaptureControlEvents!!.consumeAsFlow().collect { event ->
+                            when (event) {
+                                is VideoCaptureControlEvent.StartRecordingEvent -> {
+                                    recording = startVideoRecordingInternal(
+                                        camera,
+                                        videoCapture!!,
+                                        event.onVideoRecord
+                                    )
+                                }
+                                VideoCaptureControlEvent.StopRecordingEvent -> {
+                                    recording?.stop()
+                                }
+                                is VideoCaptureControlEvent.MuteEvent -> {
+                                    recording?.mute(event.muted)
+                                }
+                            }
+                        }
+                    }
+
+                    launch {
+                        cameraInfo.torchState.asFlow().collectLatest { torchState ->
+                            _currentCameraState.update { old ->
+                                old.copy(torchEnabled = torchState == TorchState.ON)
+                            }
                         }
                     }
 
@@ -508,9 +548,24 @@ constructor(
     override suspend fun startVideoRecording(
         onVideoRecord: (CameraUseCase.OnVideoRecordEvent) -> Unit
     ) {
-        if (videoCaptureUseCase == null) {
-            throw RuntimeException("Attempted video recording with null videoCapture use case")
+        if (videoCaptureControlEvents == null) {
+            throw RuntimeException("Attempted video recording with null videoCapture control")
         }
+
+        videoCaptureControlEvents!!.send(
+            VideoCaptureControlEvent.StartRecordingEvent(onVideoRecord)
+        )
+    }
+
+    override fun stopVideoRecording() {
+        videoCaptureControlEvents?.trySend(VideoCaptureControlEvent.StopRecordingEvent)
+    }
+
+    private suspend fun startVideoRecordingInternal(
+        camera: Camera,
+        videoCaptureUseCase: VideoCapture<Recorder>,
+        onVideoRecord: (CameraUseCase.OnVideoRecordEvent) -> Unit
+    ): Recording {
         Log.d(TAG, "recordVideo")
         // todo(b/336886716): default setting to enable or disable audio when permission is granted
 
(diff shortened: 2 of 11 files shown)


core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraUseCase.kt
```diff
@@ -560,12 +565,24 @@ constructor(
                 .start(callbackExecutor) { onVideoRecordEvent ->
                     run {
                         Log.d(TAG, onVideoRecordEvent.toString())
+                        val isTorchRequired = camera.cameraInfo.hasFlashUnit() &&
+                            currentSettings.value?.flashMode == FlashMode.ON
                         when (onVideoRecordEvent) {
+                            is VideoRecordEvent.Start -> {
+                                if (isTorchRequired) {
+                                    camera.cameraControl.enableTorch(true)
```
temcguir, line 573: This should probably happen before we start the recording. Otherwise the beginning of the video may have a larger period where the torch isn't on.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraUseCase.kt
```diff
@@ -560,12 +565,24 @@ constructor(
...
+                                if (isTorchRequired) {
+                                    camera.cameraControl.enableTorch(false)
+                                }
                                 when (onVideoRecordEvent.error) {
                                     ERROR_NONE ->
                                         onVideoRecord(
-                                            CameraUseCase.OnVideoRecordEvent.OnVideoRecorded
+                                            CameraUseCase.OnVideoRecordEvent.OnVideoRecorded(
+                                                onVideoRecordEvent.outputResults.outputUri
+                                            )
```
temcguir, line 585: We've been meaning to put this entire function into its own coroutine within the `runCamera` block so we can get rid of the shared instance variables for `videoCaptureUseCase` and `recording`, similar to what we did [here](https://github.com/google/jetpack-camera-app/pull/202/files#diff-8f5d934183d510bc928fe366b2c1bf6924dcbdc8afae80131a05ea9d469195eeR133)

It's not absolutely necessary for this PR, but if you would like it could make this a bit cleaner.

We would use a `Channel<VideoCaptureControlEvent>` where `VideoCaptureControlEvent` is a private sealed interface that has start and stop events. We could either pass the callback from the arguments of this function into the start event.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraUseCase.kt
```diff
@@ -606,6 +623,11 @@ constructor(
     private val _surfaceRequest = MutableStateFlow<SurfaceRequest?>(null)
     override fun getSurfaceRequest(): StateFlow<SurfaceRequest?> = _surfaceRequest.asStateFlow()
 
+    private val _currentCamera = MutableStateFlow<Camera?>(null)
+
+    @VisibleForTesting
+    internal fun getCurrentCamera(): StateFlow<Camera?> = _currentCamera.asStateFlow()
+
```
temcguir, line 630: I would prefer to not add an API just for testing here. If you don't want to refactor the `startVideoRecording` code to use a Channel, I would at a minimum make the `startVideoRecording` code suspend until a camera is available (maybe with a timeout).

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraUseCase.kt
```diff
@@ -560,12 +565,24 @@ constructor(
                 .start(callbackExecutor) { onVideoRecordEvent ->
                     run {
                         Log.d(TAG, onVideoRecordEvent.toString())
+                        val isTorchRequired = camera.cameraInfo.hasFlashUnit() &&
```
temcguir, line 568: What happens if we don't check if there is a flash unit? We like to remove this kind of check so that the app will crash if you try to enable the torch when it's not available. The UI should only allow developers to enable the torch if the camera constraints allow it.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraUseCase.kt
aaa72: I am not sure if I got this correctly. I create this camera flow for two reasons:
1. CameraXCameraUseCaseTest
The reason to be `@VisibleForTesting and internal fun` since the test needs to get TorchState LiveData to verify result.
2.  fun startVideoRecording()
This method needs to get CameraControl to enable/disable torch.

The refactoring you mentioned can resolve 2 but not 1. 
Or do you think I can get the camera directly from ProcessCameraProvider API?

temcguir: For 1, can you just expose the torch state via [this class](https://github.com/google/jetpack-camera-app/blob/fb39aabf1919b704531e919831858859ebd57a84/core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraUseCase.kt#L134)? Something like:
```
data class CameraState(
    val zoomScale: Float = 1f,
    val sessionFirstFrameTimestamp: Long = 0L
    val torchEnabled: Boolean = false
)
```

Then you would just require listening to the torch state `LiveData` in the camera coroutine:
```
launch {
    cameraInfo.torchState.asFlow().collectLatest { torchState ->
        _currentCameraState.update { old ->
            old.copy(torchEnabled = torchState == TorchState.ON)
        }
    }
}
```

You can then listen to the `CameraState` Flow from `CameraUseCase.currentCameraState`. This will also be useful for the UI so that we could show some UI when the torch is enabled.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraUseCase.kt
aaa72: > What happens if we don't check if there is a flash unit?
The returned future will fail immediately since the camera control will also check the flash unit.

Change the code not to check the flash unit, so it will crash if there is no flash unit.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraUseCase.kt
aaa72: Thanks for the sample and done.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraUseCase.kt
aaa72: Update the code as needed. However, I discovered an issue in the existing code:
```
launch {
    focusMeteringEvents.consumeAsFlow().collect {
        ...
    }
}
```
This code has a bug. When you log the associated camera lens facing, you'll notice that after switching the camera and tapping to focus, the focus events may sent to the wrong camera for each tap.
```
Log.d(TAG, "Starting focus and metering " + camera.cameraInfo.lensFacing)
```
The root cause seems to be that the code launches a new coroutine and listener for the channel every time the camera lens is switched (or any setting is changed). This results in multiple coroutines running concurrently, all doing the same thing but with different local variables. While the "tap to focus" issue isn't immediately apparent, it becomes easily noticeable when implementing video controls, especially after switching cameras.

The new commit attempts to address the issue by canceling the previous collection of the video control channel, however, this might not be as elegant as you expected.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraUseCase.kt
aaa72: Relocating it before the `PendingRecording.start()` call. Additionally, wrap the `start()` method within a try-catch block to handle any exceptions that might arise, ensuring the torch is properly closed.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraUseCase.kt
temcguir: That's odd. Switching cameras should cancel the previous camera coroutine (and all child coroutines due to structured concurrency). If that's not happening then we should figure out why. There shouldn't be a need to close the channel and create a new one. That actually defeats the purpose of using a channel for synchronization since it's just a mutable instance variable at that point which could technically be called from different threads.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraUseCase.kt
```diff
@@ -543,54 +598,71 @@ constructor(
...
                         }
                     }
-                }
-        currentSettings.value?.audioMuted?.let { recording?.mute(it) }
-    }
+            currentSettings.value?.audioMuted?.let { recording.mute(it) }
+        } catch (e: Exception) {
+            torchDisableDeferred?.complete(Unit)
+            throw e
+        }
```
temcguir, line 663: nit: just us a `finally` block here.

gradle/libs.versions.toml
```diff
@@ -59,6 +60,7 @@ androidx-datastore = { module = "androidx.datastore:datastore", version.ref = "a
 androidx-espresso-core = { module = "androidx.test.espresso:espresso-core", version.ref = "androidxTestEspresso" }
 androidx-graphics-core = { module = "androidx.graphics:graphics-core", version.ref = "androidxGraphicsCore" }
 androidx-junit = { module = "androidx.test.ext:junit", version.ref = "androidxTestJunit" }
+androidx-lifecycle-livedata = { module = "androidx.lifecycle:lifecycle-livedata-ktx", version.ref = "androidxLifecycle" }
```
Jaehwa-Noh, line 63: I can't found anywhere useing LiveData, is there I am missing point?

core/camera/build.gradle.kts
```diff
@@ -94,10 +94,16 @@ dependencies {
...
+    androidTestImplementation(libs.kotlinx.coroutines.test)
+    androidTestImplementation(libs.rules)
+    androidTestImplementation(libs.truth)
 
     // Futures
     implementation(libs.futures.ktx)
 
+    // LiveData
+    implementation(libs.androidx.lifecycle.livedata)
+
```
Jaehwa-Noh, line 106: I can't found anywhere useing LiveData, is there I am missing point?

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraUseCase.kt
aaa72: Oh, the normal path to disable the torch is when the Finalize event is received. This place is to avoid any exception is thrown when a recording can't be started. Put it in finally block will disable the torch too early.

gradle/libs.versions.toml
aaa72: It is used by `cameraInfo.torchState.asFlow()` in `CameraXCameraUseCase.kt`.

core/camera/build.gradle.kts
aaa72: It is used by cameraInfo.torchState.asFlow() in CameraXCameraUseCase.kt.

Jaehwa-Noh: Oh, I didn't know that `cameraInfo` had been using liveData. Thank you.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraUseCase.kt
```diff
@@ -357,6 +365,41 @@ constructor(
...
+                        var recording: Recording? = null
+                        for (event in videoCaptureControlEvents) {
+                            when (event) {
+                                is VideoCaptureControlEvent.StartRecordingEvent -> {
+                                    if (videoCapture == null) {
+                                        throw RuntimeException(
+                                            "Attempted video recording with null videoCapture"
+                                        )
+                                    }
+                                    recording = startVideoRecordingInternal(
```
temcguir, line 379: Once a recording is started, we should listen to changes in the `flashMode` and `audioMuted` fields from `transientSettings` to determine whether we should turn on/off the flash and mute/unmute the audio. This can be done by launching another coroutine in this `when` block that listens to `transientSettings`. Creating a suspend version of `PendingRecording.startRecording` that is similar to how `ProcessCameraProvider.runWith` works would make this easier. We can discuss more offline if you're interested.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewViewModel.kt
```diff
@@ -533,7 +533,7 @@ class PreviewViewModel @AssistedInject constructor(
                     var audioAmplitude = 0.0
                     var snackbarToShow: SnackbarData? = null
                     when (it) {
-                        CameraUseCase.OnVideoRecordEvent.OnVideoRecorded -> {
+                        is CameraUseCase.OnVideoRecordEvent.OnVideoRecorded -> {
```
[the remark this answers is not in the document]
aaa72: Since the `OnVideoRecorded` has been changed from an "object" to a "class", verification now relies on class type  rather than instance.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CoroutineCameraProvider.kt
```diff
@@ -36,7 +37,7 @@ import kotlinx.coroutines.coroutineScope
 suspend fun <R> ProcessCameraProvider.runWith(
     cameraSelector: CameraSelector,
     useCases: UseCaseGroup,
-    block: suspend (Camera) -> R
+    block: suspend CoroutineScope.(Camera) -> R
```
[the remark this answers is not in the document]
aaa72: This code, originating from the initial commit of pull request #251 by @temcguir , aims to fix a [bug](https://github.com/google/jetpack-camera-app/pull/250#discussion_r1696138031) where coroutines within the camera scope weren't canceled when switching cameras.

The intention, as I understand it, is to have the `block` function call occur within the coroutine scope established by `runWith`, specifically inside `= coroutineScope { ... }`. This would ensure that any launch calls within the `block` are also cancelled when the `runWith` block is cancelled. The actual cancellation is triggered by the below `collectLatest` (refer to the [doc](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/collect-latest.html) for details) when new settings, such as camera switching, are detected, and this cancellation propagates to `runWith` and `block`.

```
currentSettings.
    }.distinctUntilChanged()
    .collectLatest {
         // ...
    }
```

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraUseCase.kt
```diff
@@ -1126,3 +1198,9 @@ private fun Int.toAppImageFormat(): ImageOutputFormat? {
         else -> null
     }
 }
+
+private fun UseCaseGroup.getVideoCapture() = getUseCaseOrNull<VideoCapture<Recorder>>()
+
+private inline fun <reified T : UseCase> UseCaseGroup.getUseCaseOrNull(): T? {
```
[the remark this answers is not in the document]
aaa72: In fact, this is also new to me :) and AI help me to correct the syntax of my intention. 
The key seems to lie in the `inline` function, which likely allows the compiler to determine the precise type being used.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraUseCase.kt
```diff
@@ -543,54 +597,71 @@ constructor(
             )
                 .setContentValues(contentValues)
                 .build()
-
+        val torchDisableDeferred = if (currentSettings.value?.flashMode == FlashMode.ON) {
```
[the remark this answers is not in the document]
aaa72: Thanks for pointing out. `initialTransientSettings` seems will not change when changing flash mode so I guess I need to pass `transientSettings` and get the value from it?

By the way, there is also another line that access `currentSettings`, I think I can modify it as well.
```
currentSettings.value?.audioMuted?.let { recording.mute(it) }
```

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraUseCase.kt
```diff
@@ -357,6 +365,41 @@ constructor(
...
+                                    )
+                                }
+                                VideoCaptureControlEvent.StopRecordingEvent -> {
+                                    recording?.stop()
+                                }
+                                is VideoCaptureControlEvent.MuteEvent -> {
+                                    recording?.mute(event.muted)
+                                }
+                            }
+                        }
```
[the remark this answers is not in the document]
aaa72: Updated on commit#3.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraUseCase.kt
aaa72: Per offline discussion. Thanks for the example and updated on commit#3.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraUseCase.kt
```diff
@@ -913,7 +985,7 @@ constructor(
 
     override suspend fun setAudioMuted(isAudioMuted: Boolean) {
         // toggle mute for current in progress recording
-        recording?.mute(!isAudioMuted)
+        videoCaptureControlEvents.send(VideoCaptureControlEvent.MuteEvent(!isAudioMuted))
```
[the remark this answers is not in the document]
aaa72: Updated on commit#3.

PR #251, opened 2024-07-30
Fix child coroutines in runWith scope and update focus events

temcguir:
* Ensure coroutines launched in runWith use correct parent scope
    
     Ensures coroutines in runWith are launched within the runWith
     coroutineScope so cancellation will be applied properly.

   

* Process tap-to-focus events within camera coroutine
    
     Pushes points directly to the camera coroutine which will
     wait for SurfaceRequests to create the
     SurfaceOrientedMeteringPointFactory before processing them.
    
     Any tap-to-focus events that happen without a SurfaceRequest
     will be ignored.

### DIFF
--- a/core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraEvent.kt
+++ b/core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraEvent.kt
@@ -15,8 +15,6 @@
  */
 package com.google.jetpackcamera.core.camera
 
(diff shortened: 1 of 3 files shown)


aaa72 (approved):
Thanks for the fix and it works!

aaa72 (commented):
Opps, I only reviewed the first commit and selected approve, not knowing that all changes would be approved.

aaa72:
All 3 commits looks good to me.

PR #257, opened 2024-08-15
Display current camera physical Id on preview screen

davidjiagoogle:
https://github.com/user-attachments/assets/f5eac0ae-0bb0-4510-b55d-14c5f6159bf8

I want to rush this in first before integrating it with the advanced settings screen.


feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
```diff
@@ -430,13 +430,24 @@ fun ZoomScaleText(zoomScale: Float) {
     )
     Text(
         modifier = Modifier
+            .padding(horizontal = 10.dp)
             .alpha(contentAlpha.value)
             .testTag(ZOOM_RATIO_TAG),
         text = "%.1fx".format(zoomScale),
         fontSize = 20.sp
```
temcguir, line 437: Since we'll be setting a lot of font sizes for all of this debug text, you can wrap the entire block of all text like:
```
CompositionLocalProvider(
    LocalTextStyle provides LocalTextStyle.current.copy(fontSize = 20.sp)
) {
    if (showZoomLevel) {
        ZoomScaleText(zoomLevel)
    }
    if (previewUiState.isDebugMode) {
        CurrentCameraIdText(physicalCameraId, logicalCameraId)
    }
}
```
Then you will no longer need to set it on each element.

core/camera/src/androidTest/java/com/google/jetpackcamera/core/camera/CameraXCameraUseCaseTest.kt
```diff
@@ -183,7 +190,7 @@ class CameraXCameraUseCaseTest {
         val onRecorded = CompletableDeferred<Unit>()
         val onRecordStatus = CompletableDeferred<Unit>()
         var statusCount = 0
-        startVideoRecording {
+        startVideoRecording(null, false) {
```
temcguir, line 193: Is this change related to this PR?

app/src/main/java/com/google/jetpackcamera/MainActivity.kt
```diff
@@ -160,6 +162,11 @@ class MainActivity : ComponentActivity() {
         }
     }
 
+    private fun isDebugMode(): Boolean {
+        return intent != null && intent.hasExtra(KEY_DEBUG_MODE) &&
+            intent.getBooleanExtra(KEY_DEBUG_MODE, false)
+    }
+
```
temcguir, line 169: nit: No need to check if it has the extra if you're already providing a default. You can just do:
```
private fun isDebugMode(): Boolean =
    intent?.getBooleanExtra(KEY_DEBUG_MODE, false) ?: false
```

or as a property:
```
private val isDebugMode: Boolean
    get() = intent?.getBooleanExtra(KEY_DEBUG_MODE, false) ?: false
```

core/camera/src/androidTest/java/com/google/jetpackcamera/core/camera/CameraXCameraUseCaseTest.kt
davidjiagoogle: No but somehow the two parameters are missing and causes the compilation to fail. Somehow this got merged

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
```diff
@@ -432,8 +431,16 @@ fun ZoomScaleText(zoomScale: Float) {
         modifier = Modifier
             .alpha(contentAlpha.value)
             .testTag(ZOOM_RATIO_TAG),
-        text = "%.1fx".format(zoomScale),
-        fontSize = 20.sp
+        text = "%.1fx".format(zoomScale)
```
Jaehwa-Noh, line 434: Is this `%.1fx.format(zoomScale)` supposed to be replaced with formatting strings?
https://developer.android.com/guide/topics/resources/string-resource#formatting-strings

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
```diff
@@ -430,13 +430,24 @@ fun ZoomScaleText(zoomScale: Float) {
...
         fontSize = 20.sp
     )
 }
 
+@Composable
+fun CurrentCameraIdText(physicalCameraId: String?, logicalCameraId: String?) {
+    Text(
+        modifier = Modifier
+            .testTag(CURRENT_CAMERA_ID_TAG),
+        text = "physical id: $physicalCameraId, logical id: $logicalCameraId",
```
temcguir: To make this easier to parse by UI automator, we should split it up so that the physical and logical IDs have their own tag. Something like this:
```
@Composable
fun CurrentCameraIdText(physicalCameraId: String?, logicalCameraId: String?) {
    Column(horizontalAlignment = Alignment.CenterHorizontally) {
        Row {
            Text(text = "Logical ID: ")
            Text(
                modifier = Modifier.testTag(LOGICAL_CAMERA_ID_TAG),
                text = logicalCameraId ?: "---"
            )
        }
        Row {
            Text(text = "Physical ID: ")
            Text(
                modifier = Modifier.testTag(PHYSICAL_CAMERA_ID_TAG),
                text = physicalCameraId ?: "---"
            )
        }
    }
}
```

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
temcguir: Technically this is correct. It should be a localized string.

Jaehwa-Noh: Aha, thank you.

temcguir (approved):

PR #259, opened 2024-08-22
Update VideoRecordingDeviceTest.kt

davidjiagoogle:


app/src/androidTest/java/com/google/jetpackcamera/VideoRecordingDeviceTest.kt
```diff
@@ -73,7 +73,11 @@ internal class VideoRecordingDeviceTest {
         composeTestRule.waitUntil(timeoutMillis = VIDEO_CAPTURE_TIMEOUT_MILLIS) {
             composeTestRule.onNodeWithTag(VIDEO_CAPTURE_SUCCESS_TAG).isDisplayed()
         }
-        Truth.assertThat(File(DIR_PATH).lastModified() > timeStamp).isTrue()
+        val dir = File(DIR_PATH)
+        Truth.assertWithMessage(
+            "Parent dir: ${dir.parentFile.absolutePath}, " +
+                "last modified: ${dir.lastModified()}"
+        ).that(dir.parentFile.lastModified()).isGreaterThan(timeStamp)
```
temcguir, line 80: Why `dir.parentFile`? If `DIR_PATH` resolves to something like `storage/emulated/0/Movies`, then we would be checking for the modified timestamp of `storage/emulated/0`?

PR #261, opened 2024-08-29
Save Camera Properties to external file when in debug mode

davidjiagoogle:
1. When in debug mode, compile the needed camera properties for all cameras and write into a text file in JSON format
2. Included a fix in setOnCaptureCompleteCallback() where the debugInfo in currentCameraState is being constantly updated for each preview frame even if the camera didn't change


core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraUseCase.kt
```diff
@@ -284,6 +384,15 @@ constructor(
                 .tryApplyImageFormatConstraints()
                 .tryApplyFrameRateConstraints()
                 .tryApplyStabilizationConstraints()
+        if (isDebugMode && Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
+            val cameraProperties = getAllCamerasPropertiesJSONArray().toString()
+            val file = File(
+                Environment.getExternalStoragePublicDirectory(DIRECTORY_DOCUMENTS),
+                "JCACameraProperties"
+            )
+            writeFileExternalStorage(file, cameraProperties)
```
temcguir, line 393: This should be done on a background thread so we're not blocking the main thread with disk IO. You can use `withContext(defaultDispatcher) { ...  }` to run this.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraUseCase.kt
```diff
@@ -197,6 +215,87 @@ constructor(
...
+                    getCameraPropertiesJSONObject(physicalCameraId, cameraManager)
+                )
+            }
+            result.put(logicalCameraData)
+        }
+        return result
+    }
+
+    private fun getCameraPropertiesJSONObject(
+        cameraId: String,
```
temcguir, line 242: Can we move both of these methods to a separate file so we're not adding too much to this file? Maybe as free functions in a file called `DebugCameraInfo.kt`?

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraUseCase.kt
```diff
@@ -197,6 +215,87 @@ constructor(
...
+            file.createNewFile()
+            //second argument of FileOutputStream constructor indicates whether
+            //to append or create new file if one exists
+            outputStream = FileOutputStream(file, true)
+
+            outputStream.write(textToWrite.toByteArray())
+            outputStream.flush()
+            outputStream.close()
+        } catch (e: java.lang.Exception) {
+            e.printStackTrace()
```
temcguir, line 295: Don't use `e.printStackTrace()`. Instead use `Log.e(TAG, <message>, <exception>)`.

However, we should also consider whether we want to catch the exception here or just have it thrown and crash the app. If we're in debug mode, we're expecting the file to be written, so I would lean towards just throwing the exception and allowing it to crash the app.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraUseCase.kt
```diff
@@ -197,6 +215,87 @@ constructor(
...
+        var outputStream: FileOutputStream? = null
+        try {
+            file.createNewFile()
+            //second argument of FileOutputStream constructor indicates whether
+            //to append or create new file if one exists
+            outputStream = FileOutputStream(file, true)
+
+            outputStream.write(textToWrite.toByteArray())
+            outputStream.flush()
+            outputStream.close()
```
temcguir, line 293: `close()` should always happen, regardless of whether an exception was thrown or not. I would recommend using kotlins [`use`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.io/use.html) function here. So instead of the `try` block, it would look something like:
```
FileOutputStream(file).use { outputStream ->
    outputStream.write(textToWrite.toByteArray())
}
```

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraUseCase.kt
```diff
@@ -197,6 +215,87 @@ constructor(
...
+        //This point and below is responsible for the write operation
+        var outputStream: FileOutputStream? = null
+        try {
+            file.createNewFile()
+            //second argument of FileOutputStream constructor indicates whether
+            //to append or create new file if one exists
+            outputStream = FileOutputStream(file, true)
+
+            outputStream.write(textToWrite.toByteArray())
+            outputStream.flush()
```
temcguir, line 292: `flush` doesn't actually do anything for `FileOutputStream`. The implementation is empty. So you don't need to call this.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraUseCase.kt
Jaehwa-Noh: Is that any reason you recommend for using a defaultDispatcher, not an IO related dispatcher?

temcguir: You're right, an IO dispatcher would be better. I only recommended `defaultDispatcher` because we already have it available in this class. We should inject an IO dispatcher to use instead.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraUseCase.kt
```diff
@@ -197,6 +215,87 @@ constructor(
...
+                )
+            }
+            result.put(logicalCameraData)
+        }
+        return result
+    }
+
+    private fun getCameraPropertiesJSONObject(
+        cameraId: String,
+        cameraManager: CameraManager
```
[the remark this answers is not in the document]
davidjiagoogle: I don't think there is a way to get cameraId from CameraInfo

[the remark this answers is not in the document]
temcguir: `Camera2CameraInfo.from(cameraInfo).cameraId`

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraUseCase.kt
```diff
@@ -385,13 +295,18 @@ constructor(
...
                 .tryApplyStabilizationConstraints()
         if (isDebugMode && Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
-            val cameraProperties = getAllCamerasPropertiesJSONArray().toString()
-            val file = File(
-                Environment.getExternalStoragePublicDirectory(DIRECTORY_DOCUMENTS),
-                "JCACameraProperties"
-            )
-            writeFileExternalStorage(file, cameraProperties)
-            Log.d(TAG, "JCACameraProperties written to ${file.path}. \n" + "$cameraProperties")
+            withContext(Dispatchers.IO) {
```
Jaehwa-Noh, line 298: The better way is using injection with Hilt like a defaultDispatcher. If it is hard to deal with in this PR, it seems better file it into issue and makes dedicated PR.

temcguir: Agreed. You can stil use Dispatchers.IO, but it should be injected. See how `provideDefaultDispatcher` works in `CommonModule`. That way we can inject test dispatchers in tests if needed.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/DebugCameraInfoUtil.kt
```diff
@@ -0,0 +1,113 @@
...
+            ?.let {
+                jsonObject.put(
+                    CameraCharacteristics.LENS_INFO_MINIMUM_FOCUS_DISTANCE.name,
+                    it
+                )
+            }
+        cameraCharacteristics.get(CameraCharacteristics.REQUEST_AVAILABLE_CAPABILITIES)
+            ?.let { jsonObject.put(CameraCharacteristics.REQUEST_AVAILABLE_CAPABILITIES.name, it) }
+
+        return jsonObject
```
temcguir, line 93: After trying this out, many of these are types like `FloatArray` which don't have a good `toString()` implementation. You should check the return type of each of these `CameraCharacteristics` and use functions like `FloatArray?.contentToString()` to write them to the file.

davidjiagoogle: good catch!

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraUseCase.kt
```diff
@@ -186,6 +196,18 @@ constructor(
                 .tryApplyFrameRateConstraints()
                 .tryApplyStabilizationConstraints()
                 .tryApplyConcurrentCameraModeConstraints()
+        if (isDebugMode && Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
+            withContext(iODispatcher) {
+                val cameraProperties =
+                    getAllCamerasPropertiesJSONArray(cameraProvider.availableCameraInfos).toString()
+                val file = File(
+                    Environment.getExternalStoragePublicDirectory(DIRECTORY_DOCUMENTS),
+                    "JCACameraProperties"
```
temcguir, line 205: Please use the `.json` extension for this file.

PR #263, opened 2024-09-10
Use camera-compose and restore double-tap to flip camera

temcguir:
- **Move HDR color mode code out of CameraXViewfinder**
- **Replace CameraXViewfinder with camera-compose version**
- **Move double tap logic to same pointerInput as tapToFocus**


feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
```diff
@@ -316,21 +312,94 @@ fun PreviewDisplay(
...
+        )
+        val currentOnRequestWindowColorMode: (Int) -> Unit by rememberUpdatedState(
+            onRequestWindowColorMode
+        )
+
+        LaunchedEffect(Unit) {
+            val colorModeSnapshotFlow =
+                snapshotFlow { Pair(currentSurfaceRequest.dynamicRange, currentImplementationMode) }
+                    .map { (dynamicRange, implMode) ->
+                        val isSourceHdr = dynamicRange.encoding != CXDynamicRange.ENCODING_SDR
```
Jaehwa-Noh, line 378: Is there any possible `isSourceHdr` would be not HDR Encoding, even it is not Encoding SDR, such as Encoding Dolby Vision, Encoding HLG, Encoding Unspecified?

davidjiagoogle (approved):

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
temcguir: Dolby Vision, HLG10, HDR10 and HDR10+ are all considered HDR formats. Technically, the DynamicRange here from CameraX should never be UNSPECIFIED, since it is coming from `SurfaceRequest`. I'll make a note to update the CameraX documentation to better communicate that.

temcguir: The CameraX documentation improvements will be tracked with https://issuetracker.google.com/366536534

Jaehwa-Noh: Thank you for everything.

PR #265, opened 2024-09-12
Fix tests which don't specify explicit URI

temcguir:
When running JCA without an explicit intent mode, the files are
 saved to MediaStore. This installs a watch on the MediaStore to
 delete any files that are created during tests.


app/src/androidTest/java/com/google/jetpackcamera/utils/AppTestUtil.kt
```diff
@@ -15,12 +15,124 @@
...
+                    ) {
+                        // Construct URI for a single file
+                        val outputUri = if (uri.lastPathSegment?.equals("$id") == false) {
+                            uri.buildUpon().appendPath("$id").build()
+                        } else {
+                            uri
+                        }
+                        put(displayName, outputUri)
+                    }
+                    cursor.moveToNext()
```
davidjiagoogle, line 95: just making sure, this cursor doesn't need to be closed right?

temcguir: The cursor is automatically closed by virtue of accessing it via a `use {}` block (on line 69).

app/src/androidTest/java/com/google/jetpackcamera/utils/UiTestUtil.kt
```diff
@@ -21,30 +21,103 @@ import android.content.ComponentName
...
+) = runBlocking {
+    val debugTag = "MediaStoreAutoDelete"
+    val instrumentation = InstrumentationRegistry.getInstrumentation()
+    val insertedMediaStoreEntries = mutableMapOf<String, Uri>()
+    val observeFilesJob = launch(fileObserverContext) {
+        mediaStoreInsertedFlow(
+            mediaUri = mediaUri,
+            instrumentation = instrumentation,
+            filePrefix = filePrefix
+        ).take(expectedNumFiles)
```
davidjiagoogle, line 74: I'm having trouble understanding this line. The observer sends all the files upon a change and they are put into the entries. What does ".take(expectedNumFiles)" do?

temcguir: [Flow.take](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/take.html) will transform the flow into another flow that only contains `expectedNumFiles` elements. This means the flow will complete once the expected number of files has been received, and execution will continue past the `collect` block. When the number of files produced is less than `expectedNumFiles`, the `collect` block will suspend while waiting for the next result.

davidjiagoogle (approved):

PR #268, opened 2024-09-30
Mute Audio Default Setting

Kimblebee:
Persistent setting to start video recordings muted or with audio.

notable changes:
* The permission to record audio is used as a constraint for the Mute Audio setting -- It will be disabled if the permission is not granted

* Same process of adding a new setting...


feature/settings/src/main/res/values/strings.xml
```diff
@@ -113,7 +122,7 @@
     <string name="fps_unsupported"> %1$s is unsupported at %2$d fps</string>
     <string name="stabilization_unsupported">%$1s is unsupported by the current stabilization</string>
     <string name="current_lens_unsupported">%$s is unsupported by the current lens</string>
-    <string name="rear_lens_unsupported">%$s is unsupported by the rear lens</string>
+    <string name="rear_lens_unsupported">%$s is unsupported by the rear lens</string>s
```
Jaehwa-Noh, line 125:
```suggestion
    <string name="rear_lens_unsupported">%$s is unsupported by the rear lens</string>
```
Is this typo?

Kimblebee: it is a typo. I must have accidentally hit the button before committing. thank you for catching it!

temcguir:
In the current code, you can now start a recording muted with `PendingRecording.withAudioEnabled(initialMuted = true)`. Maybe you can add that to this PR?

https://developer.android.com/reference/androidx/camera/video/PendingRecording#withAudioEnabled(kotlin.Boolean)

.idea/androidTestResultsUserPreferences.xml
temcguir, line 1: Remove this file

.idea/inspectionProfiles/Project_Default.xml
temcguir, line 1: Remove this file

.idea/other.xml
temcguir, line 1: Remove this file

gradle/libs.versions.toml
```diff
@@ -21,7 +21,7 @@ protobufPlugin = "0.9.4"
 androidxActivityCompose = "1.8.2"
 androidxAppCompat = "1.6.1"
 androidxBenchmark = "1.2.3"
-androidxCamera = "1.5.0-SNAPSHOT"
+androidxCamera = "1.5.0-alpha02"
```
temcguir, line 24: This should stay as a `SNAPSHOT`. If you need a newer API, then update the build ID in settings.gradle.

feature/settings/src/main/java/com/google/jetpackcamera/settings/ui/SettingsComponents.kt
```diff
@@ -622,6 +623,33 @@ fun StabilizationSetting(
...
+        modifier = modifier,
+        title = stringResource(id = R.string.mute_audio_title),
+        description = if (mutedUiState is MuteAudioUiState.Enabled) {
+            if (mutedUiState.isMuted) {
+                stringResource(R.string.mute_selector_on)
+            } else {
+                stringResource(R.string.mute_selector_off)
+            }
+        } else {
+            TODO("mute toggle currently has no disabled criteria")
```
temcguir, line 642: What about if the audio permission wasn't granted?

feature/settings/src/main/java/com/google/jetpackcamera/settings/SettingsUiState.kt
```diff
@@ -167,8 +172,18 @@ sealed interface StabilizationUiState {
     // Stabilization selection completely disabled. Cannot open dialog.
     data class Disabled(val disabledRationale: DisabledRationale) : StabilizationUiState
 }
+// ////////////////////////////////////////////////////////////
+//
+// Settings that DON'T currently depend on constraints
```
temcguir, line 177: Would the audio permission be considered a constraint here?

feature/settings/src/main/java/com/google/jetpackcamera/settings/ui/SettingsComponents.kt
Kimblebee: For now I'll disable if the permission isnt granted.
But I guess it could display some UI to prompt the user to grant the permission. That can be addressed a later PR though, since there would also need a case to navigate to the app settings page.

feature/settings/src/main/java/com/google/jetpackcamera/settings/SettingsUiState.kt
Kimblebee: 🫣 yes it should be be

feature/settings/src/main/java/com/google/jetpackcamera/settings/SettingsViewModel.kt
```diff
@@ -40,14 +41,15 @@ import kotlinx.coroutines.flow.stateIn
...
-val fpsOptions = setOf(FPS_15, FPS_30, FPS_60)
+private val fpsOptions = setOf(FPS_15, FPS_30, FPS_60)
 
 /**
  * [ViewModel] for [SettingsScreen].
  */
 @HiltViewModel
 class SettingsViewModel @Inject constructor(
     private val settingsRepository: SettingsRepository,
+    private val permissionChecker: PermissionChecker,
```
temcguir, line 52: So I tried to research this a little, and from what I can tell, the permissions are supposed to be fully owned by the UI layer. We should treat them as if they are user interaction events, like pressing a button. With that in mind, we should check permissions in the UI (Activity or Screen level) and send them down to the ViewModel, where the ViewModel can generate the UiState that depends on them.

So the `SettingsViewModel` should have something like:
```
class SettingsViewModel {
    fun setAudioPermissionGranted(granted: Boolean)
}
```

or if you need multiple permissions to be communicated to the ViewModel:

```
class SettingsViewModel {
    fun setGrantedPermissions(permissions: Set<String>)
}
```

This will align with [the guidance](https://developer.android.com/topic/architecture/recommendations#viewmodel) that:
image

whereas currently, this code is storing a `Context` (albeit an _Application_ context) in the ViewModel, which violates this guidance.

feature/settings/src/main/java/com/google/jetpackcamera/settings/SettingsViewModel.kt
```diff
@@ -49,9 +52,9 @@ private val fpsOptions = setOf(FPS_15, FPS_30, FPS_60)
 @HiltViewModel
 class SettingsViewModel @Inject constructor(
     private val settingsRepository: SettingsRepository,
-    private val permissionChecker: PermissionChecker,
     constraintsRepository: ConstraintsRepository
 ) : ViewModel() {
+    private var grantedPermissions = mutableSetOf<String>()
```
temcguir, line 59: You should make this a `MutableStateFlow<Set<String>>` so that it can be combined into the below flow. Otherwise you have a race between this being set and the settings/constraints being updated.

yasith (approved):

PR #269, opened 2024-10-09
Add LLB as a FlashMode, and remove existing LLB toggle related settings

yasith:
LLB Flash Toggle

- Add Low Light Boost (LLB) as an additional Flash option
- This will be combined with #236 later on (NightLight icon outline vs filled when LLB is active)


Kimblebee:
just curious, is there ever a scenario where someone would want LLB and Flash to be on simultaneously? Like in a wide open area where the flash may help give a little extra light but not enough for it to fully illuminate? or taking a photo of someone at a bit of a distance in the dark

yasith:
> just curious, is there ever a scenario where someone would want LLB and Flash to be on simultaneously? Like in a wide open area where the flash may help give a little extra light but not enough for it to fully illuminate? or taking a photo of someone at a bit of a distance in the dark

From my understanding LLB doesn't work well with Flash. 
For all existing implementations of LLB and Night Capture, flash is turned off.

It's not something in the current scope.

data/settings/src/main/java/com/google/jetpackcamera/settings/model/FlashMode.kt
```diff
@@ -18,5 +18,6 @@ package com.google.jetpackcamera.settings.model
 enum class FlashMode {
     OFF,
     ON,
-    AUTO
+    AUTO,
+    LLB
```
Kimblebee, line 22: i think it'd be ok to just use `LOW_LIGHT` or `LOW_LIGHT_BOOST` for readability

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/quicksettings/QuickSettingsEnums.kt
```diff
@@ -96,6 +95,12 @@ enum class CameraFlashMode : QuickSettingsEnum {
         override fun getImageVector() = Icons.Filled.FlashOn
         override fun getTextResId() = R.string.quick_settings_flash_on
         override fun getDescriptionResId() = R.string.quick_settings_flash_on_description
+    },
+    LLB {
```
Kimblebee, line 99: again something like `LOW_LIGHT` would help with readability instead of just `LLB`

data/settings/src/main/java/com/google/jetpackcamera/settings/LocalSettingsRepository.kt
```diff
@@ -105,6 +105,7 @@ class LocalSettingsRepository @Inject constructor(
             FlashMode.AUTO -> FlashModeProto.FLASH_MODE_AUTO
             FlashMode.ON -> FlashModeProto.FLASH_MODE_ON
             FlashMode.OFF -> FlashModeProto.FLASH_MODE_OFF
+            FlashMode.LLB -> FlashModeProto.FLASH_MODE_OFF
```
Kimblebee, line 108: `FlashModeProto.FLASH_MODE_LLB`

feature/settings/src/main/java/com/google/jetpackcamera/settings/ui/SettingsComponents.kt
```diff
@@ -235,6 +236,12 @@ fun FlashModeSetting(
                     enabled = true,
                     onClick = { setFlashMode(FlashMode.OFF) }
                 )
+                SingleChoiceSelector(
```
Kimblebee, line 239: low light boost is not supported on all devices? 

How can we verify that a device supports low light boost? Would need to check for that and add a condition to disable this option.

This PR is mostly refactor and UI, and the constraint doesn't affect other settings. so now we could just file a bug and write a TODO.

Kimblebee (changes requested):
Looks good. 
Noted that we may have to include a constraint for this new option.
Besides that, only small changes to be addressed and it should be good to go :-)

feature/settings/src/main/java/com/google/jetpackcamera/settings/ui/SettingsComponents.kt
yasith: Added a TODO. This will be integrated with #236

Kimblebee:
Looks good, but you need to do one more thing. 

Go into `LocalSettingsRepository.kt`

add this line: `FlashModeProto.FLASH_MODE_LOW_LIGHT_BOOST -> FlashMode.LOW_LIGHT_BOOST`

into
 ```
flashMode = when (it.flashModeStatus) {
                    FlashModeProto.FLASH_MODE_AUTO -> FlashMode.AUTO
                    FlashModeProto.FLASH_MODE_ON -> FlashMode.ON
                    FlashModeProto.FLASH_MODE_OFF -> FlashMode.OFF
                    FlashModeProto.FLASH_MODE_LOW_LIGHT_BOOST -> FlashMode.LOW_LIGHT_BOOST
                    else -> FlashMode.OFF
                }
```

The default setting will not set to LLB unless you include that line

File: feature/preview/src/main/res/values/strings.xml [hunk 5]
```diff
@@ -69,9 +69,11 @@
     <string name="quick_settings_flash_off">OFF</string>
     <string name="quick_settings_flash_auto">AUTO</string>
     <string name="quick_settings_flash_on">ON</string>
+    <string name="quick_settings_flash_llb">LLB</string>
```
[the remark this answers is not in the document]
yasith: This string has to be short, since it's part of a quick settings tile

Kimblebee (approved):

PR #273, opened 2024-10-15
Use FileDescriptorOutput instead of FileOutputs for external uri video recording

davidjiagoogle:
Refactor out getPendingRecording() from runVideoRecording(). During obtaining PendingRecording, check that Android Build Version supports FileDescriptor. If so, use FileDescriptorOutputOptions for video recording. During opening file descriptor, if an exception is thrown, translate it to OnVideoRecordError. If Android Build Version doesn't support openFileDescriptor, check if the uri scheme is file, if so, use FileOutputOptions for recording. If not, fire a "Uri scheme not supported." OnVideoRecordError

Verified that this supports both content and file uri passed into JCA on version 26 and above


core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraSession.kt
```diff
@@ -582,48 +623,57 @@ private suspend fun runVideoRecording(
...
+        shouldUseUri,
+        onVideoRecord
+    )
 
-        transientSettings.filterNotNull()
-            .onCompletion {
-                // Could do some fancier tracking of whether the torch was enabled before
-                // calling this.
-                camera.cameraControl.enableTorch(false)
+    if (pendingRecording != null) {
```
temcguir, line 636: Instead of creating a local `pendingRecording` variable, just use:
```
getPendingRecording(
...
)?.let {
  ...
}
```

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraSession.kt
```diff
@@ -470,34 +472,50 @@ private fun setFlashModeInternal(
...
-private suspend fun startVideoRecordingInternal(
-    initialMuted: Boolean,
+private fun getPendingRecording(
+    context: Context,
     videoCaptureUseCase: VideoCapture<Recorder>,
     maxDurationMillis: Long,
     captureTypeSuffix: String,
-    context: Context,
     videoCaptureUri: Uri?,
     shouldUseUri: Boolean,
```
temcguir, line 481: We should really merge these into a single sealed interface to clean up the conditional code below. If we put them into something like:
```
sealed interface VideoOutput {
    data class MediaStore(val captureTypeSuffix: String) : VideoOutput
    data class File(val file: File) : VideoOutput
    data class ContentUri(val uri: Uri) : VideoOutput
}
```

Then we could get rid of the messy nullable Uri / boolean combination to specify whether we should use the Uri. It would also allow us to validate the URI well ahead of trying to start the recording.

davidjiagoogle: Let's leave this for another PR that specifically improves uri validation as we talked about

temcguir (approved):

PR #274, opened 2024-10-16
implement pause/resume

Kimblebee:
wip


core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraSession.kt
```diff
@@ -615,13 +615,11 @@ private suspend fun startVideoRecordingInternal(
...
                         )
                     }
-
                     ERROR_DURATION_LIMIT_REACHED -> {
                         currentCameraState.update { old ->
                             old.copy(
                                 videoRecordingState = VideoRecordingState.Inactive(
-                                    // cleanly display the max duration
-                                    finalElapsedTimeNanos = maxDurationMillis * 1_000_000
+                                    finalElapsedTimeNanos = maxDurationMillis * 1_000_000_000
```
temcguir, line 622: Wouldn't milliseconds * 1,000,000,000 be picoseconds?

If we want to be more explicit about this conversion, we can do `maxDurationMillis.milliseconds.inWholeNanoseconds`.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraUseCase.kt
```diff
@@ -428,6 +428,14 @@ constructor(
...
+    override suspend fun pauseVideoRecording() {
+        videoCaptureControlEvents.trySendBlocking(VideoCaptureControlEvent.PauseRecordingEvent)
+    }
+
+    override suspend fun resumeVideoRecording() {
+        videoCaptureControlEvents.trySendBlocking(VideoCaptureControlEvent.ResumeRecordingEvent)
+    }
+
     override fun stopVideoRecording() {
         videoCaptureControlEvents.trySendBlocking(VideoCaptureControlEvent.StopRecordingEvent)
```
temcguir, line 440: Just realized we're using `trySendBlocking` here. `stopVideoRecording` should be converted to a `suspend` fun and `trySendBlocking` should be `send`.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraUseCase.kt
```diff
@@ -428,6 +428,14 @@ constructor(
         )
     }
 
+    override suspend fun pauseVideoRecording() {
+        videoCaptureControlEvents.trySendBlocking(VideoCaptureControlEvent.PauseRecordingEvent)
+    }
+
+    override suspend fun resumeVideoRecording() {
+        videoCaptureControlEvents.trySendBlocking(VideoCaptureControlEvent.ResumeRecordingEvent)
```
temcguir, line 436: Use `send` instead of `trySendBlocking`.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraUseCase.kt
```diff
@@ -428,6 +428,14 @@ constructor(
         )
     }
 
+    override suspend fun pauseVideoRecording() {
+        videoCaptureControlEvents.trySendBlocking(VideoCaptureControlEvent.PauseRecordingEvent)
```
temcguir, line 432: Use `send` instead of `trySendBlocking`.

temcguir (commented):
Instead of the blue capture button when the recording is paused, can we just make it red with `37%` alpha. This sort of lines up with a ["Disabled" state]( https://m3.material.io/foundations/interaction/states/applying-states#4aff9c51-d20f-4580-a510-862d2e25e931), though the thing that is disabled is the recording, not the element itself (which is why we keep the red color).

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraSession.kt
Kimblebee: oops, that looks like an accidental change. 
Thanks for the conversion snippet :)

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
```diff
@@ -17,6 +17,7 @@ package com.google.jetpackcamera.feature.preview.ui
 
 import android.content.pm.ActivityInfo
 import android.content.res.Configuration
+import android.graphics.Color.alpha
```
temcguir, line 20: nit: I think this import is unused.

temcguir (approved):

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
```diff
@@ -138,10 +140,50 @@ fun ElapsedTimeText(
     }
 }
 
+@Composable
+fun PauseResumeToggleButton(
+    modifier: Modifier = Modifier,
+    onSetPause: (Boolean) -> Unit,
+    size: Int = 75,
```
Jaehwa-Noh, line 147: I think size is only used as Float type, what about change type of size to Float?

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
```diff
@@ -583,19 +626,29 @@ fun CaptureButton(
...
-            drawCircle(
-                color =
-                when (videoRecordingState) {
-                    is VideoRecordingState.Inactive -> {
-                        if (isPressedDown) currentColor else Color.Transparent
-                    }
+        Canvas(
+            modifier = Modifier
+                .size(110.dp)
+                .apply {
```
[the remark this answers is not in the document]
Kimblebee: this was supposed to apply a test tag for a test. I had some issues with it so I moved it into another branch. thanks for catching

PR #275, opened 2024-10-16
New external intent capture mode for multiple image captures

davidjiagoogle:
Adds a new PreviewMode named ExternalMultipleImageCaptureMode. This PreviewMode is launched through external intent action MediaStore.INTENT_ACTION_STILL_IMAGE_CAMERA. 

A set of uris can be supplied as a StringArrayListExtra at key MediaStore.EXTRA_OUTPUT, after which the user can take picture at number of times equal to the number of uris supplied in the extra, then the app will finish with RESULT_OK. 

If the format of any of the uris supplied is illegal, that particular capture would fail. The capture failure of one uri does not impact those of other uris or the activity result.

If no extra is supplied at key MediaStore.EXTRA_OUTPUT, the app enters IMAGE_ONLY mode and user can take as many pictures without specifying file locations. If user took one or more picture, exiting JCA will result in RESULT_OK, otherwise exiting results in RESULT_CANCELLED.


feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewMode.kt
```diff
@@ -44,4 +44,13 @@ sealed interface PreviewMode {
         val videoCaptureUri: Uri?,
         val onVideoCapture: (PreviewViewModel.VideoCaptureEvent) -> Unit
     ) : PreviewMode
+
+    /**
+     * Under this mode, the app is launched by an external intent to capture multiple images.
+     */
+    data class ExternalMultipleImageCaptureMode(
+        val imageCaptureUris: List<Uri>?,
+        var currentUriIndex: Int,
```
temcguir, line 53: Let's avoid the "Mode" containing mutable state. Can you just keep track of the `currentUriIndex` in an instance variable inside of `PreviewViewModel`?

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewViewModel.kt
```diff
@@ -550,14 +562,34 @@ class PreviewViewModel @AssistedInject constructor(
                         }
                     }, contentResolver, imageCaptureUri, ignoreUri).savedUri
                 },
-                onSuccess = { savedUri -> onImageCapture(ImageCaptureEvent.ImageSaved(savedUri)) },
+                onSuccess = { savedUri ->
+                    onImageCapture(ImageCaptureEvent.ImageSaved(savedUri), uriIndex)
+                    incrementExternalMultipleImageCaptureModeUriIndexIfNeeded()
```
temcguir, line 567: Should this be incremented inline with `captureImageInternal`? If you wait for the success/failure callback, you could potentially create multiple image capture requests for the same URI if you press the capture button fast enough.

app/src/androidTest/java/com/google/jetpackcamera/ImageCaptureDeviceTest.kt
```diff
@@ -152,6 +154,147 @@ internal class ImageCaptureDeviceTest {
...
+                    composeTestRule.onNodeWithTag(CAPTURE_BUTTON).isDisplayed()
+                }
+
+                composeTestRule.onNodeWithTag(CAPTURE_BUTTON)
+                    .assertExists()
+                    .performClick()
+                composeTestRule.waitUntil(timeoutMillis = IMAGE_CAPTURE_TIMEOUT_MILLIS) {
+                    composeTestRule.onNodeWithTag(IMAGE_CAPTURE_SUCCESS_TAG).isDisplayed()
+                }
+                composeTestRule.waitUntil(timeoutMillis = APP_START_TIMEOUT_MILLIS) {
```
temcguir, line 183: You should create a different timeout constant and not use `APP_START_TIMEOUT_MILLIS` for waiting for the toast to go away.

app/src/androidTest/java/com/google/jetpackcamera/ImageCaptureDeviceTest.kt
```diff
@@ -152,6 +154,147 @@ internal class ImageCaptureDeviceTest {
...
+                )
+            ) {
+                // Wait for the capture button to be displayed
+                composeTestRule.waitUntil(timeoutMillis = APP_START_TIMEOUT_MILLIS) {
+                    composeTestRule.onNodeWithTag(CAPTURE_BUTTON).isDisplayed()
+                }
+
+                composeTestRule.onNodeWithTag(CAPTURE_BUTTON)
+                    .assertExists()
+                    .performClick()
```
temcguir, line 179: Instead of explicitly calling performClick() and waiting for toasts 3 times, can you just do this in a `for` loop?

app/src/androidTest/java/com/google/jetpackcamera/ImageCaptureDeviceTest.kt
```diff
@@ -152,6 +154,147 @@ internal class ImageCaptureDeviceTest {
...
+                composeTestRule.waitUntil(timeoutMillis = APP_START_TIMEOUT_MILLIS) {
+                    composeTestRule.onNodeWithTag(CAPTURE_BUTTON).isDisplayed()
+                }
+                uiDevice.pressBack()
+            }
+        Truth.assertThat(result.resultCode).isEqualTo(Activity.RESULT_CANCELED)
+    }
+
+    @Test
+    fun multiple_image_capture_external_illegal_uri() {
```
temcguir, line 262: It's usually good if the test name states the expected result. Names should follow:
`<function/behavior under test>_<underscore-separated conditions>_<expected behavior>`

So something like:
`multipleImageCaptureExternal_withIllegalUri_returnsResultOk`

This also applies to the other tests in this file.

app/src/androidTest/java/com/google/jetpackcamera/ImageCaptureDeviceTest.kt
```diff
@@ -152,6 +154,147 @@ internal class ImageCaptureDeviceTest {
...
+
+                composeTestRule.onNodeWithTag(CAPTURE_BUTTON)
+                    .assertExists()
+                    .performClick()
+                composeTestRule.waitUntil(timeoutMillis = IMAGE_CAPTURE_TIMEOUT_MILLIS) {
+                    composeTestRule.onNodeWithTag(IMAGE_CAPTURE_SUCCESS_TAG).isDisplayed()
+                }
+                composeTestRule.waitUntil(timeoutMillis = APP_START_TIMEOUT_MILLIS) {
+                    composeTestRule.onNodeWithTag(IMAGE_CAPTURE_SUCCESS_TAG).isNotDisplayed()
+                }
```
temcguir, line 238: As with above, this can be collapsed into a `for` loop.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewMode.kt
davidjiagoogle: It'll have to go into PreviewUiState so that UI component get use it to get the right uri

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewViewModel.kt
```diff
@@ -540,6 +543,28 @@ class PreviewViewModel @AssistedInject constructor(
...
+                        } else {
+                            (
+                                (previewUiState.value as PreviewUiState.Ready).previewMode
+                                    as PreviewMode.ExternalMultipleImageCaptureMode
+                                ).imageCaptureUris!![externalUriIndex]
+                        }
+                    Pair(externalUriIndex, uri)
+                } else {
+                    Pair(-1, imageCaptureUri)
+                }
```
temcguir, line 567: This can be rewritten more concisely as:
```
                ((previewUiState.value as? PreviewUiState.Ready)?.previewMode as?
                        PreviewMode.ExternalMultipleImageCaptureMode)?.let {
                    val uri = if (ignoreUri || it.imageCaptureUris.isNullOrEmpty()) {
                        null
                    } else {
                        it.imageCaptureUris[externalUriIndex]
                    }
                    Pair(externalUriIndex, uri)
                } ?: Pair(-1, imageCaptureUri)
```

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/CameraControlsOverlay.kt
```diff
@@ -344,16 +344,32 @@ private fun CaptureButton(
...
+                    is PreviewMode.ExternalMultipleImageCaptureMode -> {
+                        val ignoreUri = previewUiState.previewMode.imageCaptureUris.isNullOrEmpty()
+                        onCaptureImageWithUri(
+                            context.contentResolver,
+                            null,
+                            if (previewUiState.previewMode.imageCaptureUris.isNullOrEmpty()) {
+                                true
+                            } else {
+                                ignoreUri
+                            },
```
temcguir, line 372: Wouldn't this be equivalent to:
`previewUiState.previewMode.imageCaptureUris.isNullOrEmpty() || ignoreUri`?

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewViewModel.kt
```diff
@@ -548,13 +573,29 @@ class PreviewViewModel @AssistedInject constructor(
...
+    private fun incrementExternalMultipleImageCaptureModeUriIndexIfNeeded() {
+        if (previewUiState.value is PreviewUiState.Ready &&
+            (previewUiState.value as PreviewUiState.Ready).previewMode
+                is PreviewMode.ExternalMultipleImageCaptureMode &&
+            !(
+                (previewUiState.value as PreviewUiState.Ready).previewMode
+                    as PreviewMode.ExternalMultipleImageCaptureMode
+                ).imageCaptureUris.isNullOrEmpty()
+        ) {
+            externalUriIndex++
```
temcguir, line 598: nit: Can make this more concise
```
private fun incrementExternalMultipleImageCaptureModeUriIndexIfNeeded() {
    ((previewUiState.value as? PreviewUiState.Ready)
        ?.previewMode as? PreviewMode.ExternalMultipleImageCaptureMode)?.let {
        if (!it.imageCaptureUris.isNullOrEmpty()) {
            externalUriIndex++
        }
    }
}
```

Or you could just do something like initialize `externalUriIndex` to `-1` if the mode isn't `ExternalMultipleImageCaptureMode`. Then you can do
```
if (externalUriIndex >= 0) {
    externalUriIndex++
}
```

temcguir (commented):
It looks like `multipleImageCaptureExternal_returnsResultOk` is failing on the API 28 emulator. You should check to make sure this isn't a real failure.

temcguir (approved):
Approved for when test is passing.

PR #280, opened 2024-11-02
Content description set null

### LINKED ISSUE #279: ContentDescription should be null when no need a ContentDescription.
https://github.com/google/jetpack-camera-app/blob/21fdaf120d14abfee3e884f6ec90f120c5ddea14/feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt#L452-L461

I know `else` condition not meet in current condition, but contentDescription should be null when it is no need.
And then stringResource(id = 0) is not null, just a dummy string.

Jaehwa-Noh:
Fix #279 

Set contentDescription null when it no need.


temcguir (approved):

temcguir:
Hi @Jaehwa-Noh,

Thanks for the contribution. If this is OK to merge, I can merge for you. Let me know.

Jaehwa-Noh:
@temcguir 
I'm totally okay to merge it anytime. It's up to you.
Thank you.

PR #282, opened 2024-11-08
Debug Overlay

davidjiagoogle:
Current has two options: set zoom ratio and show camera properties

https://github.com/user-attachments/assets/eb8303e7-3fbb-4da7-b535-4ffe2a81523e


feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewScreen.kt
```diff
@@ -262,6 +266,13 @@ private fun ContentScreen(
                 onStopVideoRecording = onStopVideoRecording,
                 zoomLevelDisplayState = remember { ZoomLevelDisplayState(isDebugMode) }
             )
+
+            DebugOverlayComponent(
+                toggleIsOpen = onToggleDebugOverlay,
+                previewUiState = previewUiState,
```
Kimblebee, line 272: could we not just directly pass the `debugUiState`?

Kimblebee (approved):
looks good, one comment

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewScreen.kt
davidjiagoogle: at l:199, previewUiState.zoomRatio is needed as the denominator to calculate the new absolute zoomRatio, so we have to pass in previewUiState

PR #286, opened 2024-11-15
Update AGP and remove Preview SDK

temcguir:
The VanillaIceCream preview SDK is now stable, so we can update to useing `targetSdk` 35.

This removes the `targetSdkPreview` and `preview` flavor, as there is no longer a preview SDK to use.


.gitignore
```diff
@@ -1,19 +1,18 @@
...
 .gradle
 /local.properties
-/.idea/caches
-/.idea/libraries
-/.idea/modules.xml
-/.idea/workspace.xml
-/.idea/navEditor.xml
-/.idea/assetWizardSettings.xml
+**/.idea/**
+!**/.idea/codeStyleSettings.xml
```
Kimblebee, line 5: what does the "!" do in gitignore?

core/common/build.gradle.kts
```diff
@@ -53,11 +52,6 @@ android {
             dimension = "flavor"
             isDefault = true
         }
-
-        create("preview") {
```
Kimblebee, line 57: what were these `preview` configs for?

Kimblebee (approved):

.gitignore
temcguir: It's supposed to mean "don't ignore". It's paired with the wildcard ignore on the line above. Essentially instead of having to add each file from `.idea/` that we want to ignore, we can say "ignore everything in `.idea/` *except* for these files.

core/common/build.gradle.kts
temcguir: They were for creating a version of the app that depended on the pre-release Android V SDK (a.k.a. the "Preview" SDK). Since Android V is now released as Android 15, we no longer need it.

In the future, if we want to add a preview flavor again, we can. I didn't want to leave it in, however, since even AGP is complaining that a lot of the preview features are deprecated. So there may be a new way to do this when the next SDK preview is ready.

PR #287, opened 2024-11-18
Update outdated dependencies

temcguir:
This updates all dependencies to the latest. After upgrading there were a few issues that needed to be fixed:
- Compose update caused visuals of setting screen popups to change. Modified code to restore visuals.
- Updating CameraX required changing a package name and removing some annotations
- navigation-compose caused the PreviewScreen to glitch out when coming from PermissionsScreen. Updated usage of navigation-compose to fix the issue.


temcguir:
Updating to the latest version of Compose has caused the following change:
**Old**:
image


**New**:
image

temcguir:
The color change was added in `androidx.compose.material3:material3:1.3.0-alpha02` where ListItems started to use the `surface` color for their container color. I've added an override to the `BasicPopupSetting` content to make `surface` transparent.

davidjiagoogle (approved):

PR #290, opened 2024-11-27
Flip camera while recording in-progress 

Kimblebee:
✏️ Big changes:
* `videoCapture` use case is now a parameter of createUseCaseGroup. 
* `CameraInfo` is moved to `TransientSettings` instead of `PersistentSettings`
   * Single Camera session no longer restarts when switching lens
* move torch management outside of `runVideoRecording`
✅  What works:
* if flash is set to on, torch will be  enabled whenever the rear camera is active during a recording
* Can zoom between either lens
* can tap to focus on either lens
* torch stops when switching lenses and when recording stops
* Pausing works fine as long as you don't flip the camera while paused.

❌ Current Issues:
* flipping a paused recording freezes the paused frame for a duration until resumed.
   * Frozen frame is appended to the video for the length of the realtime pause duration
   * Flaky Issues with rotation changes when unpaused
   * Audio is out of sync 
* Flipping recording while camera is stabilized will cause the recording to freeze on the last frame before switching lens directions.
* Zoom ratio is not persisted between flips.
* Auto Torch for video recording is effectively off


core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraSession.kt
```diff
@@ -99,63 +100,124 @@ internal suspend fun runSingleCameraSession(
...
+}
 
-        applyDeviceRotation(initialTransientSettings.deviceRotation, useCaseGroup)
-        processTransientSettingEvents(
-            camera,
-            useCaseGroup,
-            initialTransientSettings,
-            transientSettings
-        )
+context(CameraSessionContext)
```
temcguir, line 206: Is this context needed?

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraSession.kt
```diff
@@ -182,7 +244,14 @@ internal suspend fun processTransientSettingEvents(
                 }
             }
         }
+        // todo(kc): Enable torch toggle while recording in progress
+        if (currentCameraState.value.videoRecordingState is VideoRecordingState.Active &&
```
temcguir, line 248: This will only be applied when other `transientSettings` change. You will want to `combine` the `currentCameraState` with teh `transientSettings` flow so you get events whjen the `videoRecordingState` changes.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraSession.kt
```diff
@@ -99,63 +100,124 @@ internal suspend fun runSingleCameraSession(
...
-            camera,
-            useCaseGroup,
-            initialTransientSettings,
-            transientSettings
-        )
+context(CameraSessionContext)
+/**
+ * enables torch for the provided camera based on the current flash settings
+ */
+suspend fun toggleTorch(camera: Camera, newTorchOn: Boolean) {
```
temcguir, line 210: This isn't really toggling the torch as much as it's setting the torch. Maybe call it `setTorch` instead.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraSession.kt
```diff
@@ -182,7 +244,14 @@ internal suspend fun processTransientSettingEvents(
                 }
             }
         }
+        // todo(kc): Enable torch toggle while recording in progress
+        if (currentCameraState.value.videoRecordingState is VideoRecordingState.Active &&
+            newTransientSettings.flashMode == FlashMode.ON
```
temcguir, line 249: You'll want to adapt the code from #283 to keep track of when the torch is on locally. It ensures the torch is toggled properly when the device supports it.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraSession.kt
temcguir, line 248: nit: The `VideoRecordingState` only switches to `Active` once a `VideoRecordEvent.Start` event is received. This means the torch won't turn on until _after_ the recording is in progress. We may get a few frames of the torch turning on in that case. We could probably turn on the torch sooner than that, but you won't be able to go off the `VideoRecordingState`, unless you do something like add a new "Starting" state.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraSessionSettings.kt
```diff
@@ -58,6 +58,7 @@ internal sealed interface PerpetualSessionSettings {
  * The use cases typically will not need to be re-bound.
  */
 internal data class TransientSessionSettings(
+    val cameraInfo: CameraInfo,
```
temcguir, line 61: As discussed offline, the `TransientSessionSettings` are supposed to include settings that will not require the camera to be re-bound. You should find some other way to communicate the camera that should be bound outside of using `TransientSessionSettings`.

Kimblebee: sure. I saw "typically" and thought it would still be ok. I can move the cameraselector into a flow in `CameraXCameraUseCase`

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraUseCase.kt
```diff
@@ -118,6 +118,8 @@ constructor(
 
     private val _surfaceRequest = MutableStateFlow<SurfaceRequest?>(null)
 
+    private val cameraSelector = MutableStateFlow<CameraSelector?>(null)
```
temcguir, line 121: This can be a local variable inside the coroutine. Generally we want to try to avoid varaibles with scope longer than the coroutine if possible.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraUseCase.kt
```diff
@@ -260,35 +262,37 @@ constructor(
...
+                                LensFacing.BACK -> CameraSelector.DEFAULT_BACK_CAMERA
+                            }
+                        }
+
+                        transientSettings.value = TransientSessionSettings(
+                            isAudioMuted = currentCameraSettings.audioMuted,
+                            deviceRotation = currentCameraSettings.deviceRotation,
+                            flashMode = currentCameraSettings.flashMode,
+                            zoomScale = currentCameraSettings.zoomScale
+                        )
```
temcguir, line 279: Moving this here seems to have made it so we can no longer zoom in concurrent camera mode. I assume it might also do the same for muting audio and flash.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraUseCase.kt
Kimblebee: i'd still need to be aware of changes to the lensfacing setting within the CameraSession. My first thought was to place the `CameraAppSettings.cameraLensFacing` in the Transient settings, but I recall you advising against including settings that would rebind the camera. What do you think would be a better location to place it?

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraSession.kt
```diff
@@ -720,34 +771,17 @@ private suspend fun runVideoRecording(
...
-                    }
                     .collectLatest { newTransientSettings ->
                         if (currentSettings.isAudioMuted != newTransientSettings.isAudioMuted) {
                             recording.mute(newTransientSettings.isAudioMuted)
                         }
                         if (currentSettings.isFlashModeOn() !=
                             newTransientSettings.isFlashModeOn()
                         ) {
-                            setTorchOn(newTransientSettings.isFlashModeOn())
+                            currentSettings = newTransientSettings
```
temcguir, line 783: Why is this only updated when the flash mode changes?

temcguir (approved):
Is the plan to add the toggle to switch camera in a future PR?

Kimblebee:
> Is the plan to add the toggle to switch camera in a future PR?

Yeah, I thought the current UI on the bottom screen would be a bit too overloaded with the addition of a `flip-camera` element. Unlike pause or mute, user is still able to flip the camera without the use of an extra UI element, so I didn't think it was necessary for now.
I think it should come in a UI refactor once a video can continue recording without holding the capture button.

PR #293, opened 2024-12-10
Zoom Refactor

Kimblebee:
Refactors current zoom implementation to improve code readability, flexibility, and accuracy of the displayed state.

* Rename values using zoomscale/zoomratio to be more readable
* supports setting zoom ratio directly, via a scaling factor, or an increment/decrement value   
   *   _foundation included for modifying secondary (concurrent) zoom value (not yet implemented)_
* zoom ratio persists when switching between lenses
* the zoom display on the preview screen is taken directly from the camera's `zoomState`
* _any changes made to zoom, lensfacing, and mute during a recording will reset to their pre-recording values_


core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraSession.kt
```diff
@@ -314,25 +364,14 @@ internal suspend fun processTransientSettingEvents(
...
-            val finalScale =
-                (zoomScale.value * zoomScaleRelative).coerceIn(
-                    zoomState.minZoomRatio,
-                    zoomState.maxZoomRatio
-                )
-            camera.cameraControl.setZoomRatio(finalScale)
-            zoomScale.update { finalScale }
-            currentCameraState.update { old ->
-                old.copy(zoomScale = finalScale)
+        if (prevTransientSettings.primaryLensFacing == newTransientSettings.primaryLensFacing) {
```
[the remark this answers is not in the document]
Kimblebee: `prevTransientSettings` is first initialized with the current `TransientSessionSettings` available whenever the camera starts up or switches lenses at [line 293](https://github.com/google/jetpack-camera-app/blob/cdfd4b23b6fe5df9ade7a29f267eedb9865d9d6d/core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraSession.kt#L293).

we can't track `prevTransientSettings` with [combine](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/combine.html) because it isn't a flow, and it doesn't need to be..


Think of it as a snapshot of the **_previous_** value of `TransientSessionSettings` that was processed in `processTransientSettingEvents`. 

We can then use the value of `prevTransientSettings` and `newTransientSettings` to make sure  we only process the values that are different.

After processing everything, we reassign `prevTransientSettings` to the new transient settings and wait for the next update.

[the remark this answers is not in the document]
Jaehwa-Noh: Yes, I got it that `prevTransientSettings` would work as though a snapshot.
While I organize my though, I get new question that is one of the combine arguments `transientSettings.filterNotNull()` or `currentCameraState.asStateFlow()` swtiching so fast,
line 374 would be canceled because it is in the collectLatest.
So that there are likely `prevTransientSettings` isn't a latest value?

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/CameraControlsOverlay.kt
```diff
@@ -294,7 +294,11 @@ private fun ControlsBottom(
         ) {
             Column(horizontalAlignment = Alignment.CenterHorizontally) {
                 if (showZoomLevel) {
-                    ZoomScaleText(zoomLevel)
+                    zoomRatios[previewUiState.currentCameraSettings.cameraLensFacing]?.let {
```
temcguir, line 297: We shouldn't introduce new usages of `currentCameraSettings` into the UI. The zoom ratio to show should be determined at the ViewModel level. You shouldn't need to pipe through a Map<LensFacing, Float>. If you want to show multiple zoom ratios in the UI, that kind of state should be encapsulated by something like a `ZoomRatioUiState`.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/ConcurrentCameraSession.kt
```diff
@@ -122,6 +124,32 @@ internal suspend fun runConcurrentCameraSession(
             }
         }
 
+        // update cameraState to mirror the current zoomState
+        launch {
+            primaryCamera.cameraInfo.zoomState.asFlow().filterNotNull().distinctUntilChanged()
+                .collectLatest { zoomState ->
+                    // TODO(): buggy race condition between our setZoomRatio call updating the ZoomState and camera startup
```
temcguir, line 131: Add a link to the android issue tracker with the bug for this condition inside the parenthesist of `TODO()`.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraUseCase.kt
```diff
@@ -641,6 +665,43 @@ constructor(
...
+        }
+        return systemConstraints.perLensConstraints[lensFacing]?.let { constraints ->
+            val newZoomRatio = constraints.supportedZoomRange?.let { zoomRatioRange ->
+                when (val change = newZoomState.changeType) {
+                    is ZoomChange.Absolute -> change.value
+                    is ZoomChange.Scale -> (
+                        this.defaultZoomRatios[lensFacing]
+                            ?: 1.0f
+                        ) *
+                        change.value
```
temcguir, line 689: You can use `defaultZoomRatios.getOrDefault(lensFacing, change.value)`

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraSession.kt
```diff
@@ -252,13 +308,7 @@ internal suspend fun processTransientSettingEvents(
         val newTransientSettings = it.first
         val cameraState = it.second
```
temcguir, line 309: Because of zoom updates, the cameraState is going to be changing pretty rapidly. It's probably a good idea to transform the CameraState into whatever we need from it (such as videoRecordingState), so we don't need to run the `collectLatest` block every time the zoom level changes.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraSession.kt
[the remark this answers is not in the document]
temcguir: This is a good point. It's possible that the `collectLatest` block could complete before reaching the `prevTransientSettings` assignment line. One way around this would be to make sure we aren't doing any suspending operations inside `collectLatest`, or to change to a `collect` block.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraUseCase.kt
```diff
@@ -554,9 +571,16 @@ constructor(
...
-        currentSettings.update { old ->
-            old?.copy(zoomScale = scale)
+    override fun changeZoom(newZoomState: CameraZoomState) {
+        when (newZoomState) {
+            is CameraZoomState.Ratio -> {
+                currentSettings.update { old ->
+                    old?.tryApplyNewZoomRatio(newZoomState) ?: old
+                }
+            }
+            is CameraZoomState.Linear -> {
```
temcguir, line 581: Since we don't have a use for linear zoom, can we remove it?

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraSession.kt
```diff
@@ -150,80 +150,132 @@ internal suspend fun runSingleCameraSession(
...
+                            }
+                        }
+                        .collectLatest { zoomState ->
+                            // TODO(): remove checks after buggy zoomState is fixed
+                            if (Build.VERSION.SDK_INT < Build.VERSION_CODES.R) {
+                                if (zoomState.zoomRatio != 1.0f ||
+                                    zoomState.zoomRatio == currentTransientSettings
+                                        .zoomRatios[currentTransientSettings.primaryLensFacing]
+                                ) {
+                                    currentCameraState.update { old ->
```
temcguir, line 245: Will this cause issues with the zoom ratios camera state update in `ConcurrentCameraSession`?

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraSession.kt
```diff
@@ -150,80 +150,132 @@ internal suspend fun runSingleCameraSession(
...
+                            // reset current camera state when changing cameras.
+                            currentCameraState.update { old ->
+                                old.copy(
+                                    zoomRatios = emptyMap(),
+                                    linearZoomScales = emptyMap()
+                                )
+                            }
+                        }
+                        .collectLatest { zoomState ->
+                            // TODO(): remove checks after buggy zoomState is fixed
```
temcguir, line 239: Add a link to the android issue tracker with the bug for this condition inside the parenthesist of `TODO()`.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraSession.kt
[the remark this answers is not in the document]
Jaehwa-Noh: Thank you for answering.

.idea/codeStyles/Project.xml
temcguir, line 1: Does this need to be checked in?

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraSession.kt
temcguir: We still want to handle zoom changes that come from the transient settings, but we don't want to trigger this block when the cameraState updates, otherwise we're just wasting cycles doing nothing when those updates happen. So if you pull out the `videoRecordingState` (such as through `.transform{}` on the `currentCameraState.asStateFlow()`) , it should cut down the number of times this block needs to run.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraSession.kt
temcguir: Yeah in the capture result callback we need to be careful about making sure which lens we're getting information for. It may require some special handling to ensure we're not sending the wrong results to the camera state.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraSession.kt
Kimblebee: i see, i think i misunderstood your original comment. done!

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraSession.kt
Kimblebee: ah i see, done

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraUseCase.kt
temcguir: Sorry what I meant was the `kotlin.collections.getOrDefault` extension. It can be used like this:
`defaultZoomRatios.getOrDefaultKt(lensFacing, 1.0f) * change.value`
where you have to import:
`import kotlin.collections.getOrDefault as getOrDefaultKt`

temcguir (approved):

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraUseCase.kt
Kimblebee: yeah, the kotlin collections extension directly maps to the jdk8 collections function.  I guess that's where the compatibility issue comes from

PR #294, opened 2024-12-10
VideoQuality proto and VideoQualitySetting

davidjiagoogle:
## Settings module
1. Added the proto and settings.modle files for VideoQuality 
2. SettingsViewModel.setVideoQuality() and all the associated settingsRepository methods
3. VideoQualityUiState in SettingsUiState
4. Associated Ui components and string resources
## core/camera
1. Added VideoQuality in CameraAppSettings
2. supportedVideoQualitiesMap in CameraConstraints

3. setVideoQuality in CameraUseCase
5. Populating logic and tryApplyVideoQualityConstraints in CameraXCameraUseCase
## Preview module
1. Added setVideoQuality to applySettingsDiff()
2. Added videoQuality in PreviewUiState
3. Added VideoQualityIcon and the four drawables to display video quality in preview
4. Added videoResolution in DebugUiState and the associated UI in the debug overlay

https://github.com/user-attachments/assets/4ea32e02-d5ee-48eb-808c-a6d489d1e8f4


feature/settings/src/main/java/com/google/jetpackcamera/settings/SettingsUiState.kt
```diff
@@ -206,6 +209,15 @@ sealed interface MaxVideoDurationUiState {
     ) : MaxVideoDurationUiState
 }
 
+sealed interface VideoQualityUiState {
+    data class Enabled(
+        val currentVideoQuality: VideoQuality,
+        val availableVideoQualities: List<VideoQuality>
```
temcguir, line 215: The pattern we have been using for other settings is to explicitly list out all of the options and use `SingleSelectableState` for each of them. If the device doesn't support one of the options, it will still show up in the list, but will be disabled. See FpsUiState or StabilizationUiState as examples.

feature/settings/src/main/res/values/strings.xml
```diff
@@ -112,6 +112,19 @@
     <string name="duration_description_none">Unlimited duration</string>
     <string name="duration_description_seconds">%d seconds</string>
 
+    <!-- video quality strings -->
+    <string name="video_quality_title">Set Video Quality</string>
+    <string name="video_quality_description_unsupported">Setting Video Quality is not supported for the current camera</string>
```
temcguir, line 117: nit: "Video quality cannot be configured for the current camera"

feature/settings/src/main/java/com/google/jetpackcamera/settings/SettingsViewModel.kt
```diff
@@ -322,6 +325,23 @@ class SettingsViewModel @Inject constructor(
         )
     }
 
+    private fun getVideoQualityUiState(
```
temcguir, line 328: The selected video quality may also not be supported on the other camera lens. You need to disable the default lens flip setting in this case like on line 267 for stabilization or line 236 for FPS.

data/settings/src/main/java/com/google/jetpackcamera/settings/model/VideoQuality.kt
```diff
@@ -0,0 +1,83 @@
...
+    companion object {
+        fun toQuality(videoQuality: VideoQuality): Quality? {
+            return when (videoQuality) {
+                LOWEST -> Quality.LOWEST
+                HIGHEST -> Quality.HIGHEST
+                SD -> Quality.SD
+                HD -> Quality.HD
+                FHD -> Quality.FHD
+                UHD -> Quality.UHD
+                else -> null
```
temcguir, line 38: Since `when` is exhaustive, you should call out explicitly that `DEFAULT -> null` instead of using an `else`.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraSession.kt
```diff
@@ -332,15 +339,24 @@ private fun createVideoUseCase(
...
-    return VideoCapture.Builder(recorder).apply {
+        .setExecutor(backgroundDispatcher.asExecutor())
+    if (videoQuality != VideoQuality.DEFAULT) {
+        recorderBuilder.setQualitySelector(
+            QualitySelector.from(
+                VideoQuality.toQuality(videoQuality)!!,
+                FallbackStrategy.higherQualityOrLowerThan(VideoQuality.toQuality(videoQuality)!!)
+            )
+        )
+    }
```
temcguir, line 358: Try to avoid using `!!`. Use something like:
```
val recorder = Recorder.Builder()
    .setAspectRatio(
        getAspectRatioForUseCase(sensorLandscapeRatio, aspectRatio)
    )
    .setExecutor(backgroundDispatcher.asExecutor())
    .apply {
        VideoQuality.toQuality(videoQuality)?.let { quality ->
            setQualitySelector(QualitySelector.from(quality, ...))
        }
    }.build()
```

data/settings/src/main/java/com/google/jetpackcamera/settings/model/VideoQuality.kt
```diff
@@ -0,0 +1,83 @@
...
+    DEFAULT,
+    LOWEST,
+    HIGHEST,
+    SD,
+    HD,
+    FHD,
+    UHD;
+
+    companion object {
+        fun toQuality(videoQuality: VideoQuality): Quality? {
```
temcguir, line 30: Since this is a conversion to a CameraX class, it should only exist in the core/camera module, not in the model. You can add it as an extension function in `CameraExt.kt` in core/camera:
```
fun VideoQuality.toQuality(): Quality? {
...
}
```

feature/settings/src/main/java/com/google/jetpackcamera/settings/SettingsViewModel.kt
davidjiagoogle: I don't believe this is necessary because in CameraXCameraUseCase, tryApplyVideoQualityConstraints() is added. Upon switching lens, if the video quality is unsupported, it should be set to AUTO. The same mechanism exist for dynamic range

temcguir: As discussed offline, we want the Settings screen to be self-consistent since it is the settings that will be used when the app is launched. So even if we are able to make sure things won't crash, setting the quality to UHD on a camera lens that doesn't support it will mean the app will never start up into a state that matches the settings screen.

Think of the `tryApplyVideoQualityConstraints` as a way to handle incompatibilities in quick settings that we don't want to block access to (such as switching cameras), and anything in the Settings screen should follow all the constraints of other settings in the settings screen.

feature/settings/src/main/java/com/google/jetpackcamera/settings/SettingsViewModel.kt
```diff
@@ -362,6 +363,22 @@ class SettingsViewModel @Inject constructor(
...
         }
 
+        if (currentSettings.videoQuality != VideoQuality.AUTO &&
+            newLensConstraints.supportedVideoQualitiesMap[DynamicRange.SDR]?.contains(
+                currentSettings.videoQuality
+            ) != true
+        ) {
+            return FlipLensUiState.Disabled(
+                currentLensFacing = currentSettings.cameraLensFacing,
+                disabledRationale = StabilizationUnsupportedRationale(
```
temcguir, line 373: I don't think `StabilizationUnsupportedRationale` is the right rationale to use here. You'll need to create a new one for video quality.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewViewModel.kt
```diff
@@ -203,6 +210,22 @@ class PreviewViewModel @AssistedInject constructor(
         }
     }
 
+    private fun getVideoQualityFromCropRect(cropRect: Rect?): VideoQuality {
+        if (cropRect == null) {
+            return VideoQuality.AUTO
```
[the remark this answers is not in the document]
Kimblebee: What should we do in the case of unknown or unspecified? currently other settings handle unknown datastore values with a fallback default value

Kimblebee (commented):
Looks good at a first pass!
One thing though,
in `JcaSettingsSerializer` / `FakeJcaSettingsSerializer` , be sure to include the default value for `defaultValue.setVideoQuality`

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraSession.kt
```diff
@@ -397,14 +402,22 @@ internal fun createVideoUseCase(
...
     val sensorLandscapeRatio = cameraInfo.sensorLandscapeRatio
     val recorder = Recorder.Builder()
         .setAspectRatio(
             getAspectRatioForUseCase(sensorLandscapeRatio, aspectRatio)
         )
-        .setExecutor(backgroundDispatcher.asExecutor()).build()
+        .setExecutor(backgroundDispatcher.asExecutor())
+        .apply {
+            videoQuality.toQuality()?.let { quality ->
+                // No fallback strategy is used. The app will crash if the quality is unsupported
```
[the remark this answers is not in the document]
davidjiagoogle: I commented out the UHD option in the settings for now so that UHD is no longer available as a selectable option. Let's reduce the complexity and address this problem in a separate PR. This one is very cluttered as it is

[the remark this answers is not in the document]
temcguir: I don't know if disabling UHD is sufficient. We don't know that we won't run into the same issue on other devices with FHD, for instance. 

Before addressing this in a separate PR, I would just add back UHD and put in a fallback strategy for this quality selector. Then as long as the UI is showing the actual quality from `CameraState`, any tests can check that the quality actually matches the chosen quality from settings.

temcguir (changes requested):
There are several unresolved comments. I think the most important one has to do with publishing the correct VideoQuality [here](https://github.com/google/jetpack-camera-app/pull/294#discussion_r1893630938). You can either use the suggestion there or use the new CameraX API ([VideoQuality.getSelectedQuality()](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:camera/camera-video/src/main/java/androidx/camera/video/VideoCapture.java;l=361;drc=f24822959fa0cc4dc6b1c5400b1d4ab46fb2c18b)) to get the correct selected quality. When I try the current code, the icon shown does not match with the resolution of the recorded video.

temcguir (approved):
It looks like you still need to run spotless.

PR #298, opened 2024-12-16
Enable Hardware LLB, when FlashMode set to LLB

yasith:
If the FlashMode is set to Low Light Boost, set `CONTROL_AE_MODE` to `CONTROL_AE_MODE_ON_LOW_LIGHT_BOOST_BRIGHTNESS_PRIORITY`

When the FlashMode is changed to something else, clear CaptureRequestOptions. 

I tried updating CaptureRequestOptions with an empty Bundle, and also a Bundle with just clearing CONTROL_AE_MODE, both of these didn't work as expected.


core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraSession.kt
```diff
@@ -244,6 +249,25 @@ internal suspend fun processTransientSettingEvents(
             }
         }
 
+        if (BuildCompat.isAtLeastV()) {
```
temcguir, line 252: Since we're no longer using the preview SDK, this should be changed to:
`Build.VERSION.SDK_INT >= Build.VERSION_CODES.VANILLA_ICE_CREAM`

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraSession.kt
```diff
@@ -878,6 +902,20 @@ private fun Preview.Builder.updateCameraStateWithCaptureResults(
                 result: TotalCaptureResult
             ) {
                 super.onCaptureCompleted(session, request, result)
+
+                if (BuildCompat.isAtLeastV()) {
```
temcguir, line 906: Since we're no longer using the preview SDK, this should be changed to:
`Build.VERSION.SDK_INT >= Build.VERSION_CODES.VANILLA_ICE_CREAM`

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraSession.kt
```diff
@@ -244,6 +249,25 @@ internal suspend fun processTransientSettingEvents(
             }
         }
 
+        if (BuildCompat.isAtLeastV()) {
+            when (newTransientSettings.flashMode) {
```
temcguir, line 253: You should check that the flashMode has changed here before performing this action. It should be gated with a:
```
if (prevTransientSettings.flashMode != newTransientSettings.flashMode) {

}
```

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraSession.kt
```diff
@@ -899,6 +926,28 @@ private fun Preview.Builder.updateCameraStateWithCaptureResults(
...
+
+                        CameraMetadata.CONTROL_LOW_LIGHT_BOOST_STATE_INACTIVE -> {
+                            currentCameraState.update { old ->
+                                old.copy(
+                                    lowLightBoostState = LowLightBoostState.INACTIVE
+                                )
+                            }
+                        }
+                    }
+                }
```
temcguir, line 949: Similar to above, you should check if the LLB state has actually changed before calling `copy`. Something like this:
```
if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.VANILLA_ICE_CREAM) {
    val nativeBoostState = result.get(CaptureResult.CONTROL_LOW_LIGHT_BOOST_STATE)
    val boostState = when (nativeBoostState) {
        CameraMetadata.CONTROL_LOW_LIGHT_BOOST_STATE_ACTIVE -> LowLightBoostState.ACTIVE
        else -> LowLightBoostState.INACTIVE
    }
    currentCameraState.update { old ->
        if (old.lowLightBoostState != boostState) {
            old.copy(lowLightBoostState = boostState)
        } else {
            old
        }
    }
}
```

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewUiState.kt
```diff
@@ -87,6 +88,11 @@ sealed interface StabilizationUiState {
     }
 }
 
+sealed interface LowLightBoostUiState {
+    data object Active : LowLightBoostUiState
+    data object Inactive : LowLightBoostUiState
+}
```
temcguir, line 94: I think this could just be encoded within the `FlashModeUiState.Available` class. Instead of creating a separate class for LLB, just adding an `isActive` property to `FlashModeUiState.Available` should be sufficient. You can then determine when to show different UI when the flash mode is active. This works for low-light boost for now, but in the future we can also extend it to the other flash modes (like making the icon yellow when the torch is on, for example).

temcguir (approved):

PR #300, opened 2024-12-17
update  UseCaseMode to CaptureMode

Kimblebee:
effectively convert `UseCaseMode` to `CaptureMode`, and move into `CameraAppSettings`.
Update references to UseCaseMode account for captureMode now residing in `CameraAppSettings`.

`CameraAppSettings.applyPreviewMode()` in viewModel to initialize the CameraAppSettings captureMode to `VIDEO_ONLY` or `IMAGE_ONLY` if the PreviewMode doesn't support `STANDARD`

### DIFF
--- a/core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraSessionSettings.kt
+++ b/core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraSessionSettings.kt
@@ -17,6 +17,7 @@ package com.google.jetpackcamera.core.camera
 
 import androidx.camera.core.CameraInfo
 import com.google.jetpackcamera.settings.model.AspectRatio
+import com.google.jetpackcamera.settings.model.CaptureMode
 import com.google.jetpackcamera.settings.model.DeviceRotation
 import com.google.jetpackcamera.settings.model.DynamicRange
 import com.google.jetpackcamera.settings.model.FlashMode
@@ -34,9 +35,11 @@ import com.google.jetpackcamera.settings.model.VideoQuality
  */
 internal sealed interface PerpetualSessionSettings {
     val aspectRatio: AspectRatio
+    val captureMode: CaptureMode
 
     data class SingleCamera(
         override val aspectRatio: AspectRatio,
+        override val captureMode: CaptureMode,
         val streamConfig: StreamConfig,
         val targetFrameRate: Int,
         val stabilizationMode: StabilizationMode,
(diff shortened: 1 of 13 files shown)


core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraSessionSettings.kt
```diff
@@ -49,7 +52,9 @@ internal sealed interface PerpetualSessionSettings {
         val primaryCameraInfo: CameraInfo,
         val secondaryCameraInfo: CameraInfo,
         override val aspectRatio: AspectRatio
-    ) : PerpetualSessionSettings
+    ) : PerpetualSessionSettings {
+        override val captureMode: CaptureMode = CaptureMode.VIDEO_ONLY
```
temcguir, line 56: This should be added to the constructor arguments.

Also, why default to `VIDEO_ONLY`?

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewViewModel.kt
```diff
@@ -116,12 +117,24 @@ class PreviewViewModel @AssistedInject constructor(
...
+            isDebugMode = isDebugMode
         ) { cameraPropertiesJSON = it }
     }
 
+    /**
+     * updates the capture mode based on the preview mode
+     */
+    private fun CameraAppSettings.applyPreviewMode(previewMode: PreviewMode): CameraAppSettings {
+        val captureMode = previewMode.toCaptureMode()
+        return if (captureMode == CaptureMode.STANDARD) {
```
temcguir, line 131: Should this "if" condition be `if (captureMode == this.captureMode)`?

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraSessionSettings.kt
Kimblebee: i left it like there since this is for concurrent camera's perpetual settings, which currently only supports `VIDEO_ONLY`. do you still think i should move to constructor?

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewViewModel.kt
Kimblebee: good catch, i think  i glazed over this 😅

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraSessionSettings.kt
temcguir: Ah, gotcha. In that case it's fine how it is. Maybe just add a comment explaining why it's like that.

temcguir (approved):

PR #301, opened 2024-12-17
Refactor HDR settings that were only being used as constants

Kimblebee:


temcguir (approved):

temcguir:
These were originally intended to be settable on the Settings screen, but we don't actually support any other HDR formats besides HLG10 and Jpeg/UltraHDR currently. We can revisit how we would set the default HDR formats later, so this change is OK for now.

PR #305, opened 2025-01-29
ZoomScale persistence 

davidjiagoogle:
Added a private var to keep track of zoomScale of the cameras (mapped to lensFacing). This is needed because camera.cameraInfo.zoomState.zoomScale resets to 1 at some point after runSingleCameraSession. See the buggy video below:

https://github.com/user-attachments/assets/23a5128f-c86f-4b4a-b511-d65595053be8

So instead of using cameraInfo.zoomState to keep track of the current zoomScale, have our own var to do this bypass the bug. Below is the fixed behavior (I also verified lens flipping persistence but didn't include it in the video):

https://github.com/user-attachments/assets/137db24c-bd8d-4c54-9447-451a6b6ee6a0

### DIFF
--- a/core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraSession.kt
+++ b/core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraSession.kt
@@ -111,6 +111,7 @@ private val QUALITY_RANGE_MAP = mapOf(
     HD to Range.create(720, 1079),
     SD to Range.create(241, 719)
 )
+private var zoomScaleMap = mutableMapOf<LensFacing, Float>()
 
 context(CameraSessionContext)
 internal suspend fun runSingleCameraSession(
@@ -216,6 +217,7 @@ internal suspend fun runSingleCameraSession(
             }
 
             applyDeviceRotation(currentTransientSettings.deviceRotation, useCaseGroup)
+            setZoomScale(camera, 1f)
             processTransientSettingEvents(
                 camera,
                 useCaseGroup,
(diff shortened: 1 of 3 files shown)


core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraSession.kt
```diff
@@ -111,6 +111,7 @@ private val QUALITY_RANGE_MAP = mapOf(
     HD to Range.create(720, 1079),
     SD to Range.create(241, 719)
 )
+private var zoomScaleMap = mutableMapOf<LensFacing, Float>()
```
temcguir, line 114: This is essentially a global variable, which we don't want to do. This should be declared within the camera coroutine in CameraXCameraUseCase and added to the CameraSessionContext.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraSession.kt
```diff
@@ -327,6 +320,28 @@ internal suspend fun processTransientSettingEvents(
...
+                zoomScaleMap[transientSettings.primaryLensFacing] = 1f
+            }
+            val oldScale = zoomScaleMap[transientSettings.primaryLensFacing]!!
+            val finalScale =
+                (oldScale * zoomScale).coerceIn(
+                    zoomState.minZoomRatio,
+                    zoomState.maxZoomRatio
+                )
+            camera.cameraControl.setZoomRatio(finalScale)
+            zoomScaleMap[transientSettings.primaryLensFacing] = finalScale
```
temcguir, line 337: We don't necessarily want to persist the zoom when the lens is flipped. We should reset this to 1f when the camera lens flips, which could happen in CameraXCameraUseCase.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraUseCase.kt
```diff
@@ -117,6 +117,9 @@ constructor(
 
     private val _surfaceRequest = MutableStateFlow<SurfaceRequest?>(null)
 
+    private val cameraSessionZoomScale = MutableStateFlow(1f)
+    private var prevCameraSessionLensFacing: LensFacing? = null
```
temcguir, line 121: These don't need to be instance variables. They should be moved to be local variables right after line 291.

temcguir (approved):

PR #306, opened 2025-01-31
Preview Cosmetic Animations

Kimblebee:
Some cosmetic UI changes. summary:

_Minor changes to existing:_
* added padding to bottom capture controls to avoid snackbar overlap
* hide the square, grayish touch indicator present on some clickable components

 _new animations:_
 * slide transition when opening settings screen
* bouncy buttons 
* flip camera icon spins whenever lensfacing changes
* smoother animations for quick settings drop down
* animation for for focusing (expanding) a quick setting

_Mild non-cosmetic refactors:_
* renamed ExpandedQuickSetting to FocusedQuickSetting

**_Android Developers page on_** [**_Animation Modifiers and Composables_**](https://developer.android.com/develop/ui/compose/animation/composables-modifiers)

https://github.com/user-attachments/assets/f4c8818e-951a-4b7d-8d94-215d18227645

image

### DIFF
--- a/app/src/main/java/com/google/jetpackcamera/ui/JcaApp.kt
+++ b/app/src/main/java/com/google/jetpackcamera/ui/JcaApp.kt
@@ -123,18 +123,15 @@ private fun JetpackCameraNavHost(
             route = SETTINGS_ROUTE,
             enterTransition = {
                 fadeIn(
-                    animationSpec = tween(
-                        300,
-                        easing = LinearEasing
-                    )
+                    animationSpec = tween(easing = LinearEasing)
                 ) + slideIntoContainer(
-                    animationSpec = tween(300, easing = EaseIn),
+                    animationSpec = tween(easing = EaseIn),
                     towards = AnimatedContentTransitionScope.SlideDirection.Start
                 )
             },
             exitTransition = {
                 slideOutOfContainer(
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/google/jetpackcamera/ui/JcaApp.kt
```diff
@@ -113,7 +119,26 @@ private fun JetpackCameraNavHost(
                 isDebugMode = isDebugMode
             )
         }
-        composable(SETTINGS_ROUTE) {
+        composable(
+            route = SETTINGS_ROUTE,
+            enterTransition = {
+                fadeIn(
+                    animationSpec = tween(
+                        300,
```
yasith, line 127: nit: maybe we can use 300ms as an animation time constant.

SCREEN_TRANSITION_TIME_MS = 300

or something similar.

yasith (approved):

app/src/main/java/com/google/jetpackcamera/ui/JcaApp.kt
Kimblebee: i dug in a little and it looks like the default value from `AnimationConstants` is already 300ms, so i'll just remove the usages of 300 and leave it as the default

PR #308, opened 2025-02-06
New Capture Button States

Kimblebee:
Big change:
* New functionality and appearances for the capture button based on the current `CaptureMode` and `VideoRecordingState`
Capture button's functionality and appearance is based on the new `CaptureButtonUiState`.
* CaptureButtonUiState is either `Idle(CaptureMode)`, `PressedRecording`, or `LockedRecording`.

Currently, different capture modes can only be set using the HDR or Conncurrent Camera settings.
Behaviors for different capture modes:
`STANDARD`:
* Tap for image capture
* Long press to start recording / release to stop

`IMAGE_ONLY`:
* Tap for image capture
* Long press is no op

`VIDEO_ONLY`:
* Tap to start recording in locked state (no need to keep finger pressed on screen).
* Tap while recording is active to stop recording
* Long press to start recording / release to stop

`ImageCaptureDeviceTest` and `VideoRecordingDeviceTest` updated to account for the new behaviors

https://github.com/user-attachments/assets/df3e481d-251b-4e39-a6f2-ec71e3a166ae

**_the capture mode selector is not included in this PR_**


feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
```diff
@@ -153,39 +159,54 @@ fun ElapsedTimeText(
...
-            imageVector = when (currentRecordingState) {
-                is VideoRecordingState.Active.Recording -> Icons.Filled.Pause
-                is VideoRecordingState.Active.Paused -> Icons.Filled.PlayArrow
-            },
-            contentDescription = "pause resume toggle"
-        )
+                .size(size = size.dp)
+                .scale(scale = animatedToggleScale)
+                .clip(CircleShape)
+                // .border(4.dp, Color.White, CircleShape)
```
temcguir, line 193: nit: remove this comment?

app/src/androidTest/java/com/google/jetpackcamera/VideoRecordingDeviceTest.kt
```diff
@@ -104,20 +105,21 @@ internal class VideoRecordingDeviceTest {
...
+                composeTestRule.tapStartLockedVideoRecording()
+
+                // stop recording
+                composeTestRule.onNodeWithTag(CAPTURE_BUTTON).assertExists().performClick()
             }
-        Truth.assertThat(result.resultCode).isEqualTo(Activity.RESULT_CANCELED)
-        Truth.assertThat(doesImageFileExist(uri, "video")).isFalse()
+        Truth.assertThat(result.resultCode).isEqualTo(Activity.RESULT_OK)
+        Truth.assertThat(doesImageFileExist(uri, "image")).isFalse()
+        Truth.assertThat(doesImageFileExist(uri, "video")).isTrue()
```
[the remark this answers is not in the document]
Kimblebee: this is awesome, i think referencing this will greatly improve some of our existing tests, and help with new ones

app/src/androidTest/java/com/google/jetpackcamera/utils/UiTestUtil.kt
```diff
@@ -192,15 +195,60 @@ fun deleteFilesInDirAfterTimestamp(
...
     return false
 }
 
+fun doesImageExist(uri: Uri): Boolean {
+    val bitmap = uri.path?.let { path -> BitmapFactory.decodeFile(path) }
+    val mimeType = URLConnection.guessContentTypeFromName(uri.path)
+    return mimeType != null && mimeType.startsWith(IMAGE_PREFIX) && bitmap != null
+}
+
+// this does not work with the image capture tests...
```
temcguir, line 215: If this is the case, then you should be validating arguments. Add something like this to the beginning of the function:
```
require(prefix == VIDEO_PREFIX) { "doesVideoExist() only works for videos. Can't handle prefix: $prefix" }
```

app/src/androidTest/java/com/google/jetpackcamera/utils/UiTestUtil.kt
```diff
@@ -192,15 +195,60 @@ fun deleteFilesInDirAfterTimestamp(
...
+            return@useAndRelease false
+        }
+
+        if (durationMs != null && it.getDurationMs() != durationMs) {
+            return@useAndRelease false
+        }
+
+        return@useAndRelease true
+    }
+    return hasCorrectContents == true
```
temcguir, line 249: Assuming you check that `VIDEO_PREFIX` is used for the prefix and you get rid of the `try/catch` block, this set of cases / return statements can be simplified to:
```
    return MediaMetadataRetriever().useAndRelease {
        it.setDataSource(uri.path)

        it.getMimeType().startsWith(prefix) &&
        it.hasVideo() &&
        (!checkAudio || it.hasAudio()) &&
        (durationMs == null || it.getDurationMs() == durationMs)
    } == true
```

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
```diff
@@ -733,61 +750,163 @@ fun CurrentCameraIdText(physicalCameraId: String?, logicalCameraId: String?) {
...
+                exit = fadeOut()
+            ) {
+                var smallBoxSize = (captureButtonSize / 5f).dp
+                Box(
+                    modifier = Modifier
+                        .size(smallBoxSize)
+                        .clip(RoundedCornerShape(smallBoxSize * .15f))
+                        .background(color = Color.White)
+                )
             }
```
temcguir, line 908: This seems to still have some jerkiness when animating. It's likely due to the fact that it is placed inside the outer circle box that is changing side, so the center position is changing while animating.

You should just unnest this box from the "inner circle" box. In fact, it's probably better to not use `Box`s here at all, as they require going through layout. Instead, you could use `Canvas` for each of these elements, like:
```
        // inner circle
        Canvas(modifier = Modifier) {
            drawCircle(
                color = animatedColor.copy(
                    alpha =
                    if (isPressedDown &&
                        currentUiState.value ==
                        CaptureButtonUiState.Enabled.Idle(CaptureMode.IMAGE_ONLY)
                    ) {
                        .5f // transparency to indicate click ONLY on IMAGE_ONLY
                    } else {
                        1f // solid alpha the rest of the time
                    }
                ),
                radius = centerShapeSize.toPx() / 2f
            )
        }
        // central "square" stop icon
        AnimatedVisibility(
            visible = currentUiState.value is
                    CaptureButtonUiState.Enabled.Recording.LockedRecording,
            enter = scaleIn(initialScale = .5f) + fadeIn(),
            exit = fadeOut()
        ) {
            val smallBoxSize = (captureButtonSize / 5f).dp
            Canvas(modifier = Modifier) {
                drawRoundRect(
                    color = Color.White,
                    topLeft = Offset(-smallBoxSize.toPx() / 2f, -smallBoxSize.toPx() / 2f),
                    size = Size(smallBoxSize.toPx(), smallBoxSize.toPx()),
                    cornerRadius = CornerRadius(smallBoxSize.toPx() * .15f)
                )
            }
        }
```

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
```diff
@@ -733,61 +750,163 @@ fun CurrentCameraIdText(physicalCameraId: String?, logicalCameraId: String?) {
...
+                .background(animatedColor)
+        ) {
+            // central "square" stop icon
+            AnimatedVisibility(
+                visible = currentUiState.value is
+                    CaptureButtonUiState.Enabled.Recording.LockedRecording,
+                enter = scaleIn(initialScale = .5f) + fadeIn(),
+                exit = fadeOut()
+            ) {
+                var smallBoxSize = (captureButtonSize / 5f).dp
```
temcguir, line 901: nit: this should be a `val`.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
```diff
@@ -733,61 +750,164 @@ fun CurrentCameraIdText(physicalCameraId: String?, logicalCameraId: String?) {
...
                     },
                     // TODO: @kimblebee - stopVideoRecording is being called every time the capture
                     // button is pressed -- regardless of tap or long press
                     onPress = {
                         isPressedDown = true
                         awaitRelease()
                         isPressedDown = false
-                        onRelease()
+                        isLongPressing = false
+                        var uiState = currentUiState.value
```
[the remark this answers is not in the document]
temcguir: This still should be a `val`.

app/src/androidTest/java/com/google/jetpackcamera/utils/UiTestUtil.kt
```diff
@@ -191,15 +195,60 @@ fun deleteFilesInDirAfterTimestamp(
...
+    durationMs: Long? = null
+): Boolean {
+    if (!doesFileExist(uri)) {
+        return false
+    }
+    var hasCorrectContents = MediaMetadataRetriever().useAndRelease {
+        try {
+            it.setDataSource(uri.path)
+        } catch (e: Exception) {
+            Log.e(TAG, e.stackTraceToString())
```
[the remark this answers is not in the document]
Kimblebee: oops. that was leftover from trying to figure out why this wasn't working for images.

temcguir (approved):

PR #310, opened 2025-02-07
Elapsed time UI state

Kimblebee:
small refactor to give elapsed time a uistate.

should be visible long enough for future tests that want to track duration

### DIFF
--- a/feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewUiState.kt
+++ b/feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewUiState.kt
@@ -67,10 +67,7 @@ data class DebugUiState(
 sealed interface ElapsedTimeUiState {
     data object Unavailable : ElapsedTimeUiState
 
-    data class Enabled(
-        val elapsedTimeNanos: Long,
-        val maxVideoDurationMillis: Long
-    ) : ElapsedTimeUiState
+    data class Enabled(val elapsedTimeNanos: Long) : ElapsedTimeUiState
 }
 
 sealed interface AudioUiState {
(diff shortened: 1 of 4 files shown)


feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
```diff
@@ -129,24 +127,14 @@ private const val TAG = "PreviewScreen"
...
-        Text(
-            modifier = modifier,
-            text = elapsedNs.nanoseconds.toComponents { minutes, seconds, _ ->
-                "%02d:%02d".format(minutes, seconds)
-            },
-            textAlign = TextAlign.Center
-        )
-    }
+fun ElapsedTimeText(modifier: Modifier = Modifier, elapsedTimeUiState: ElapsedTimeUiState.Enabled) {
+    val currentUiState = rememberUpdatedState(elapsedTimeUiState)
```
temcguir, line 131: You shouldn't need this. Changes to the `elapsedTimeUiState` argument will already cause a recomposition. You're not using it anywhere outside of composition (such as in a `LaunchedEffect`).

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewUiState.kt
```diff
@@ -53,19 +52,27 @@ sealed interface PreviewUiState {
...
     val isDebugMode: Boolean = false,
     val isDebugOverlayOpen: Boolean = false
 )
 
+sealed interface ElapsedTimeUiState {
+    data object Unavailable : ElapsedTimeUiState
+
+    data class Enabled(
+        val elapsedTimeNanos: Long,
+        val maxVideoDurationMillis: Long
```
temcguir, line 72: Is this being used anywhere in the UI or can it be removed?

Kimblebee: it isn't currently in use but i thought about including some visual indicator the presence of a duration limit.

 i.e. an animated circle  based on the max duration, decrementing the elapsed time text instead of incrementing, appending the max duration to the elapsed time text, etc. I figured I could do that another time.

temcguir (approved):

PR #311, opened 2025-02-10
Volume button capture

Kimblebee:
features:
* as long as the capture button is present, you can also initiate capture using volume buttons
* supports volume up / down and selfie sticks that use volume keycode
* pressed recording can only be stopped by the button that initiated the capture. 
* locked recordings can be stopped with any button

notable changes:
* Majority of important changes occur `PreviewScreenComponents.kt`
* new `CaptureKeyHandler` composable  attached to capture button. Listens for the appropriate key events and calls `onpress`/ `onKeyUp`
* capturebutton functionality hoisted up into their own functions, so they can be reused by all capture initiators.
* mutable states local to capture button so that capture sources do not interfere with each other


feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
```diff
@@ -130,15 +136,165 @@ import com.google.jetpackcamera.settings.model.LensFacing
...
+            }
+        }
+
+        ViewCompat.addOnUnhandledKeyEventListener(view, keyEventDispatcher)
+
+        onDispose {
+            ViewCompat.removeOnUnhandledKeyEventListener(view, keyEventDispatcher)
+        }
+    }
+}
```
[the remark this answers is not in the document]
Kimblebee: no, this is an accident. I just meant to rename `VolumeButtonsHandler` to `CaptureKeyHandler` Thanks for pointing out!

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
```diff
@@ -903,6 +1105,53 @@ fun CaptureButton(
...
+    fun keyCodeToCaptureSource(keyCode: Int): CaptureSource = when (keyCode) {
+        KeyEvent.KEYCODE_VOLUME_UP -> CaptureSource.VOLUME_UP
+        KeyEvent.KEYCODE_VOLUME_DOWN -> CaptureSource.VOLUME_DOWN
+        else -> TODO("Keycode not assigned to CaptureSource")
+    }
+
+    DisposableEffect(context) {
+        val keyEventDispatcher = ViewCompat.OnUnhandledKeyEventListenerCompat { _, event ->
+            when (event.keyCode) {
+                KeyEvent.KEYCODE_VOLUME_UP, KeyEvent.KEYCODE_VOLUME_DOWN -> {
```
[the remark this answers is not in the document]
Kimblebee: just to avoid repetition -- keeps the code a bit less bloated :-) .  pressing the up or down volume key will perform the same action.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
```diff
@@ -130,15 +136,165 @@ import com.google.jetpackcamera.settings.model.LensFacing
...
```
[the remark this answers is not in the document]
Kimblebee: answered in [this thread](https://github.com/google/jetpack-camera-app/pull/311#discussion_r1966972865)

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
```diff
@@ -130,15 +136,165 @@ import com.google.jetpackcamera.settings.model.LensFacing
...
+                    Log.d(TAG, "Stopping recording")
+                    onStopRecording()
+                }
+            }
+            // on click
+            else {
+                when (val current = currentUiState.value) {
+                    is CaptureButtonUiState.Enabled.Idle -> when (current.captureMode) {
+                        CaptureMode.STANDARD,
+                        CaptureMode.IMAGE_ONLY -> onImageCapture()
```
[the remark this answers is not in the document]
Kimblebee: If we were to capture an image onPress, we would completely ignore the potential for that press to become a long press. 

For example, If I wanted to start a long press to record a video, I would be confused if it captured an image first!

Therefore, we want to perform the image capture on click (press then quickly release)

I think an exception to this would be when the app is using the `CaptureMode.IMAGE_ONLY` capture mode, where you will never be able to record a video. But for consistency, I think we should keep the same onClick functionality as `CaptureMode.STANDARD`

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
```diff
@@ -903,6 +1105,53 @@ fun CaptureButton(
...
+        }
+
+        ViewCompat.addOnUnhandledKeyEventListener(view, keyEventDispatcher)
+
+        onDispose {
+            ViewCompat.removeOnUnhandledKeyEventListener(view, keyEventDispatcher)
+        }
+    }
+}
+
```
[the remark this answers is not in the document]
Kimblebee: answered in [this thread](https://github.com/google/jetpack-camera-app/pull/311#discussion_r1966972138)

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
```diff
@@ -741,88 +751,133 @@ fun CurrentCameraIdText(physicalCameraId: String?, logicalCameraId: String?) {
...
-        mutableStateOf(false)
+    var currentUiState = rememberUpdatedState(captureButtonUiState)
+    val firstKeyPressed = remember { mutableStateOf<CaptureSource?>(null) }
+    val isLongPressing = remember { mutableStateOf<Boolean>(false) }
+    var longPressJob by remember { mutableStateOf<Job?>(null) }
+    val scope = rememberCoroutineScope()
+
+    suspend fun onLongPress() {
+        delay(ViewConfiguration.getLongPressTimeout().toLong())
+        withContext(NonCancellable) {
```
temcguir, line 778: Nothing inside this block is suspending, so this `withContext()` won't do anything.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
```diff
@@ -741,88 +751,133 @@ fun CurrentCameraIdText(physicalCameraId: String?, logicalCameraId: String?) {
...
-                            CaptureButtonUiState.Enabled.Recording.LockedRecording -> {
-                                onStopVideoRecording()
-                            }
-
-                            CaptureButtonUiState.Unavailable,
-                            CaptureButtonUiState.Enabled.Recording.PressedRecording -> {
-                            }
-                        }
+                        isCaptureButtonPressed = false
+                        onRelease(CaptureSource.CAPTURE_BUTTON)
```
temcguir, line 880: nit: This no longer handles the "Cancel" event from a press. For instance, if you press inside the shutter button and then quickly drag your finger outside of the shutter button, it would have cancelled taking a picture with the previous code ("onTap" wouldn't have been called), but now it will take a picture. We can talk about how to fix this later.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
```diff
@@ -741,88 +751,133 @@ fun CurrentCameraIdText(physicalCameraId: String?, logicalCameraId: String?) {
...
+    onStopRecording: () -> Unit,
     onLockVideoRecording: (Boolean) -> Unit,
     captureButtonUiState: CaptureButtonUiState,
     captureButtonSize: Float = 80f
 ) {
-    val currentUiState = rememberUpdatedState(captureButtonUiState)
-    var isPressedDown by remember {
-        mutableStateOf(false)
+    var currentUiState = rememberUpdatedState(captureButtonUiState)
+    val firstKeyPressed = remember { mutableStateOf<CaptureSource?>(null) }
```
temcguir, line 771: Rather than keeping track of the first key pressed, could we just keep track of whether _any_ of the keys are pressed? Then when the last one is released, we treat it as if the capture button is released.

There are some good examples on how to do this with `MutableInteractionSource` in the documentation. I think that could be a good approach for this.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
```diff
@@ -741,88 +751,133 @@ fun CurrentCameraIdText(physicalCameraId: String?, logicalCameraId: String?) {
...
+    val isLongPressing = remember { mutableStateOf<Boolean>(false) }
+    var longPressJob by remember { mutableStateOf<Job?>(null) }
+    val scope = rememberCoroutineScope()
+
+    suspend fun onLongPress() {
+        delay(ViewConfiguration.getLongPressTimeout().toLong())
```
temcguir, line 777: nit: use `LocalViewConfiguration.current` to get the `longPressTimeoutMillis`.

temcguir, line 777: Rather than delaying at the beginning of `onLongPress`, you should delay at the beginning of the `scope.launch` where you call `onLongPress`. With that organization, it's clearer that `onLongPress` means you've reached a long press event.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
```diff
@@ -741,88 +751,133 @@ fun CurrentCameraIdText(physicalCameraId: String?, logicalCameraId: String?) {
...
+    onStartRecording: () -> Unit,
+    onStopRecording: () -> Unit,
     onLockVideoRecording: (Boolean) -> Unit,
     captureButtonUiState: CaptureButtonUiState,
     captureButtonSize: Float = 80f
 ) {
-    val currentUiState = rememberUpdatedState(captureButtonUiState)
-    var isPressedDown by remember {
-        mutableStateOf(false)
+    var currentUiState = rememberUpdatedState(captureButtonUiState)
```
temcguir, line 770: You can use `by` instead of `=` here (and for other `State` variables) so you don't need to access everything with `.value` inside your lambdas.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
```diff
@@ -903,6 +959,53 @@ fun CaptureButton(
...
+        KeyEvent.KEYCODE_VOLUME_DOWN -> CaptureSource.VOLUME_DOWN
+        else -> TODO("Keycode not assigned to CaptureSource")
+    }
+
+    DisposableEffect(context) {
+        val keyEventDispatcher = ViewCompat.OnUnhandledKeyEventListenerCompat { _, event ->
+            when (event.keyCode) {
+                KeyEvent.KEYCODE_VOLUME_UP, KeyEvent.KEYCODE_VOLUME_DOWN -> {
+                    val captureSource = keyCodeToCaptureSource(event.keyCode)
+                    if (event.action == KeyEvent.ACTION_DOWN) {
```
temcguir, line 984: Since volume buttons are meant to turn up/down volume, the "ACTION_DOWN" event will be called continually until you release. We probably only need to call `onPress()` once until it's released. You can keep a boolean `var` in this `DisposableEffect` to check if `ACTION_DOWN` has been called already, and set it to `false` when `ACTION_UP` is called.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
```diff
@@ -903,6 +959,53 @@ fun CaptureButton(
...
+                        onRelease(captureSource)
+                    }
+                    // consume the event
+                    true
+                }
+                else -> {
+                    false
+                }
+            }
+        }
```
Jaehwa-Noh, line 999: It is as though consumed a VOLUME UP and DOWN key events regardless the users intend.
Is there any chance of adding a feature that turn this feature on and off in the future?

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
Kimblebee: Yeah it was something I considered. I ended up following first key pressed pattern because I found that to be the behavior in the default camera app and few other popular social media apps.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
Kimblebee: we consume the event here because we only want to capture an image; we don't want to also increase the volume. Similar behavior to other camera apps. This behavior only occurs while the capture button is present... i.e. whenever the user would want to capture an image.

I can see an issue arising when using JCA in split screen mode on foldable devices; someone may intend to increase the volume while the camera is open. I think in that case it would make sense to would either disable the volume capture feature completely or not consume the event. That's something we'd need to look a bit into and test!

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
Kimblebee: i'm not entirely sure why yet but this doesn't work for the captureButtonUiState.
the UI visuals will not update properly, nor will the recording stop when you lift off the capture button.

Its as if all functions that refer directly to `captureButtonUiState` or `currentUiState by remember{mutableStateOf(captureButtonUiState)}` only observe the initial value when the component is first composed, but don't recognize any subsequent updates

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
temcguir: Ok SGTM. I played around with the default camera app on a device I had and saw similar behavior. However, it didn't feel very natural to me. It would be nice to be able to "transfer" my finger from the volume button to the screen, for instance when recording a video (this would also allow locking the video, which wouldn't be possible with the volume button). I wonder if there is a good reason the default camera app didn't implement it that way or if it is just because it would have been harder to implement.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
temcguir: Ah you're right. I think this has to do with the way compose captures values in lambdas during composition. I believe the value is taken from the MutableState and inserted into the lambda during the first composition that creates the lambda. So using the MutableState directly would be the way to go. 

One way to avoid using the MutableState would be to `remember(captureUiState) { <lambda> }`. This would recreate the lambda any time the captureUiState changes. However, I think the way it is being done now is probably better for performance.

temcguir (approved):

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
Kimblebee: i think it may have been an edge case that was overlooked. I would need to try the default camera app on a few other device manufacturers to see if they all follow similar patterns.
Because the behavior was the same for the social media apps that I looked at, I decided to just follow that pattern. I'll merge for now, and this can be something revisited later :-)

PR #312, opened 2025-02-10
Add test tags to SettingsComponents

davidjiagoogle:


feature/settings/src/main/java/com/google/jetpackcamera/settings/ui/SettingsComponents.kt
```diff
@@ -235,25 +238,29 @@ fun FlashModeSetting(
...
                 SingleChoiceSelector(
+                    modifier = Modifier.testTag(FLASH_MODE_OPTION_OFF_TAG),
                     text = stringResource(id = R.string.flash_mode_selector_off),
                     selected = flashUiState.currentFlashMode == FlashMode.OFF,
                     enabled = true,
                     onClick = { setFlashMode(FlashMode.OFF) }
                 )
                 // TODO(yasith): Add logic to only show LLB toggle if current lens supports LLB
                 SingleChoiceSelector(
+                    modifier = Modifier.testTag(FLASH_MODE_OPTION_LOW_LIGHT_BOOST_TAG),
```
Kimblebee, line 263: nit: could shorten to `FLASH_MODE_OPTION_LLB_TAG`

feature/settings/src/main/java/com/google/jetpackcamera/settings/ui/SettingsComponents.kt
```diff
@@ -139,7 +139,7 @@ fun DarkModeSetting(
     modifier: Modifier = Modifier
 ) {
     BasicPopupSetting(
-        modifier = modifier,
+        modifier = modifier.testTag(DARK_MODE_SELECTOR_TAG),
```
Kimblebee, line 142: nit: could just do DARK_MODE_SETTING_TAG for clarity instead of DARK_MODE_SELECTOR. Esp. since there is a different component called `singleChoiceSelector`

feature/settings/src/main/java/com/google/jetpackcamera/settings/ui/TestTags.kt
```diff
@@ -32,11 +32,63 @@ const val DEVICE_UNSUPPORTED_TAG = "DeviceUnsupportedTag"
...
+const val STREAM_CONFIG_OPTION_SINGLE_STREAM_TAG = "StreamConfigOptionSingleStreamTag"
+const val STREAM_CONFIG_OPTION_MULTI_STREAM_CAPTURE_TAG = "StreamConfigOptionMultiStreamCaptureTag"
+
+// Max Video Duration
+const val MAX_VIDEO_DURATION_SELECTOR_TAG = "MaxVideoDurationSelectorTag"
+const val MAX_VIDEO_DURATION_OPTION_UNLIMITED_TAG = "MaxVideoDurationOptionUnlimitedTag"
+const val MAX_VIDEO_DURATION_OPTION_ONE_SEC_TAG = "MaxVideoDurationOptionOneSecTag"
+const val MAX_VIDEO_DURATION_OPTION_TEN_SEC_TAG = "MaxVideoDurationOptionTenSecTag"
+const val MAX_VIDEO_DURATION_OPTION_THIRTY_SEC_TAG = "MaxVideoDurationOptionThirtySecTag"
+const val MAX_VIDEO_DURATION_OPTION_SIXTY_SEC_TAG = "MaxVideoDurationOptionSixtySecTag"
```
Kimblebee, line 70: nit: numbers in the tag instead of the spelling to match precedent set by fps and aspect ratio tags.

feature/settings/src/main/java/com/google/jetpackcamera/settings/ui/TestTags.kt
Kimblebee, line 1: moving forward, i actually think we should use a different text for the tags themselves. something that would keep the prefixes consistent and easily readable.
I found [**this**](https://github.com/umesh0492/android-guidelines/blob/master/project_and_code_guidelines.md#232-resources-naming) guidelines bible for an old project that establishes a great naming convention scheme. I think we can adopt something similar for JCA.

flash mode is a dialog setting, for example could be:
```
const val FLASH_MODE_SETTINGS_TAG = "btn_open_dialog_setting_flash"
const val FLASH_MODE_OPTION_AUTO_TAG = "btn_dialog_flash_option_auto"
const val FLASH_MODE_OPTION_ON_TAG = "btn_dialog_flash_option_on"
const val FLASH_MODE_OPTION_OFF_TAG = "btn_dialog_flash_option_off"
const val FLASH_MODE_OPTION_LLB_TAG = "btn_dialog_flash_option_llb"
```
lens facing is a switch setting, for example could be:
`const val DEFAULT_CAMERA_FACING_SWITCH_TAG = "btn_switch_setting_lensFacing"`

davidjiagoogle: Done. Renamed to keep the screaming snake notation and also adhere to android guideline

PR #315, opened 2025-02-19
add constraints to flash setting

Kimblebee:
Default flash setting will now be disabled with an appropriate message when:
* device only supports `OFF`
* default lens only supports `OFF`

Default flash options will be disabled if the current lens doesnt support the particular mode.

important stuff to check is in `SettingsViewModel`

### DIFF
--- a/data/settings/src/main/java/com/google/jetpackcamera/settings/model/Constraints.kt
+++ b/data/settings/src/main/java/com/google/jetpackcamera/settings/model/Constraints.kt
@@ -15,24 +15,15 @@
  */
 package com.google.jetpackcamera.settings.model
 
-import kotlin.reflect.KProperty1
-
 data class SystemConstraints(
     val availableLenses: List<LensFacing> = emptyList(),
     val concurrentCamerasSupported: Boolean = false,
     val perLensConstraints: Map<LensFacing, CameraConstraints> = emptyMap()
-) {
+)
 
-    inline fun <reified R> forDevice(property: KProperty1<CameraConstraints, Any>): R =
-        perLensConstraints.values.asSequence()
-            .flatMap {
-                when (val value = property.get(it)) {
-                    is Iterable<*> -> value.asSequence()
-                    is Sequence<*> -> value
-                    else -> sequenceOf(value)
-                }
-            }.toSet() as R
-}
+inline fun <reified T> SystemConstraints.forDevice(
+    crossinline constraintSelector: (CameraConstraints) -> Iterable<T>
+) = perLensConstraints.values.asSequence().flatMap { constraintSelector(it) }.toSet()
 
 data class CameraConstraints(
     val supportedStabilizationModes: Set<StabilizationMode>,
--- a/data/settings/src/main/java/com/google/jetpackcamera/settings/model/Constraints.kt
+++ b/data/settings/src/main/java/com/google/jetpackcamera/settings/model/Constraints.kt
@@ -15,11 +15,24 @@
  */
 package com.google.jetpackcamera.settings.model
 
+import kotlin.reflect.KProperty1
+
 data class SystemConstraints(
     val availableLenses: List<LensFacing> = emptyList(),
     val concurrentCamerasSupported: Boolean = false,
     val perLensConstraints: Map<LensFacing, CameraConstraints> = emptyMap()
-)
+) {
+
+    inline fun <reified R> forDevice(property: KProperty1<CameraConstraints, Any>): R =
+        perLensConstraints.values.asSequence()
+            .flatMap {
+                when (val value = property.get(it)) {
+                    is Iterable<*> -> value.asSequence()
+                    is Sequence<*> -> value
+                    else -> sequenceOf(value)
+                }
+            }.toSet() as R
+}
 
 data class CameraConstraints(
     val supportedStabilizationModes: Set<StabilizationMode>,
--- a/feature/settings/src/main/java/com/google/jetpackcamera/settings/SettingsUiState.kt
+++ b/feature/settings/src/main/java/com/google/jetpackcamera/settings/SettingsUiState.kt
@@ -104,8 +104,7 @@ sealed interface DisabledRationale {
     data class VideoQualityUnsupportedRationale(
         override val affectedSettingNameResId: Int,
         val currentDynamicRange: Int = R.string.video_quality_rationale_suffix_default
-    ) :
-        DisabledRationale {
+    ) : DisabledRationale {
         override val reasonTextResId = R.string.video_quality_unsupported
         override val testTag = VIDEO_QUALITY_UNSUPPORTED_TAG
     }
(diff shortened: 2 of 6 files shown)


yasith (approved):

feature/settings/src/main/java/com/google/jetpackcamera/settings/SettingsUiState.kt
```diff
@@ -189,25 +188,28 @@ sealed interface AudioUiState {
...
         data class Mute(override val additionalContext: String = "") : Enabled
     }
 
     data class Disabled(val disabledRationale: DisabledRationale) : AudioUiState
 }
 
+sealed interface FlashUiState {
+    data class Enabled(
+        val currentFlashMode: FlashMode,
+        val lowLightSelectableState: SingleSelectableState,
```
temcguir, line 202: This is fine to ensure LLB is disabled when it's not available, but we also need to do this for ON and AUTO (and then disable the setting altogether if OFF is the only available mode). We currently do this in the quick setting UI and for other settings like Stabilization. I'm fine with doing that in a separate PR if needed.

temcguir: Emulators are an example of a device with only OFF as an available flash mode.

Kimblebee: done. could you please take another look?

data/settings/src/main/java/com/google/jetpackcamera/settings/model/Constraints.kt
```diff
@@ -15,11 +15,24 @@
...
+    inline fun <reified R> forDevice(property: KProperty1<CameraConstraints, Any>): R =
+        perLensConstraints.values.asSequence()
+            .flatMap {
+                when (val value = property.get(it)) {
+                    is Iterable<*> -> value.asSequence()
+                    is Sequence<*> -> value
+                    else -> sequenceOf(value)
+                }
+            }.toSet() as R
+}
```
temcguir, line 35: Instead of using reflection here, you can use a higher order function which can be satisfied by property references. So instead of this, you can replace it with:
```
inline fun <reified T> SystemConstraints.forDevice(
    crossinline constraintSelector: (constraints: CameraConstraints) -> Iterable<T>
) = perLensConstraints.values.asSequence().flatMap { constraintSelector(it) }.toSet()
```

Since I believe all of the CameraConstraints are currently `Iterable`, this should be sufficient. If we need a `Sequence` version in the future, or simply a non-collection version, you could add overloads:

```
inline fun <reified T> SystemConstraints.forDevice(
    crossinline constraintSelector: (constraints: CameraConstraints) -> Sequence<T>
) = perLensConstraints.values.asSequence().flatMap { constraintSelector(it) }.toSet()

inline fun <reified T> SystemConstraints.forDevice(
    crossinline constraintSelector: (constraints: CameraConstraints) -> T
) = perLensConstraints.values.asSequence().flatMap { sequenceOf(constraintSelector(it)) }.toSet()
```

feature/settings/src/main/java/com/google/jetpackcamera/settings/SettingsViewModel.kt
```diff
@@ -103,6 +101,93 @@ class SettingsViewModel @Inject constructor(
...
+        check(currentSupportedFlashModes.isNotEmpty()) {
+            "No flash modes supported. Should at least support OFF."
+        }
+        val deviceSupportedFlashModes: Set<FlashMode> = constraints.forDevice(
+            CameraConstraints::supportedFlashModes
+        )
+        // disable entire setting when:
+        //  device only supports off... device unsupported rationale
+        //  lens only supports off... lens unsupported rationale
+        if (deviceSupportedFlashModes.size == 1) {
```
temcguir, line 120: nit: to be more precise with this `if` condition, use:
`if (deviceSupportedFlashModes == setOf(FlashMode.OFF))`

You can use this same check below for `currentSupportedFlashModes`.

feature/settings/src/main/java/com/google/jetpackcamera/settings/ui/SettingsComponents.kt
```diff
@@ -221,51 +221,58 @@ fun FlashModeSetting(
...
+                        onClick = { setFlashMode(FlashMode.ON) }
+                    )
+                    SingleChoiceSelector(
+                        modifier = Modifier.testTag(BTN_DIALOG_FLASH_OPTION_OFF_TAG),
+                        text = stringResource(id = R.string.flash_mode_selector_off),
+                        selected = flashUiState.currentFlashMode == FlashMode.OFF,
+                        enabled = true,
+                        onClick = { setFlashMode(FlashMode.OFF) }
+                    )
+                    // TODO(yasith): Add logic to only show LLB toggle if current lens supports LLB
```
temcguir, line 266: I think this comment can be removed.

temcguir, line 266: nit: we should move OFF after LLB. It should be the last option.

temcguir (approved):

feature/settings/src/main/java/com/google/jetpackcamera/settings/SettingsUiState.kt
temcguir: Thank you! Looks good.

data/settings/src/main/java/com/google/jetpackcamera/settings/model/Constraints.kt
Kimblebee: oh wow, thats really amazing!

feature/settings/src/main/java/com/google/jetpackcamera/settings/SettingsViewModel.kt
```diff
@@ -111,9 +112,9 @@ class SettingsViewModel @Inject constructor(
         check(currentSupportedFlashModes.isNotEmpty()) {
             "No flash modes supported. Should at least support OFF."
         }
-        val deviceSupportedFlashModes: Set<FlashMode> = constraints.forDevice(
-            CameraConstraints::supportedFlashModes
-        )
+        val deviceSupportedFlashModes: Set<FlashMode> = constraints.forDevice {
+            it.supportedFlashModes
```
temcguir, line 116: This is fine, but does `constraints.forDevice(CameraConstraints::supportedFlashModes)` not work here?

temcguir (approved):

feature/settings/src/main/java/com/google/jetpackcamera/settings/SettingsViewModel.kt
Kimblebee: it does, but I dont think I fully understand why one would be better than the other here

PR #316, opened 2025-02-21
disable single stream quick setting while Ultra HDR Image capture is active

Kimblebee:
user will no longer be able to switch to single-stream capture while HDR image capture is on

### DIFF
--- a/feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/quicksettings/QuickSettingsScreen.kt
+++ b/feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/quicksettings/QuickSettingsScreen.kt
@@ -154,6 +154,8 @@ private enum class FocusedQuickSetting {
     ASPECT_RATIO
 }
 
+// todo: Add UI states for Quick Settings buttons
+
 /**
  * The UI component for quick settings when it is focused.
  */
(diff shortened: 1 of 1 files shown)


feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/quicksettings/QuickSettingsScreen.kt
```diff
@@ -222,8 +224,12 @@ private fun ExpandedQuickSettingsUi(
                             ),
                             setStreamConfig = { c: StreamConfig -> onStreamConfigClick(c) },
                             currentStreamConfig = currentCameraSettings.streamConfig,
-                            enabled = currentCameraSettings.concurrentCameraMode ==
-                                ConcurrentCameraMode.OFF
+                            enabled = (
+                                currentCameraSettings.concurrentCameraMode ==
+                                    ConcurrentCameraMode.OFF &&
```
yasith, line 229: Should this be OR?

When ConcurrentCameraMode is OFF, OR, image format is ULTRA HDR?

Kimblebee: The logic up there works... its just phrased terribly woops.
we **don't** want to enable this setting if concurrent mode is `DUAL` OR if image format is `JPEG_ULTRA_HDR`.

I think the structure of this is weird to read. i'll update to :
`!( currentCameraSettings.concurrentCameraMode == ConcurrentCameraMode.DUAL || currentCameraSettings.imageFormat == ImageOutputFormat.JPEG_ULTRA_HDR )`

yasith (approved):

PR #317, opened 2025-02-21
update settings constraints related to video recording quality

Kimblebee:
update disabled rationale text when lens setting is constrained by video quality

video quality selectable states now disabled when unsupported by the default lens

### DIFF
--- a/feature/settings/src/main/java/com/google/jetpackcamera/settings/SettingsViewModel.kt
+++ b/feature/settings/src/main/java/com/google/jetpackcamera/settings/SettingsViewModel.kt
@@ -370,10 +370,10 @@ class SettingsViewModel @Inject constructor(
         cameraAppSettings: CameraAppSettings
     ): VideoQualityUiState {
         val cameraConstraints = systemConstraints.forCurrentLens(cameraAppSettings)
-        val supportedVideoQualities: Set<VideoQuality> =
+        val supportedVideoQualities: List<VideoQuality> =
             cameraConstraints?.supportedVideoQualitiesMap?.get(
                 cameraAppSettings.dynamicRange
-            ) ?: setOf(VideoQuality.UNSPECIFIED)
+            ) ?: listOf(VideoQuality.UNSPECIFIED)
 
         return if (supportedVideoQualities != listOf(VideoQuality.UNSPECIFIED)) {
             VideoQualityUiState.Enabled(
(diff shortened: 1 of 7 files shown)


feature/settings/src/main/java/com/google/jetpackcamera/settings/SettingsViewModel.kt
```diff
@@ -370,15 +370,15 @@ class SettingsViewModel @Inject constructor(
         cameraAppSettings: CameraAppSettings
     ): VideoQualityUiState {
         val cameraConstraints = systemConstraints.forCurrentLens(cameraAppSettings)
-        val supportedVideoQualities: List<VideoQuality>? =
-            cameraConstraints?.supportedVideoQualitiesMap?.get(cameraAppSettings.dynamicRange)
-        return if (!supportedVideoQualities.isNullOrEmpty()) {
+        val supportedVideoQualities: List<VideoQuality> =
```
Jaehwa-Noh, line 373: This would be better to use `Set<VideoQuality>` than `List<VideoQuality>`.

feature/settings/src/main/java/com/google/jetpackcamera/settings/SettingsViewModel.kt
```diff
@@ -407,13 +407,8 @@ class SettingsViewModel @Inject constructor(
...
     private fun getSingleVideoQualityState(
         videoQuality: VideoQuality,
-        supportedVideQualities: List<VideoQuality>?
-    ): SingleSelectableState = if (videoQuality == VideoQuality.UNSPECIFIED ||
-        (
-            !supportedVideQualities.isNullOrEmpty() &&
-                supportedVideQualities.contains(videoQuality)
-            )
-    ) {
+        supportedVideQualities: List<VideoQuality>
```
Jaehwa-Noh, line 410: This would be better to use `Set<VideoQuality>` than `List<VideoQuality>`.

feature/settings/src/main/java/com/google/jetpackcamera/settings/SettingsViewModel.kt
Kimblebee: the function `VideoCapabilities.getSupportedQualities` returns a sorted `List` data structure, so I believe we were just maintaining that.

Though it seems to essentially be a sortedSet, I dont think there should be a problem converting to a set.

yasith (approved):

PR #319, opened 2025-02-27
fix pause while recording bug

Kimblebee:
pausing a pressed recording would cause the ui to treat it as a locked recording

### DIFF
--- a/feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewViewModel.kt
+++ b/feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewViewModel.kt
@@ -420,18 +420,11 @@ class PreviewViewModel @AssistedInject constructor(
                 .Enabled.Idle(captureMode = cameraAppSettings.captureMode)
 
         // display different capture button UI depending on if recording is pressed or locked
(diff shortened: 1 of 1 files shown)


feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewViewModel.kt
```diff
@@ -425,12 +425,14 @@ class PreviewViewModel @AssistedInject constructor(
...
         }
-        // todo: how to handle pause...
         is VideoRecordingState.Active.Paused ->
-            CaptureButtonUiState
-                .Enabled.Recording.LockedRecording
+            if (lockedState) {
+                CaptureButtonUiState
+                    .Enabled.Recording.LockedRecording
+            } else {
+                CaptureButtonUiState.Enabled.Recording.PressedRecording
```
Jaehwa-Noh, line 433: It is the same logic with a line 423.

You just add the case at the line 423
```
 is VideoRecordingState.Active.Recording, is VideoRecordingState.Active.Paused -> if (lockedState) {
            CaptureButtonUiState.Enabled.Recording.LockedRecording
        } else {
            CaptureButtonUiState.Enabled.Recording.PressedRecording
        }
```

yasith (approved):

PR #320, opened 2025-02-27
drag to lock recording

Kimblebee:
Big-ish changes to capture button:
* Capture Button Components moved to its own file
* The visual components (ring, center "nucleus", and lock switch) are now decoupled from each other.
   * The capture button can use a boolean to use either the original capture button center or one with a sliding lock switch. **There is currently no setting for this, so you would need to change the code directly**
* Since the nucleus is now a shared component, **the animations should all be smood like butter 🧈** 
 
Lock Switch feature:
* basically turns the red recording indicator into a sliding switch.
* the lock can be tapped to toggle lock/unlock.
* The switch tray is offset to the left, and will not shift around other components on the preview.
* The nucleus will only traverse the width of the slider.
* The ordering must place the nucleus behind the lock icon, and in front of the switch background

free up a hand :-)
Screenshot 2025-03-21 at 10 13 52 AM

these gifs are from an earlier iteration where the animations weren't as smooth


feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
```diff
@@ -751,153 +846,205 @@ fun CaptureButton(
...
-                                            onLockVideoRecording(true)
-                                            onStartVideoRecording()
+                                CaptureButtonUiState.Enabled.Recording.LockedRecording -> {}
+                            }
+                        },
+                        onTap = {
+                            val uiState = currentUiState.value
+                            when (uiState) {
+                                is CaptureButtonUiState.Enabled.Idle -> {
+                                    if (!isLongPressing) {
```
Jaehwa-Noh, line 932: I wonder that this already has a `onLongPress` lambda(line 889), but why here to check Long press too by variable isLongPressing?
Can't control a condition in `onLongPress`?

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
```diff
@@ -751,153 +846,205 @@ fun CaptureButton(
...
-                                        CaptureMode.STANDARD,
-                                        CaptureMode.IMAGE_ONLY -> onCaptureImage()
-
-                                        CaptureMode.VIDEO_ONLY -> {
-                                            onLockVideoRecording(true)
-                                            onStartVideoRecording()
+                                CaptureButtonUiState.Enabled.Recording.LockedRecording -> {}
+                            }
+                        },
+                        onTap = {
```
Jaehwa-Noh, line 928: All of gestures: onTap, onPress, onLongPress. need to consider?
Can't merge onTap and onPress both into one?

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
```diff
@@ -741,6 +747,95 @@ fun CurrentCameraIdText(physicalCameraId: String?, logicalCameraId: String?) {
...
+                    },
+                    contentDescription = null
+                )
+            }
+        }
+    }
+}
+
 @Composable
 fun CaptureButton(
```
yasith, line 840: I think it'll be helpful if you can break down some of the `Box(..)` composables into @Composable functions with names, that will help improve being able to understand the code.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
Kimblebee: on press is anytime when the finger is touching the composable.  On press always executes, regardless of if the gesture was a tap or long press

on tap only executes is when the composable is pressed and released before the long-press threshold.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
Kimblebee: done, moved to `CaptureButtonComponents.kt`

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/PreviewScreenComponents.kt
```diff
@@ -741,6 +747,95 @@ fun CurrentCameraIdText(physicalCameraId: String?, logicalCameraId: String?) {
     }
 }
 
+// todo(kc): clean up animations
+@Composable
+private fun DraggableLockSwitch(
```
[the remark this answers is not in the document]
Kimblebee: done, moved to `CaptureButtonComponents.kt`

yasith (approved):

PR #321, opened 2025-02-28
implement drag to zoom

Kimblebee:
You can now drag to zoom in/out while performing a pressed recording.
I think i could work on improving the tuning of the increment.

The drag zoom only works when **all** of these conditions are met:
* the recording is a pressed recording
* The gesture's `y` value is above the capture button component

Drag to Lock will only work while the y coordinate of the gesture is at or below the top of the capture button.
Dragging above the capture button will snap the red "nucleus" to locked or unlocked depending on its position when the gesture moved above the  capture button.


feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/CaptureButtonComponents.kt
```diff
@@ -315,18 +356,41 @@ private fun CaptureButton(
...
+                                    newPoint.y
+                                } else {
+                                    dragOffset.y
+                                }
+                            dragOffset = newPoint
+
+                            if (!positiveDistance.isNaN()) {
+                                // todo(kc): should improve the tuning of this.
+                                //  todo(): zoom should always return to the original when you drag
+                                //  back to 0 on the y axis
```
temcguir, line 388: How much work would this be to do? Is it something that can be addressed immediately after this is merged?

temcguir (approved):

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/CaptureButtonComponents.kt
```diff
@@ -315,18 +356,41 @@ private fun CaptureButton(
...
+                    onDrag = { change, offset ->
+                        // update position of lock switch
+                        Log.d(TAG, "move y: ${offset.y}")
+                        setLockSwitchPosition(offset.x)
+
+                        // update zoom
+                        if (currentUiState.value ==
+                            CaptureButtonUiState.Enabled.Recording.PressedRecording
+                        ) {
+                            val newPoint = dragOffset + offset
```
yasith, line 371: Could you explain the logic behind newPoint = dragOffset + offset?

Kimblebee: sorry, the names here are kind of confusing.

the value of `dragOffset` is maintained such that it basically tracks the location of the finger performing the drag gesture -- _a coordinate **relative to the origin** of the capture button component._

The offset that we receive from onDrag is just the delta from the previous gesture. With that offset alone, we have no idea **where** the gesture is, just that it moved by some amount.

at this point, `newPoint` is the current location of the gesture, and `dragOffset` is the previous.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/CaptureButtonComponents.kt
Kimblebee: My **extremely** short answer to this is no, I don't think that this is something that can be immediately addressed.
Extra complexities would be introduced... especially when we flip the camera while recording. I think we'd need to track our current zoom and the zoom ratio of the lens, we'd also  need some way to know when we reached the max zoom of the camera lens. Linear zoom can't be used because our settings wouldn't be 100% prescriptive.

 To be honest, I am taking a closer look at this feature on other apps, and will likely renege on that second `todo()`. 

For the tuning bit, i think that is a mostly subjective thought. I would want to feel the gesture on a variety of devices. I'd probably want to dot he same with other apps that implement a similar feature too.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/CaptureButtonComponents.kt
temcguir: Its this value captured by [change.position](https://developer.android.com/reference/kotlin/androidx/compose/ui/input/pointer/PointerInputChange#position())?

Kimblebee: oops

..fixed 😅

PR #324, opened 2025-03-04
quick setting to switch between capture modes.

Kimblebee:
Squashed changes from #303 but preserved the HDR toggle button

Feature:
- Adds a quick setting to switch between capture modes. Disabled if a feature only supports one capture mode. Individual options are disabled based on current settings and constraints.
- HDR Toggle refactored to use `CaptureModeUiState` instead of previous`CaptureModeToggleUiState`
- **HDR Toggle now toggles the capture modes instead of imageOutputFormat .**
- Disabling HDR resets capture mode to standard

Testing:
- adds tests to make sure other settings interact properly with Capture Mode, HDR settings, and the HDR capture mode toggle
- adds some new `ComposeTestRule` extension functions to simplify writing tests.

screen-20250304-112443 screen-20250304-112443


feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/quicksettings/ui/QuickSettingsComponents.kt
```diff
@@ -114,14 +117,107 @@ fun FocusedQuickSetRatio(
...
+        }
+    ExpandedQuickSetting(modifier = modifier, quickSettingButtons = buttons)
+}
+
+@Composable
+fun QuickSetCaptureMode(
+    modifier: Modifier = Modifier,
+    onClick: () -> Unit,
+    captureModeUiState: CaptureModeUiState,
+    assignedCaptureMode: CaptureMode?,
```
davidjiagoogle, line 171: Does this have to be nullable?

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/quicksettings/QuickSettingsScreen.kt
```diff
@@ -265,12 +278,14 @@ private fun ExpandedQuickSettingsUi(
                             selectedImageOutputFormat = currentCameraSettings.imageFormat,
                             hdrDynamicRangeSupported =
                             cameraConstraints?.hdrDynamicRangeSupported() == true,
-                            previewMode = previewUiState.previewMode,
```
davidjiagoogle, line 268: I noticed previewMode checks are removed here and in the components and cameraConstraint check is added instead.

I don't think this can be done. In PreviewModes other than StandardMode, one should not be able to change Capture Mode. ExternalImageCaptureMode & ExternalMultipleImageCaptureMode should lock the CaptureMode into ImageOnlyCapture, and ExternalVideoCaptureMode VideoOnly

Kimblebee: ah ok i see. I automatically set the capture mode based on the preview mode, but I forgot that I should also prevent the user from being able to change it. I can add previewmode as a disabling consideration for their UiStates

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/quicksettings/ui/QuickSettingsComponents.kt
Kimblebee: yes it does. 

this quick setting is focusable, much like the aspect ratio quick setting.
This same composable is reused for the button that opens the options, and each of the individual options.

when the assigned capture mode is null, then this setting will simply display the value currently used by the settings.
When the capture mode is assigned, then the button it will display the UI for that specific capture mode

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/quicksettings/ui/QuickSettingsComponents.kt
```diff
@@ -114,14 +117,107 @@ fun FocusedQuickSetRatio(
...
 }
 
+@Composable
+fun FocusedQuickSetCaptureMode(
+    modifier: Modifier = Modifier,
+    onSetCaptureMode: (CaptureMode) -> Unit,
+    captureModeUiState: CaptureModeUiState
+) {
+    val buttons: Array<@Composable () -> Unit> =
+        if (captureModeUiState is CaptureModeUiState.Enabled) {
```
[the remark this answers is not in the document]
Kimblebee: i could but i found that using a similar auxiliary function to create composables can sometimes cause the test tags not to be applied properly. Had that issue with another component a while ago.

 for now I'll just sit on the safe side and leave it as this.

davidjiagoogle (approved):

PR #328, opened 2025-03-12
Implement LocalMediaRepository

yasith:
Apply SpotlessApply Spotless


data/media/src/main/kotlin/com/google/jetpackcamera/data/media/LocalMediaRepository.kt
```diff
@@ -15,12 +15,153 @@
...
+                // Android 10 (API 29) and above: Use ImageDecoder
+                context.contentResolver.openInputStream(uri)?.use { inputStream ->
+                    BitmapFactory.decodeStream(inputStream)
+                }
+            } else {
+                // Android 9 (API 28) and below: Use BitmapFactory
+                context.contentResolver.openInputStream(uri)?.use { inputStream ->
+                    BitmapFactory.decodeStream(inputStream)
+                }
+            }
```
Jaehwa-Noh, line 80: What differs between two OS versions? It is as though same logic.

data/media/src/main/kotlin/com/google/jetpackcamera/data/media/LocalMediaRepository.kt
```diff
@@ -15,12 +15,153 @@
...
+                            null
+                        )
+                    } else { // Video
+                        MediaStore.Video.Thumbnails.getThumbnail(
+                            context.contentResolver,
+                            ContentUris.parseId(uri),
+                            MediaStore.Video.Thumbnails.MINI_KIND,
+                            null
+                        )
+                    }
```
Jaehwa-Noh, line 123: What about moving if state in the parameter level?

```kotlin 
MediaStore.Video.Thumbnails.getThumbnail(
                            context.contentResolver,
                            ContentUris.parseId(uri),
                            if (collectionUri == MediaStore.Images.Media.EXTERNAL_CONTENT_URI)  MediaStore.Images.Thumbnails.MINI_KIND
                            else MediaStore.Video.Thumbnails.MINI_KIND,
                            null
                        )
```

feature/postcapture/build.gradle.kts
```diff
@@ -129,5 +135,11 @@ dependencies {
...
+    implementation(project(":data:media"))
     testImplementation(project(":core:common"))
-}
\ No newline at end of file
+}
+
+// Allow references to generated code
+kapt {
+    correctErrorTypes = true
+}
```
Jaehwa-Noh, line 145: KAPT is in the maintenance mode.
https://kotlinlang.org/docs/kapt.html

Does dagger hilt compiler have to use `kapt`?

yasith: KSP still doesn't support Hilt. Right now there's no reason for us to migrate.

data/media/src/main/kotlin/com/google/jetpackcamera/data/media/LocalMediaRepository.kt
yasith: Thank you, good catch! I'll fix it

feature/postcapture/build.gradle.kts
Jaehwa-Noh: I've seen using ksp to hilt in here.
https://github.com/android/nowinandroid/blob/af33aca69956188ffc985cf81388453706b0d6e9/build-logic/convention/src/main/kotlin/HiltConventionPlugin.kt#L30

and I found the content about ksp in hilt release note.
https://developer.android.com/jetpack/androidx/releases/hilt#hilt_version_110_2

If you have any chance to migrate it, I suggest you could try it.

yasith: Looks like the support is still in alpha. We'll keep an eye out for that to become stable.

Kimblebee (approved):

PR #330, opened 2025-03-12
Use MediaRepository in PostCaptureScreen and the ImageWell

yasith:


feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewViewModel.kt
```diff
@@ -745,9 +751,7 @@ class PreviewViewModel @AssistedInject constructor(
                     }, contentResolver, finalImageUri, ignoreUri).savedUri
                 },
                 onSuccess = { savedUri ->
-                    savedUri?.let {
-                        updateLastCapturedImageUri(it)
-                    }
+                    updateLastCapturedMedia()
```
Kimblebee, line 754: I think this is fine for now, but I think an improvement might be to use a `ContentObserver` for new file URI's that match our search criteria? TBH I haven't had a chance to look into it much yet, so it might not work.

Kimblebee: I'd also like to emphasize that this is just a thought I had, and is probably unnecessary within the scope of JCA.

However, a change like this may be applicable should we ever want to view images taken by other applications.

feature/postcapture/src/main/java/com/google/jetpackcamera/feature/postcapture/PostCaptureViewModel.kt
```diff
@@ -16,31 +16,52 @@
...
+        }
     }
 
-    fun deleteImage(contentResolver: ContentResolver) {
-        contentResolver.delete(uiState.value.imageUri!!, null, null)
-        _uiState.update { it.copy(imageUri = null, isImageDeleted = true) }
+    fun deleteMedia(contentResolver: ContentResolver) {
+        when (val mediaDescriptor = uiState.value.mediaDescriptor) {
+            is MediaDescriptor.Image -> contentResolver.delete(mediaDescriptor.uri, null, null)
+            is MediaDescriptor.Video -> contentResolver.delete(mediaDescriptor.uri, null, null)
```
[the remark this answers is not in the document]
yasith: You can't combine these since they can't be smart casted to have a uri.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/ui/ImageWell.kt
```diff
@@ -31,22 +30,23 @@ import androidx.compose.ui.draw.clip
...
-    val context = LocalContext.current
-
     when (imageWellUiState) {
         is ImageWellUiState.LastCapture -> {
-            val bitmap = loadAndRotateBitmap(context, imageWellUiState.uri, 270f)
+            val bitmap = when (imageWellUiState.mediaDescriptor) {
+                is MediaDescriptor.Image -> imageWellUiState.mediaDescriptor.thumbnail
+                is MediaDescriptor.Video -> imageWellUiState.mediaDescriptor.thumbnail
+                is MediaDescriptor.None -> null
+            }
```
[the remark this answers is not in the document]
yasith: Same here, you can't combine these.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewViewModel.kt
yasith: I think with the new changes we're planning we'll probably write media into the `MediaRepository` and the need to update last captured media should go away on it's own.

It's a bit asymmetric right now. We write directly to `MediaStore` but read from `MediaRepository`, ideally we should be going through `MediaRepository` for both read/write.

feature/postcapture/src/main/java/com/google/jetpackcamera/feature/postcapture/PostCaptureScreen.kt
```diff
@@ -50,25 +50,26 @@ import androidx.compose.ui.graphics.nativeCanvas
...
+fun PostCaptureScreen(viewModel: PostCaptureViewModel = hiltViewModel()) {
+    Log.d(TAG, "PostCaptureScreen")
+
     val uiState: PostCaptureUiState by viewModel.uiState.collectAsState()
     val context = LocalContext.current
 
-    LaunchedEffect(imageUri) {
-        viewModel.setLastCapturedImageUri(imageUri)
+    LaunchedEffect(Unit) {
+        viewModel.getLastCapture()
```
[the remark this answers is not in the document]
yasith: Yes, moved it to the init block of the ViewModel

feature/postcapture/src/main/java/com/google/jetpackcamera/feature/postcapture/PostCaptureScreen.kt
```diff
@@ -90,10 +87,25 @@ fun PostCaptureScreen(viewModel: PostCaptureViewModel = hiltViewModel(), imageUr
                     }
                 }
             }
-        } ?: Text(
-            text = "No Image Captured",
-            modifier = Modifier.align(Alignment.Center)
-        )
+            is Media.Video -> {
+                Text(
+                    text = "Video support pending",
```
Jaehwa-Noh, line 92: Why don't you use string resource at this point? and how about leave a TODO at here?

yasith: It's immediately removed in #331, so I'm not too worried about leaving it here.

Kimblebee (approved):

PR #332, opened 2025-03-20
Turn SnackBar PreviewUiState into a FIFO queue

davidjiagoogle:
https://github.com/user-attachments/assets/e480c24a-eb01-441a-8b43-043ef2638f93

Screenshot 2025-03-20 at 3 00 50 PM


feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewViewModel.kt
```diff
@@ -673,21 +674,30 @@ class PreviewViewModel @AssistedInject constructor(
...
-                        withDismissAction = true,
-                        testTag = IMAGE_CAPTURE_EXTERNAL_UNSUPPORTED_TAG
-                    )
+                    snackBarsToShow = newQueue
                 ) ?: old
             }
         }
     }
 
+    private fun showExternalVideoCaptureUnsupportedToast() {
```
Kimblebee, line 690: should rename this since we are using a snackbar for this, not a toast.

probably something like `enqueueExternalVideoCaptureUnsupportedSnackbar` or `Alert` instead of `SnackBar` to shorten it.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewViewModel.kt
```diff
@@ -800,30 +806,21 @@ class PreviewViewModel @AssistedInject constructor(
...
-                (old as? PreviewUiState.Ready)?.copy(
-                    // todo: remove snackBar after postcapture screen implemented
-                    snackBarToShow = snackBarData
-                ) ?: old
-            }
+            addSnackBarData(snackBarData)
         }
     }
 
     fun showSnackBarForDisabledHdrToggle(disabledReason: CaptureModeToggleUiState.DisabledReason) {
```
Kimblebee, line 813: nit: again, can rename this since we are queueing and not necessarily displaying immediately anymore.

could be something like `enqueueDisabledHdrToggleSnackbar`

Kimblebee (approved):

PR #336, opened 2025-04-07
Add optimization flags in gradle.properties file

### LINKED ISSUE #260: Turn on common gradle optimization flags
Add all (or some) of these options to the `gradle.properties` file. 

```
# https://docs.gradle.org/current/userguide/build_cache.html
org.gradle.caching=true

# https://docs.gradle.org/current/userguide/performance.html#parallel_execution
org.gradle.parallel=true

# https://docs.gradle.org/current/userguide/gradle_daemon.html#sec:daemon_watch_fs
org.gradle.vfs.watch=true

# https://docs.gradle.org/current/userguide/configuration_cache.html
org.gradle.configuration-cache=true
```

See nowinandroid for reference: 
https://github.com/android/nowinandroid/blob/main/gradle.properties

henni99:
resolve #260


henni99:
@Kimblebee Are you planning to review it?

gradle.properties
```diff
@@ -20,6 +20,22 @@ android.suppressUnsupportedOptionWarnings=android.suppressUnsupportedOptionWarni
...
+
+# When configured, Gradle will run in incubating parallel mode.
+# This option should only be used with decoupled projects. More details, visit
+# http://www.gradle.org/docs/current/userguide/multi_project_builds.html#sec:decoupled_projects
+org.gradle.parallel=true
+
+# Enable caching between builds.
+org.gradle.caching=true
+
+# enables file system watching by default for supported operating systems
```
temcguir, line 32: This is actually unnecessary since we're using Gradle 8. It's been enabled by default since Gradle 7:
https://docs.gradle.org/current/userguide/file_system_watching.html#enable

temcguir (commented):
Sorry for the delay on reviewing this. I needed to allow this to run presubmit (fixed with #350). I left a few comments.

temcguir (changes requested):
Sorry forgot to mark this as request changes.

henni99:
@temcguir I'll try it by the end of this week Thank you 😊

gradle.properties
henni99: According to the link, the file system watching feature has been provided by default since Gradle 7, so I removed the corresponding property!

Was it your intention to say that only the file system watching property is necessary — not caching or parallel?

temcguir: Yeah that's right. Thanks for the change!

temcguir:
Thanks for the contribution, @henni99!

PR #347, opened 2025-04-24
Update SettingsDeviceTest to select a setting

temcguir:
Adds more tests to SettingsDeviceTest for all of the available
 settings.

 Each test now selects a setting that was not previously selected.

 The test is now parameterized to use both the front and back camera.


davidjiagoogle:
All tests are giving ComposeTimeoutException right now

feature/settings/src/main/res/values/strings.xml
```diff
@@ -28,7 +28,7 @@
     <string name="section_title_software_info">Software Information</string>
 
     <string name="default_facing_camera_title">Default to Front Camera</string>
-    <string name="default_facing_camera_description">Default Front</string>
+    <string name="default_facing_camera_description_on">Default Front</string>
```
Kimblebee: nit: should we just change the name of the string resource to `default_facing_camera_description_front` and
`default_facing_camera_description_rear `

feature/settings/src/main/java/com/google/jetpackcamera/settings/ui/TestTags.kt
```diff
@@ -25,7 +25,9 @@ package com.google.jetpackcamera.settings.ui
 //
 // ////////////////////////////////
 
+const val SETTINGS_TITLE = "SettingsTitle"
 const val BACK_BUTTON = "BackButton"
+const val CLOSE_BUTTON = "CloseButton"
```
Kimblebee, line 30: nit: i think these tags could be a little more descriptive, avoid any confusion regarding the tag's purpose.

something like: `SettingsScreenTitle` `CloseSettingDialogButton` and `NavExitSettings`. 

changing back button now might break any tests that use it though.

app/src/androidTest/java/com/google/jetpackcamera/SettingsDeviceTest.kt
```diff
@@ -15,122 +15,281 @@
...
+import com.google.jetpackcamera.utils.selectLensFacing
+import com.google.jetpackcamera.utils.visitSettingDialog
+import com.google.jetpackcamera.utils.visitSettingsScreen
 import org.junit.Rule
 import org.junit.Test
 import org.junit.runner.RunWith
+import org.junit.runners.Parameterized
 
-private const val TAG = "SettingsDeviceTest"
+@RunWith(Parameterized::class)
```
Kimblebee, line 86: nit: cool use of parameterized test class. might be helpful to include a comment to give readers who may be unfamiliar  a little more context as to what is occurring  in the companion object; that the `Parameterized` runner is injecting values from data() into the constructor.

Kimblebee (approved):
looks good, just a few nits for improving readability

feature/settings/src/main/java/com/google/jetpackcamera/settings/ui/TestTags.kt
temcguir: I think I'm OK with keeping this as-is in this case. I don't see a reason we couldn't use this same test tag in multiple places in settings. If there were multiple close buttons on the screen at the same time, then it might make sense to start getting more specific with the tag names, but when this is used in the context of a setting dialog, there will only be 1 close button, so I don't think we need to get more specific.

PR #358, opened 2025-05-16
update dependencies and revert AGP version to 8.10.0

Kimblebee:


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
 #Tue Mar 12 23:44:57 PDT 2024
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-8.13-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.9-bin.zip
```
temcguir, line 4: Based on the error message, looks like this needs to be `gradle-8.11.1`

temcguir (changes requested):

temcguir (approved):

PR #359, opened 2025-06-01
Remove a duplicated extension function.

Jaehwa-Noh:
This `tryApplyVideoQualityConstraints` function was added in #324 and it was applied twice.
See line 274 and 276
https://github.com/google/jetpack-camera-app/blob/668eac5183f545eb8d2a94c60e76787d284e6c23/core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraUseCase.kt#L274

https://github.com/google/jetpack-camera-app/blob/668eac5183f545eb8d2a94c60e76787d284e6c23/core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraUseCase.kt#L276


temcguir:
Thank you @Jaehwa-Noh!

PR #361, opened 2025-06-16
Gllb

donovanfm:


core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraSession.kt
```diff
@@ -364,30 +423,55 @@ internal suspend fun processTransientSettingEvents(
...
     camera: Camera,
     prevTransientSettings: TransientSessionSettings?,
     newTransientSettings: TransientSessionSettings,
+    sessionSettings: PerpetualSessionSettings.SingleCamera?,
     optionsBuilder: CaptureRequestOptions.Builder
 ) {
     var needsUpdate = false
+
     if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.VANILLA_ICE_CREAM &&
         prevTransientSettings?.flashMode != newTransientSettings.flashMode
```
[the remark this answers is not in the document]
donovanfm: This comment is not correct. See the reference docs where it shows this enum was added in API 35. https://developer.android.com/reference/android/hardware/camera2/CameraMetadata#CONTROL_AE_MODE_ON_LOW_LIGHT_BOOST_BRIGHTNESS_PRIORITY

.idea/other.xml
temcguir, line 1: nit: does this need to be deleted?

core/camera/src/main/java/com/google/jetpackcamera/core/camera/effects/LowLightBoostSessionContainer.kt
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    companion object {
+        private var instance: LowLightBoostSessionContainer? = null
+
+        fun getInstance(): LowLightBoostSessionContainer = instance ?: synchronized(this) {
+            instance ?: LowLightBoostSessionContainer().also { instance = it }
+        }
+    }
+}
```
temcguir: The slightly more idiomatic way to do this is to just make it an `object` rather than using a `companion object`:
```
@RequiresApi(Build.VERSION_CODES.R)
object LowLightBoostSessionContainer {
    var lowLightBoostSession: LowLightBoostSession? = null

    fun releaseSession() {
        lowLightBoostSession?.release()
        lowLightBoostSession = null
    }
}
```

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraExt.kt
```diff
@@ -135,6 +144,42 @@ val CameraInfo.isOpticalStabilizationSupported: Boolean
         ?.contains(
             CameraMetadata.LENS_OPTICAL_STABILIZATION_MODE_ON
         ) ?: false
+
+@RequiresApi(Build.VERSION_CODES.R)
```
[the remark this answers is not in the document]
donovanfm: Addressed in [this commit](https://github.com/google/jetpack-camera-app/pull/361/commits/ad6477a3c5fe3063c268a6abeb75dee49f0c761e). All LLB AE Mode code is now checking for VIC, and Google LLB is checking for R. (Gemini is confused about which SDK versions are needed for each LLB option, but the linked commit puts to rest any SDK version checking issues.)

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraUseCase.kt
```diff
@@ -130,6 +133,7 @@ constructor(
 
     override fun getSurfaceRequest(): StateFlow<SurfaceRequest?> = _surfaceRequest.asStateFlow()
 
+    @RequiresApi(Build.VERSION_CODES.R)
```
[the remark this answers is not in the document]
donovanfm: Addressed in [this commit](https://github.com/google/jetpack-camera-app/pull/361/commits/ad6477a3c5fe3063c268a6abeb75dee49f0c761e). All LLB AE Mode code is now checking for VIC, and Google LLB is checking for R. (Gemini is confused about which SDK versions are needed for each LLB option, but the linked commit puts to rest any SDK version checking issues.)

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraSession.kt
```diff
@@ -117,7 +123,11 @@
     SD to Range.create(241, 719)
 )
 
+lateinit var lowLightBoostSessionContainer: LowLightBoostSessionContainer
+
 context(CameraSessionContext)
+@ExperimentalCamera2Interop
+@RequiresApi(Build.VERSION_CODES.R)
```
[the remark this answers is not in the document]
donovanfm: Addressed in [this commit](https://github.com/google/jetpack-camera-app/pull/361/commits/ad6477a3c5fe3063c268a6abeb75dee49f0c761e). All LLB AE Mode code is now checking for VIC, and Google LLB is checking for R. (Gemini is confused about which SDK versions are needed for each LLB option, but the linked commit puts to rest any SDK version checking issues.)

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraSession.kt
```diff
@@ -118,7 +124,11 @@ private val QUALITY_RANGE_MAP = mapOf(
     SD to Range.create(241, 719)
 )
 
+lateinit var lowLightBoostSessionContainer: LowLightBoostSessionContainer
```
temcguir: If this is a singleton, there's no reason to store it in an instance variable.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraSession.kt
```diff
@@ -157,11 +169,53 @@ internal suspend fun runSingleCameraSession(
...
             cameraProvider.unbindAll()
+            lowLightBoostSessionContainer.releaseSession()
             val currentCameraSelector = currentTransientSettings.primaryLensFacing
                 .toCameraSelector()
+            val cameraInfo = cameraProvider.getCameraInfo(currentCameraSelector)
+            val camera2Info = Camera2CameraInfo.from(cameraInfo)
+            val cameraId = camera2Info.cameraId
+            val lowLightBoostAvailability = cameraInfo.getLowLightBoostAvailablity(context)
+
+            val cameraEffects = mutableListOf<CameraEffect>()
```
temcguir: I'm a little confused why we need a list of effects here? Effects are usually just one or the other, you can't stack effects on top of each other. The only reason the `UseCaseGroup.Builder.addEffect()` method isn't called `setEffect` is because you can apply different effects to different streams of the camera.

Rather than creating a list, we can just create the effect right before we add it in `runSingleCameraSession`. You can just edit the existing `when` clause so that it will choose the LowLightBoostEffect when it's needed, and fall back to checking the MULTI_STREAM/SINGLE_STREAM StreamConfig otherwise.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraSystem.kt
```diff
@@ -219,8 +224,13 @@ constructor(
                                 add(Illuminant.SCREEN)
                             }
 
-                            if (camInfo.isLowLightBoostSupported) {
-                                add(Illuminant.LOW_LIGHT_BOOST)
+                            coroutineScope {
```
temcguir: I think this `coroutineScope` is unnecessary.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/effects/LowLightBoostSessionContainer.kt
```diff
@@ -0,0 +1,38 @@
...
+
+    fun releaseSession() {
+        lowLightBoostSession?.release()
+        lowLightBoostSession = null
+    }
+
+    companion object {
+        private var instance: LowLightBoostSessionContainer? = null
+
+        fun getInstance(): LowLightBoostSessionContainer = instance ?: synchronized(this) {
```
temcguir: I would prefer if we can stay away from using a singleton here. I feel like we should be able to manage LowLightBoostSession lifecycles with the coroutine scopes that they are created in. Maybe that could be addressed in a follow-up PR.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraSystem.kt
```diff
@@ -219,8 +224,13 @@ constructor(
...
+                                if (camInfo.getLowLightBoostAvailablity(
+                                        application
+                                    ) != LowLightBoostAvailability.NONE
+                                ) {
+                                    add(Illuminant.LOW_LIGHT_BOOST)
```
temcguir: The point of this `Illuminant` is to differentiate between different ways to brighten the screen within a `FlashMode`. So we could change `Illuminant.LOW_LIGHT_BOOST` to have 2 separate entries in `Illuminant`: `HARDWARE_LOW_LIGHT_BOOST` and `SOFTWARE_LOW_LIGHT_BOOST`. With that, it becomes easier to select a specific illuminant up front, so we don't need to keep checking availability down within the camera session. When the flash mode switches to LOW_LIGHT_BOOST, we can check for available LLB illuminants, check which one we want prioritized, and set the illuminant on the camera settings alongside the flash mode. Then, for instance, in `CameraSession.updateCamera2RequestOptions`, all we need to do is check if the Illuminant is `HARDWARE_LOW_LIGHT_BOOST` (which was determined when the flash mode was set to LLB) before setting the AE mode.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/effects/processors/LowLightBoostSurfaceProcessor.kt
```diff
@@ -0,0 +1,198 @@
...
+            inputSurfaceRequestScope = null
+            releaseLowLightBoostSession()
+        }
+    }
+
+    @RequiresApi(Build.VERSION_CODES.R)
+    private suspend fun tryCreateLowLightBoostSession(outputSurfaceForLlb: Surface) {
+        inputSurfaceRequestScope?.withSurfaceRequest { currentInputRequest ->
+
+            coroutineScope.launch { // Launch in your processor's scope
```
temcguir: I don't think this launch is necessary. This code is already within a private suspend fun that is called within a flow collector. It will get cancelled if a new `SurfaceOutput` comes in, and is already within the correct scope.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/effects/processors/LowLightBoostSurfaceProcessor.kt
```diff
@@ -0,0 +1,198 @@
...
+
+                    Log.d(TAG, "LLB Session created. Providing LLB input surface to CameraX.")
+
+                    // Fulfill the CameraX SurfaceRequest with LLB's input surface
+                    currentInputRequest.provideSurface(llbInputSurface, glExecutor) { result ->
+                        Log.d(TAG, "CameraX SurfaceRequest result: ${result.resultCode}")
+                        llbInputSurface.release()
+                        when (result.resultCode) {
+                            SurfaceRequest.Result.RESULT_SURFACE_USED_SUCCESSFULLY -> {
+                                Log.i(TAG, "CameraX is using LLB input surface.")
```
temcguir: The result is sent _after_ CameraX is done using the surface. I believe we should be releasing the LLB session in every case here, no matter the result.

feature/settings/src/main/java/com/google/jetpackcamera/settings/ui/SettingsComponents.kt
```diff
@@ -389,6 +391,64 @@ fun StreamConfigSetting(
...
+                LowLightBoostPriority.PRIORITIZE_AE_MODE -> stringResource(
+                    id = R.string.low_light_boost_priority_description_ae_mode
+                )
+
+                LowLightBoostPriority.PRIORITIZE_GOOGLE_PLAY_SERVICES -> stringResource(
+                    id = R.string.low_light_boost_priority_description_google_play_services
+                )
+            }
+        } else {
+            TODO("low light boost priority currently has no disabled criteria")
```
temcguir, line 416: As mentioned in another comment, we should add different Illuminants for the 2 different flavors of LLB. If only 1 or neither of the LLB illuminants are available, this setting should be disabled. When only 1 is available, there is no reason to present the option to the user, which is why it will also be disabled in that case.

.idea/other.xml
donovanfm: I restored this file.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraSystem.kt
donovanfm: Done

feature/settings/src/main/java/com/google/jetpackcamera/settings/ui/SettingsComponents.kt
donovanfm: Done

core/camera/src/main/java/com/google/jetpackcamera/core/camera/effects/LowLightBoostSessionContainer.kt
donovanfm: Done

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraSession.kt
donovanfm: Done

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraSystem.kt
donovanfm: Done

core/camera/src/main/java/com/google/jetpackcamera/core/camera/effects/processors/LowLightBoostSurfaceProcessor.kt
donovanfm: Done

core/camera/src/main/java/com/google/jetpackcamera/core/camera/effects/processors/LowLightBoostSurfaceProcessor.kt
donovanfm: I think in that case, release is called in the onSessionDisconnected callback. Until version beta06 is available, calling release multiple times will result in an error being thrown. I tried adding `releaseLowLightBoostSession()` to both of the cases where it's not currently being called, and in both circumstances, JCA would crash.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraSession.kt
donovanfm: Good call. I've changed this back to be a single effect.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/effects/LowLightBoostSessionContainer.kt
donovanfm: Acknowledged. I'll look into this while this PR is under review if there's time. Otherwise I'll address this in a follow-up PR.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraSession.kt
```diff
@@ -117,7 +123,11 @@
     SD to Range.create(241, 719)
 )
 
+lateinit var lowLightBoostSessionContainer: LowLightBoostSessionContainer
```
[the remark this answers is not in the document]
donovanfm: This was resolved by a different comment from Trevor.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraSystem.kt
```diff
@@ -107,6 +108,8 @@ interface CameraSystem {
 
     suspend fun setLowLightBoostPriority(lowLightBoostPriority: LowLightBoostPriority)
 
+    fun getLowLightBoostEvents(): MutableSharedFlow<LowLightBoostEvent>
```
jsaund: nit: while I notice the convention here has been to use functions, would it seems more idiomatic to use val in this instance?
eg. `val lowLightBoostEvents: StateFlow<LowLightBoostEvent>`
And then in your impl you can do

`val lowLightBoostEvents = MutableSharedFlow<LowLightBoostEvent>()` ?

Also it would be best to expose a `StateFlow` instead of `MutableSharedFlow` so the consumer endpoint of this is only consuming events and not writing. I'm assuming you want to restrict writing events to happen within the impl of the class.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/effects/processors/LowLightBoostSurfaceProcessor.kt
donovanfm: I'm now calling releaseLowLightBoostSession() right after llbInputSurface.release(), so it's being called regardless of the result status.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraSystem.kt
donovanfm: I can change it to a SharedFlow for the consumers, but as for fun/val, I'll leave that up to Trevor since I was just matching how other fields are being handled.

donovanfm: Updated `getLowLightBoostEvents()` to return a `SharedFlow<LowLightBoostEvent>` in https://github.com/google/jetpack-camera-app/pull/361/commits/4a3c87d7edf3a1cbd8e5f42bed4b79fb7af9b506. Thanks for the suggestion! :)

temcguir: > nit: while I notice the convention here has been to use functions, would it seems more idiomatic to use val in this instance?

Yes, I agree it should use properties here, but because of the current convention, let's leave it as a function. We can do the switch to properties all at once.

As for the use of `SharedFlow`, we generally try to avoid it (see go/coroutine-sharedflows-bad-defaults if you're interested in why). We generally try to stick to `StateFlow` and `Channel`. We use `StateFlow` when it's ok for events to be dropped and `Channel` when we want to handle each event (best effort). I'll take a look at the current code and see if I can suggest how we can change it.

donovanfm: Ack on the use of properties.

For `ShareFlow` alternatives, I didn't use a `Channel` because I'm listening to the event in 2 places, and I wanted to keep it open if we needed to add a subscriber somewhere else too. I didn't use a `StateFlow` because this is just a way to trigger events; there's not a persisted state. I just basically need a pub/sub mechanism, so whatever the recommended approach for that is fine.

temcguir: After thinking about this some more, I think it makes sense to just put the failure into `LowLightBoostState` rather than having a separate event type and callback. I created https://github.com/donovanfm/jetpack-camera-app/pull/2 to achieve this. This allows us to avoid adding a new channel and we can just use the existing CameraState. Let me know what you think.

donovanfm: Trevor's PR has been merged and addresses the `SharedFlow` conversation. Since the val/fun conversation needs to be a refactor, I'll go ahead and mark this conversation as resolved.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/effects/LowLightBoostSessionContainer.kt
donovanfm: This was addressed some time ago. Closing this comment now.

temcguir (approved):

PR #367, opened 2025-06-29
Replace modifier `size` to `height`, improving code readability.

Jaehwa-Noh:
This `size` modifier was added in #335.
Actually, the width is controlled by an `RowScope().weight(1f)` modifier, so that `size` modifier doesn't make a width change.
That was reason, I replaced the `size` modifier to `height`.


temcguir (commented):
Thanks for the contribution!

temcguir (approved):

PR #375, opened 2025-07-14
Change default saving location for image & video capture to DCIM/Camera

davidjiagoogle:
As titled.


davidjiagoogle:
CameraXUseCaseTest.canRecordVideo() is failing on main as well. Not caused by this PR

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraCoreUtil.kt
```diff
@@ -28,8 +28,12 @@ import java.io.FileOutputStream
 import org.json.JSONArray
 import org.json.JSONObject
 
-private const val TAG = "DebugCameraInfoUtil"
-object DebugCameraInfoUtil {
+private const val TAG = "CameraCoreUtil"
+object CameraCoreUtil {
+    fun getDefaultMediaSaveLocation(): String {
+        return Environment.DIRECTORY_DCIM + File.separator + "Camera"
```
temcguir, line 34: nit: this is probably fine for now, but `DCIM/Camera` does not follow the [digital camera filesystem (DCF) standard](https://en.wikipedia.org/wiki/Design_rule_for_Camera_File_system#Directory_and_file_structure). There are many devices that use it, but there are also others that use something like `DCIM/123ABCDE`. In the future we may need to make the sub-folder under `DCIM` customizable.

Issue #345, opened 2025-04-23
kotlin-reflect causing crash when changing settings on AOSP


temcguir:
On the AOSP version of the app, whenever a setting is changed the app crashes. This seems to be due to the use of kotlin reflection:
```
04-22 15:27:18.362 15369 15369 E AndroidRuntime: FATAL EXCEPTION: main
04-22 15:27:18.362 15369 15369 E AndroidRuntime: Process: com.google.jetpackcamera, PID: 15369
04-22 15:27:18.362 15369 15369 E AndroidRuntime: kotlin.reflect.full.IllegalCallableAccessException: java.lang.IllegalAccessException: Class java.lang.Class<kotlin.reflect.jvm.internal.calls.CallerImpl$FieldGetter> cannot access private final  field com.google.jetpackcamera.settings.model.AspectRatio com.google.jetpackcamera.settings.model.CameraAppSettings.aspectRatio of class java.lang.Class<com.google.jetpackcamera.settings.model.CameraAppSettings>
04-22 15:27:18.362 15369 15369 E AndroidRuntime: 	at kotlin.reflect.jvm.internal.KCallableImpl.call(KCallableImpl.kt:280)
04-22 15:27:18.362 15369 15369 E AndroidRuntime: 	at kotlin.reflect.jvm.internal.KProperty1Impl.get(KProperty1Impl.kt:35)
04-22 15:27:18.362 15369 15369 E AndroidRuntime: 	at com.google.jetpackcamera.feature.preview.PreviewViewModel.getSettingsDiff(PreviewViewModel.kt:367)
04-22 15:27:18.362 15369 15369 E AndroidRuntime: 	at com.google.jetpackcamera.feature.preview.PreviewViewModel.access$getSettingsDiff(PreviewViewModel.kt:89)
04-22 15:27:18.362 15369 15369 E AndroidRuntime: 	at com.google.jetpackcamera.feature.preview.PreviewViewModel$1$1$invokeSuspend$$inlined$transform$1$1.emit(Emitters.kt:41)
04-22 15:27:18.362 15369 15369 E AndroidRuntime: 	at com.google.jetpackcamera.settings.LocalSettingsRepository$special$$inlined$map$1$2.emit(Emitters.kt:50)
04-22 15:27:18.362 15369 15369 E AndroidRuntime: 	at kotlinx.coroutines.flow.internal.SafeCollectorKt$emitFun$1.invoke(SafeCollector.kt:11)
[...]

PR #379, opened 2025-07-26
Set exact parameter type

Jaehwa-Noh:
Current in `CameraAppSettings.applyDiff` defined `settingExtractor` as CameraAppSettings.() -> R (edited in #346) , but it actual type is KProperty1 (e.g. `CameraAppSettings::cameraLensFacing`) so that I changed it to actual type.

https://github.com/google/jetpack-camera-app/blob/c99c8f40bf3a71a841e8ec7111bb229314b30258/feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewViewModel.kt#L349-L383

I'm not sure whether this change does any chance to occur #345 again.

### DIFF
--- a/feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewViewModel.kt
+++ b/feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewViewModel.kt
@@ -84,6 +84,7 @@ import dagger.assisted.AssistedFactory
 import dagger.assisted.AssistedInject
 import dagger.hilt.android.lifecycle.HiltViewModel
 import java.util.LinkedList
+import kotlin.reflect.KProperty1
 import kotlinx.atomicfu.atomic
 import kotlinx.coroutines.CoroutineStart
 import kotlinx.coroutines.Deferred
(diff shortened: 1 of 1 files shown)


temcguir:
While it is a good idea to make the API more explicit, my main concern with using `KProperty1` is that it requires importing from the `kotlin.reflect` package. This might inadvertently signal that using reflection is encouraged, which is a practice we should generally deter.

Do you have explicit conerns with the way the code is now?

Jaehwa-Noh:
I address the flow of my thinking.

1. I check the dependency at `:feature:preview`, having `PreviewModel.kt`, module with gradle dependencies.
It shows kotlin-reflect dependency in `com.google.dagger:dagger-compiler:2.56.2`
>   |    +--- com.squareup:kotlinpoet:1.11.0
     |    |    +--- org.jetbrains.kotlin:kotlin-reflect:1.6.10

So I  thought I can use reflection type.

2. And then `applyDiff` function still pass argument as Reflection type.
So I assume that I can use reflection.

So that I change it `KCallable` type to specific type, though when I see your comment I noticed that you don't want to import reflect package explicit.

If import refection is discouraged, this PR should closed.

temcguir:
Those dependencies are only associated with the annotation processors and won't be added to the app/library. If you run `./gradlew :feature:preview:dependencies --configuration stableDebugRuntimeClasspath`, for instance, you won't see any kotlin reflect dependencies.

Regardless, we do want to discourage explicit reflection in the code base, so I think we should leave this as-is for now and close this PR. I appreciate the attempt to make things clearer, though. Thank you!

PR #363, opened 2025-06-18
Refactor PreviewUiState into ui/uistate/capture/compound

davidjiagoogle:
Rename PreviewUiState to CaptureUiState.

Turn all its values into UiState and move them into ui/components/capture.

PR #382, opened 2025-07-31
Fix double tap lens flipping bug

davidjiagoogle:
In #363, the rememberUpdatedState() was removed. This cause the flipLensUiState used here being not updated. Use rememberUpdatedState() to track the updated uiState.

#381 shows the debugging process


feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewScreen.kt
```diff
@@ -369,13 +370,13 @@ private fun ContentScreen(
     Scaffold(
         snackbarHost = { SnackbarHost(hostState = snackbarHostState) }
     ) {
+        val flipLensState by rememberUpdatedState(
```
Kimblebee: instead of using flipLensState as a stable reference here, could we just change the `.pointerInput(Unit)` key from `Unit` to `onFlipCamera` from the `CameraXViewfinder` invocation inside `PreviewDisplay`?

PR #387, opened 2025-08-05
Centralize ExternalCaptureMode logic, use CaptureEvent Channel & Nav …

temcguir:
This commit significantly refactors how external capture modes and their
associated events are handled. It simplifies `ExternalCaptureMode` into an
enum, centralizes its related logic, introduces a Kotlin `Channel` for
communicating capture events, and passes `captureUris` via Navigation
arguments.

Key changes include:

- **`ExternalCaptureMode` as Enum:**
  - `ExternalCaptureMode` is now an `enum class` with distinct values
    (`Standard`, `ImageCapture`, `VideoCapture`, `MultipleImageCapture`),
    making it more straightforward and suitable for use as a navigation
    argument (being added in follow-up PR).

- **`CaptureEvent` Channel in `PreviewViewModel`:**
  - `PreviewViewModel` now uses a `Channel<CaptureEvent>` (`_captureEvents`)
    to broadcast image and video capture results (success & error)
    as `ImageCaptureEvent` or `VideoCaptureEvent` instances.
  - This replaces previous direct callbacks for specific events and
    centralizes event emission within the ViewModel.

- **Centralized Event Handling in `MainActivity` & Navigation:**
  - `MainActivity` now determines the `externalCaptureMode`, `captureUris`,
     and `debugSettings` using properties instead of functions.
  - `captureUris` (a list of URIs for external capture modes) is now
    passed as a navigation argument to the Preview screen.
  - A new `captureEventCallback` is introduced in `MainActivity` and passed
    to `JcaApp`. This callback listens to `CaptureEvent`s (originating
    from `PreviewViewModel`) and handles Android system interactions like
    setting activity results and finishing for external capture intents.

- **UI Layer Adaption:**
  - `JcaApp` and `PreviewScreen` are updated to receive `captureUris` via
    navigation and to propagate/use the `onCaptureEvent` callback.
  - UI components (`ControlsBottom`, `CaptureButton`, various state adapters)
    are refactored to work with the `ExternalCaptureMode` enum. Their
    responsibility for handling specific external capture outcomes is
    reduced.

- **Navigation Update:**
  - The navigation graph is updated to include `captureUris` as an
    argument for the Preview route.

- **Code Simplification:**
  - Helper functions for creating specific `ExternalCaptureMode` instances
    with callbacks have been removed. Using the `CaptureEvent` channel, there
    everything can now be handled in one place.

This refactoring aims to:
- Decouple UI components from the specifics of external intent handling.
- Provide a clearer, more reactive flow for capture events.
- Simplify state management within `PreviewViewModel`.
- Utilize Navigation Component arguments for passing necessary capture
  information like `captureUris`, and prepare for removal of AssistedInjection.


app/src/main/java/com/google/jetpackcamera/MainActivity.kt
```diff
@@ -181,100 +187,109 @@ class MainActivity : ComponentActivity() {
...
-                        if (event is ExternalCaptureMode.ImageCaptureEvent.ImageSaved) {
-                            val resultIntent = Intent()
-                            resultIntent.putExtra(MediaStore.EXTRA_OUTPUT, event.savedUri)
-                            setResult(RESULT_OK, resultIntent)
-                            Log.d(TAG, "onImageCapture, finish()")
-                            finish()
-                        }
+    private val captureEventCallback: (CaptureEvent) -> Unit
+        get() {
+            val pictureTakenUriList by lazy { arrayListOf<String>() }
```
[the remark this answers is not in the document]
temcguir: Done

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewScreen.kt
```diff
@@ -138,6 +138,13 @@ fun PreviewScreen(
         }
     }
 
+    val currentOnCaptureEvent by rememberUpdatedState(onCaptureEvent)
+    LaunchedEffect(Unit) {
+        for (event in viewModel.captureEvents) {
+            currentOnCaptureEvent(event)
```
davidjiagoogle, line 144: What does this do?

davidjiagoogle: My understanding is this would trigger the onCaptureEvent callback for all the existing capture events in the channel every time the UI refreshes? So say if we are in ExternalMultipleImageCaptureMode and already want to capture 3 images, this is my understanding of the sequence of triggers:

Capture 1st image -> UI refreshes -> callback triggers on Image1
Capture 2nd image -> UI refreshes -> callback triggers on Image1, Image2
Capture 3rd image -> UI refreshes -> callback triggers on Image1, Image2, Image3

If this is true, we would be over-triggering the callback right?

davidjiagoogle: by Image1 i meant SequentialImageSaved event for the 1st image (assume capture was successful) and so forth

temcguir: Because `ExternalCaptureMode` was converted into an enum and no longer contains a lambda to handle when images / videos are captured, the same logic is now handled by the `onCaptureEvent` callback that is passed in to `PreviewScreen`.

In the ViewModel, we now have a `Channel<CaptureEvent>` where all image / video capture events are sent. It is a Channel, rather than something like a StateFlow, so we don't elide any updates- every event gets handled. Each event is passed along to `onCaptureEvent` so that it can be passed up to MainActivity and handled there. We wrap the callback in `rememberUpdatedState` so that if the `onCaptureEvent` lambda changes during a recomposition, we send the updates to the new callback (see [this documentation](https://developer.android.com/develop/ui/compose/side-effects#rememberupdatedstate) for an explanation on that).

temcguir: Sorry I sent the last message before I saw your next 2 messages.

> Capture 1st image -> UI refreshes -> callback triggers on Image1
> Capture 2nd image -> UI refreshes -> callback triggers on Image1, Image2
> Capture 3rd image -> UI refreshes -> callback triggers on Image1, Image2, Image3
>
> If this is true, we would be over-triggering the callback right?

No, that's not right. A `for` loop on a `Channel` will remove the event from the channel on each update. When there are no other updates, the for-loop suspends until the next update.

Also, because we're in a `LaunchedEffect(Unit)`, there isn't any link to the UI refreshing and the callback triggering.

davidjiagoogle (approved):

PR #398, opened 2025-09-11
Fix flash mode adapter state logic

temcguir:
The `FlashModeUiStateAdapter` was not correctly updating its
state from `systemConstraints`. It was comparing against its
previous state, which could lead to stale UI when available
flash modes changed.

This has been corrected, and the adapter now properly derives
its state from the latest `systemConstraints`.

Tests have been added to verify the adapter's
logic.

The tests also check for referential equality to prevent
unnecessary object allocations.


donovanfm:
LGTM! Testing the app, everything works as expected.

PR #399, opened 2025-09-12
Fixes to reduce image and video capture test flakiness

davidjiagoogle:
1. Extend some of the timeouts as UI operations may lag on certain emulators, resulting in premature test failure.
2. Introduced longClickForVideoRecordingCheckingElapsedTime() to check for ElapsedTimeText to ensure the video is recording for the correct duration
3. Dismiss snackbar when multiple images need to be taken to speed up the testing process


app/src/androidTest/java/com/google/jetpackcamera/ImageCaptureDeviceTest.kt
```diff
@@ -186,9 +186,7 @@ internal class ImageCaptureDeviceTest {
                     composeTestRule.onNodeWithTag(CAPTURE_BUTTON).isDisplayed()
                 }
 
-                composeTestRule.onNodeWithTag(CAPTURE_BUTTON)
-                    .assertExists()
-                    .performTouchInput { longClick(durationMillis = 3_000) }
+                composeTestRule.longClickForVideoRecording()
 
                 composeTestRule.ensureTagNotAppears(
                     VIDEO_CAPTURE_EXTERNAL_UNSUPPORTED_TAG,
```
temcguir, line 192: This combination of `longClickForVideoRecording` and `ensureTagNotAppears` aren't compatible. If the video capture isn't supported, then `longClickForVideoRecording` will time out waiting for the `ELAPSED_TIME_TAG`.

Perhaps you could insert an optional lambda parameter into `longClickForVideoRecording()` which can run inside the `waitUntil` block in `longClickForVideoRecording`. That way you can make an assertion within that check. @gemini-code-assist can you suggest how to structure the code to insert a lambda like that and resolve this issue?

davidjiagoogle: @temcguir I don't think that's correct. Notice what you described about ElapsedTimeText happens in longClickForVideoRecordingCheckingElapsedTime() not longClickForVideoRecording()

temcguir: Ah you're right. I guess my comment was assuming they are using the same logic path. The code here is fine.

It may not matter here, but we can do something similar to [ComposeTestRule.idleForVideoDuration()](https://github.com/google/jetpack-camera-app/blob/66581816aaf37a4d4e8250e99567ef4c541e2016/app/src/androidTest/java/com/google/jetpackcamera/utils/ComposeTestRuleExt.kt#L143) where we check for the "Video capture failed" message while we're waiting for a recording to finish. This will ensure we don't wait for too long on failures. Something like this:

```
fun ComposeTestRule.longClickForVideoRecordingCheckingElapsedTime(
    durationMillis: Long = VIDEO_DURATION_MILLIS
    checkWhileWaiting: () -> Unit = {
        // If the video capture fails, there is no point to continue waiting. Assert.
        onNodeWithTag(VIDEO_CAPTURE_FAILURE_TAG).assertIsNotDisplayed()
    }
) {
    onNodeWithTag(CAPTURE_BUTTON)
        .assertExists()
        .performTouchInput {
            down(center)
        }
    waitUntil(timeoutMillis = ELAPSED_TIME_TEXT_TIMEOUT_MILLIS) {
        checkWhileWaiting()
        onNodeWithTag(ELAPSED_TIME_TAG).isDisplayed()
    }
    waitUntilVideoRecordingDurationAtLeast(durationMillis, checkWhileWaiting)
    onNodeWithTag(CAPTURE_BUTTON)
        .assertExists()
        .performTouchInput {
            up()
        }
}

private fun ComposeTestRule.waitUntilVideoRecordingDurationAtLeast(
durationMillis: Long,
checkWhileWaiting: () -> Unit = {}
) {
    waitUntil(timeoutMillis = ELAPSED_TIME_TEXT_TIMEOUT_MILLIS) {
        checkWhileWaiting()
        val text = onNodeWithTag(ELAPSED_TIME_TAG)
            .fetchSemanticsNode()
            .config.getOrNull(SemanticsProperties.Text)
            ?.firstOrNull()?.text

        val duration = text?.let { parseMinSecToMillis(it) }

        duration != null && duration >= durationMillis
    }
}
```

Note the "checkWhileWaiting()" argument which will cause the test to fail if there is an error message saying the video recording failed.

temcguir (approved):

PR #400, opened 2025-09-17
introduce new styleguide for gemini

Kimblebee:
introduces a styleguide to help direct the behavior of gemini's integration in this repository


temcguir (approved):
Looks like a great starting point. We can make incremental updates as needed once this is merged.

PR #406, opened 2025-09-30
Set swiftshader_indirect on Android Instrumented Test

Jaehwa-Noh:
Follow this documentation.
https://developer.android.com/studio/test/gradle-managed-devices#run-tests

>Note: When using Gradle-managed devices on servers that don't support hardware rendering, such as GitHub Actions, you need to specify the following flag: -Pandroid.testoptions.manageddevices.emulator.gpu=swiftshader_indirect. If you use PowerShell, you have to surround the flag with quotes.

Supposed to be set swiftshader_indirect.

### DIFF
--- a/.github/workflows/PullRequestWorkflow.yaml
+++ b/.github/workflows/PullRequestWorkflow.yaml
@@ -139,7 +139,7 @@ jobs:
       - name: Run instrumentation tests
         uses: gradle/actions/setup-gradle@v3
         with:
-          arguments: ${{ matrix.device.name }}StableDebugAndroidTest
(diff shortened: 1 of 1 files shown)


temcguir:
We currently have this flag specified [here](https://github.com/google/jetpack-camera-app/blob/d13b56e9d90c6559cbfb9affdfa25cd96e34abd3/gradle.properties#L47). Do you know if it _must_ be specified on the command line?

Jaehwa-Noh:
@temcguir Gradle.properties file is actually managing all of properties, such as system, Gradle and project.
https://docs.gradle.org/current/userguide/build_environment.html#the_gradle_properties_file

It is totally same with command line passed with property.

I didn't catch that was already in gradle.properties.
I will close this PR.

Thank you.

PR #426, opened 2025-11-11
Unit Tests for LocalMediaRepository and PostCaptureViewModel

Kimblebee:
Unit tests and associated changes

### DIFF
--- a/data/media/src/main/kotlin/com/google/jetpackcamera/data/media/LocalMediaRepository.kt
+++ b/data/media/src/main/kotlin/com/google/jetpackcamera/data/media/LocalMediaRepository.kt
@@ -31,6 +31,9 @@ import android.util.Size
 import androidx.core.net.toFile
 import com.google.jetpackcamera.core.common.IODispatcher
 import dagger.hilt.android.qualifiers.ApplicationContext
+import java.io.File
+import java.io.IOException
+import javax.inject.Inject
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.ExperimentalCoroutinesApi
@@ -40,9 +43,6 @@ import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.asStateFlow
 import kotlinx.coroutines.flow.update
 import kotlinx.coroutines.withContext
-import java.io.File
-import java.io.IOException
-import javax.inject.Inject
 
 private const val TAG = "LocalMediaRepository"
 private const val IMAGE_MIME_TYPE = "image/jpeg"
@@ -129,9 +129,7 @@ class LocalMediaRepository
     /**
      * Deletes the specified media from either the cache or the MediaStore.
      */
-    override suspend fun deleteMedia(
-        mediaDescriptor: MediaDescriptor.Content
-    ): Boolean =
+    override suspend fun deleteMedia(mediaDescriptor: MediaDescriptor.Content): Boolean =
         repositoryScope.async {
             val result =
                 if (mediaDescriptor.isCached) {
@@ -139,20 +137,21 @@ class LocalMediaRepository
                 } else {
                     context.contentResolver.delete(mediaDescriptor.uri, null, null) >= 1
                 }
-            if (result && !mediaDescriptor.isCached)
+            if (result && !mediaDescriptor.isCached) {
                 Log.d(TAG, "deleted saved media")
+            }
 
             result
         }.await()
-
 
     /**
      * Deletes a cached media file.
      */
     private fun deleteCachedMedia(mediaDescriptor: MediaDescriptor.Content): Boolean {
         val result = mediaDescriptor.uri.toFile().delete()
-        if (result)
+        if (result) {
             Log.d(TAG, "deleted cached media")
+        }
         return result
     }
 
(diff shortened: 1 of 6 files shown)


core/common/src/main/java/com/google/jetpackcamera/core/common/CommonModule.kt
```diff
@@ -45,8 +45,22 @@ class CommonModule {
...
+
+    @Module
+    @InstallIn(SingletonComponent::class)
+    object CoroutinesModule {
+        @ApplicationScope
+        @Singleton
+        @Provides
+        fun providesApplicationScope(): CoroutineScope =
+            CoroutineScope(SupervisorJob() + Dispatchers.Default)
+    }
```
temcguir: I'm a bit wary to add something like this. We don't really want to promote just running tasks scoped to the application level.

feature/postcapture/src/main/java/com/google/jetpackcamera/feature/postcapture/PostCaptureViewModel.kt
```diff
@@ -77,30 +79,53 @@ class PostCaptureViewModel @Inject constructor(
...
         releasePlayer()
 
         val mediaDescriptor = uiState.value.mediaDescriptor
 
         // todo(kc): improve cache cleanup strategy
         if ((mediaDescriptor as? MediaDescriptor.Content)?.isCached == true) {
-            viewModelScope.launch {
+            // use application scope for cleanup
+            // coroutine will not be cancelled when ViewModel dies
+            applicationScope.launch {
```
temcguir: This should be moved into the MediaRepository. The MediaRepository is already a singleton and should have its own coroutine scope for handling tasks like this.

data/media/src/test/java/com/google/jetpackcamera/media/LocalMediaRepositoryTest.kt
```diff
@@ -0,0 +1,447 @@
...
+    private lateinit var repository: LocalMediaRepository
+    private val testDispatcher = StandardTestDispatcher()
+
+    @Before
+    fun setup() {
+        MockitoAnnotations.openMocks(this)
+
+        // We strictly mock the ContentResolver to avoid Robolectric DB flakes
+        context = spy(ApplicationProvider.getApplicationContext<Context>())
+        doReturn(mockContentResolver).`when`(context).contentResolver
```
temcguir: We should not be mocking classe we don't own, especially Android framework classes (see go/tott/678, go/noandroidmocks). Instead, you could create a FakeContentProvider that extends ContentProvider.

In fact, we should add a line to the style guide that mentions we should never be mocking Android framework classes.

feature/postcapture/src/test/java/com/google/jetpackcamera/feature/postcapture/PostCaptureViewModelTest.kt
```diff
@@ -0,0 +1,300 @@
...
+
+    private lateinit var viewModel: PostCaptureViewModel
+    private lateinit var mockMediaRepository: MediaRepository
+
+    // Fakes
+
+    private val currentMediaFlow = MutableStateFlow<MediaDescriptor>(MediaDescriptor.None)
+
+    // --- SHARED TEST DATA ---
+    val mockThumbnail: Bitmap = Mockito.mock(Bitmap::class.java)
```
temcguir: Mocking the bitmap in this way might be ok since it's just being used as a placeholder and we don't call any of the bitmap methods, but you could also just use a real Bitmap with size of 1, 1 or something as a placeholder instead.

core/common/src/main/java/com/google/jetpackcamera/core/common/CommonModule.kt
Kimblebee: thats fair. my thought was that something like a call to save or delete a file shouldnt be interrupted by an accidental nav back

feature/postcapture/src/main/java/com/google/jetpackcamera/feature/postcapture/PostCaptureViewModel.kt
Kimblebee: done

core/common/src/main/java/com/google/jetpackcamera/core/common/CommonModule.kt
temcguir: You're right that it shouldn't, but that should happen when you use the scope that's part of MediaRepository since it's a singleton whose scope won't exit. I'm just saying we don't want a general scope that anyone can use to run unconstrained tasks.

core/common/src/main/java/com/google/jetpackcamera/core/common/CommonModule.kt
```diff
@@ -47,6 +47,10 @@ class CommonModule {
         CoroutineScope(SupervisorJob() + dispatcher)
 }
 
+@Retention(AnnotationRetention.RUNTIME)
+@Qualifier
+annotation class ApplicationScope
+
```
temcguir, line 53: Can this be deleted?

data/media/src/main/kotlin/com/google/jetpackcamera/data/media/LocalMediaRepository.kt
```diff
@@ -102,55 +107,61 @@ class LocalMediaRepository
...
      * @throws IOException if an I/O error occurs during the copy operation.
      */
     @Throws(IOException::class)
-    override suspend fun copyToUri(
-        contentResolver: ContentResolver,
-        mediaDescriptor: MediaDescriptor.Content,
-        destinationUri: Uri
-    ) = copyUriToUri(contentResolver, sourceUri = mediaDescriptor.uri, destinationUri)
+    override suspend fun copyToUri(mediaDescriptor: MediaDescriptor.Content, destinationUri: Uri) =
+        copyUriToUri(context.contentResolver, sourceUri = mediaDescriptor.uri, destinationUri)
```
temcguir, line 164: This seems like an operation that should be happening on the IO dispatcher. Could you wrap it in a `repositoryScope.async {...}.await()`? @gemini-code-assist

temcguir: @gemini-code-assist I'm talking about the `copyToUri` function, not `deleteMedia`.

PR #434, opened 2025-11-17
Refactor MediaRepository data model

Kimblebee:
This is the first in a series of commits to upgrade the current postcapture experience.

This commit refactors the `MediaRepository` to improve state management and adds new features for media operations.

Key changes:
- Introduces `currentMedia` as a `StateFlow` to observe the current media item.
- Adds functionality to delete media from the device and save media to the MediaStore.
- Updates the `MediaDescriptor` data model to better represent media content.
- Adds unit tests for the `LocalMediaRepository`.

### DIFF
--- a/.gemini/styleguide.md
+++ b/.gemini/styleguide.md
@@ -39,6 +39,7 @@ When reviewing a pull request, focus on the following key areas:
     * **When Tests are Missing:** If a PR introduces a significant feature or modifies logic without corresponding tests, flag this omission. Suggest a name for a new test class (e.g., `NewFeatureViewModelTest`) and outline what it should verify (e.g., "This test should check that the UI state updates correctly when the user performs X action").
     * **When Tests are Present in the PR:** If new or modified tests are included, review them for thoroughness. Check for coverage of happy paths, failure scenarios, and relevant edge cases.
     * **Analysis of Existing Tests:** Identify existing test files in the target branch that are relevant to the code being changed in the PR but were **not** modified. Analyze these files to see if the PR introduces new logic that is not covered. If you find coverage gaps, cite the filename (e.g., `ExistingViewModelTest.kt`) and suggest specific test cases to add (e.g., "Consider adding a test case here to handle the new `XYZ` state introduced in the PR.").
+    * **Testing Strategy: Fakes over Mocks for Unowned Classes:** Never mock concrete classes that the project doesn't own. Instead, for testing, use fakes that implement or extend the class's interface. This ensures that tests are robust against changes in third-party library internals and encourages testing against behavior rather than implementation details.
     * **Descriptive Test Names:** Test function names must be clear, descriptive, and follow a consistent pattern.
 
 6.  **Documentation Sync**


data/media/src/main/kotlin/com/google/jetpackcamera/data/media/LocalMediaRepository.kt
```diff
@@ -15,58 +15,266 @@
...
+    override val currentMedia = _currentMedia.asStateFlow()
+
+    /**
+     * Sets the current media descriptor.
+     *
+     * @param pendingMedia The [MediaDescriptor] to set as current.
+     */
+    override suspend fun setCurrentMedia(pendingMedia: MediaDescriptor) {
+        _currentMedia.update { pendingMedia }
+        Log.d(TAG, "set new media $pendingMedia")
```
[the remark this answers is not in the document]
Kimblebee: ack

feature/postcapture/src/main/java/com/google/jetpackcamera/feature/postcapture/PostCaptureViewModel.kt
```diff
@@ -64,8 +64,16 @@ class PostCaptureViewModel @Inject constructor(
...
+                null
+            )
+            is MediaDescriptor.Content.Video -> contentResolver.delete(
+                mediaDescriptor.uri,
+                null,
+                null
+            )
             MediaDescriptor.None -> {}
         }
         _uiState.update { it.copy(mediaDescriptor = MediaDescriptor.None, media = Media.None) }
```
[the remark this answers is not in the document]
Kimblebee: this PR is focusing on refactoring the data model. the api calls are addressed in a subsequent pr

data/media/src/main/kotlin/com/google/jetpackcamera/data/media/LocalMediaRepository.kt
```diff
@@ -15,58 +15,266 @@
...
+        if (finalResult && currentMedia.value == mediaDescriptor) {
+            setCurrentMedia(MediaDescriptor.None)
         }
+        return finalResult
+    }
+
+    /**
+     * Deletes a cached media file.
+     */
+    private fun deleteCachedMedia(mediaDescriptor: MediaDescriptor.Content): Boolean {
```
davidjiagoogle, line 210: Correct me if I'm wrong, this looks like deleting a cached Media is no different deleting a uncached media. Both media is saved to the uri location. What's the difference between using contentResolve.delete(uri) and using toFile().delete? If no difference, why differntiate between the two cases?

davidjiagoogle: To a larger extend of this question (this may be answered along with the question above), why differentiate between cached and uncached MediaDescriptors if both are just files saved to a given uri?

davidjiagoogle: Just read LocalMediaRespositoryTest. It seem to suggest that an uncached media is always saved to MediaStore and a cached media is always saved to a specific location and never to media store. Then this would make sense. I think adding this bit of clarification to the javadoc of isCached field would help with easier understanding

data/media/src/test/java/com/google/jetpackcamera/media/LocalMediaRepositoryTest.kt
```diff
@@ -0,0 +1,455 @@
...
+        // When
+        val result = repository.load(MediaDescriptor.None)
+        // Then
+        assertEquals(Media.None, result)
+    }
+
+    @Test
+    fun deleteMedia_savedMedia_callsContentResolverDelete() = runTest(testDispatcher) {
+        // Given a saved media item (isCached = false)
+        val mediaUri = ContentUris.withAppendedId(
```
davidjiagoogle: Does this media already exist in mediaStore at the beginning of this test?

data/media/src/main/kotlin/com/google/jetpackcamera/data/media/LocalMediaRepository.kt
Kimblebee: that is correct. cached media is saved to the application's `cachedir` and is **not** registered to MediaStore. So I cannot delete it using a ContentResolver. 

I check what Uri's are or aren't managed by a contentProvider using `uri.scheme == ContentResolver.SCHEME_CONTENT`

I can add extra Kdocs to help clarify that.

data/media/src/test/java/com/google/jetpackcamera/media/LocalMediaRepositoryTest.kt
Kimblebee: thanks for catching. I've updated the test to insert and check for the media before and after deleting

PR #438, opened 2025-11-17
Add support for caching captured media flow

Kimblebee:
This commit introduces a new capture flow that allows media (images and videos) to be cached temporarily instead of being saved directly to the MediaStore. This enables features like post-capture review and editing before final saving. _Note that the only flow currently available is the `Immediate` flow, unless you manually change the code to use `CacheAndReview` before building._

Key changes:
- **`SaveMode` Model:** Introduced a new `SaveMode` sealed interface (`Immediate` vs. `CacheAndReview`) to control the capture destination.
- **`SaveLocation.Cache`:** Added a `Cache` option to the `SaveLocation` model, directing captures to the app's cache directory.
- **Updated `CaptureEvent`s:** Added `SingleImageCached` and `VideoCached` events to the `ImageCaptureEvent` and `VideoCaptureEvent` models to signal when a capture is cached.
- **Camera System Implementation:** Modified `CameraXCameraSystem` and `CameraSession` to handle saving files to the cache directory when `SaveLocation.Cache` is specified.
- **ViewModel Logic:** The `PreviewViewModel` now uses the `SaveMode` to determine the save location. When media is cached, it triggers navigation to the post-capture screen and updates the `MediaRepository` with the cached media descriptor. Snackbar notifications for successful captures are now suppressed when caching.


feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewViewModel.kt
```diff
@@ -127,6 +128,8 @@ class PreviewViewModel @Inject constructor(
     private val constraintsRepository: ConstraintsRepository,
     private val mediaRepository: MediaRepository
 ) : ViewModel() {
+    private val saveMode: SaveMode = SaveMode.Immediate
```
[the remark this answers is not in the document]
Kimblebee: yes, there is another PR after this that configures the savemode using a new BuildConfig field.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraSession.kt
```diff
@@ -915,6 +915,33 @@ private fun getPendingRecording(
                     .build()
             videoCaptureUseCase.output.prepareRecording(context, mediaStoreOutput)
         }
+
+        is SaveLocation.Cache -> {
+            try {
+                // 1. Get the app's cache directory
+                val cacheDir = context.applicationContext.cacheDir
```
temcguir: Should this also try to use saveLocation.cacheDir before using the application's cacheDir?

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewViewModel.kt
```diff
@@ -637,11 +650,21 @@ class PreviewViewModel @Inject constructor(
...
+                            ImageCaptureEvent.SingleImageCached(savedUri)
+                        } else {
+                            ImageCaptureEvent.SingleImageSaved(savedUri)
+                        }
+                    }
+                    if (saveLocation !is SaveLocation.Cache) {
+                        updateLastCapturedMedia()
+                    } else {
+                        savedUri?.let {
+                            setMediaRepository(MediaDescriptor.Content.Image(it, null, true))
```
temcguir: `setMediaRepository` seems like a non-descriptive name for this. Maybe `postCurrentMediaToMediaRepository` or something similar?

PR #439, opened 2025-11-17
Implement the capture-cache-review user flow

Kimblebee:
Introduces a new post-capture workflow that allows users to review and decide whether to save or discard captured media. This feature is enabled by passing a `KEY_REVIEW_AFTER_CAPTURE` boolean extra in the launch intent.

Key changes include:
- Added `CachedImageCaptureDeviceTest`, `CachedVideoRecordingDeviceTest`, and `PostCaptureTest` to validate the cache-and-review functionality.
- Modified `MainActivity` to read the `KEY_REVIEW_AFTER_CAPTURE` intent extra and propagate it to the navigation graph.
- Updated `PreviewNavigation` to handle the `shouldCacheReview` navigation argument.
- Adjusted `PreviewViewModel` to select the appropriate `SaveMode` based on the navigation arguments, with `SaveMode.Immediate` as the default.
- Created an `AppModule` to provide the default `SaveMode` via dependency injection.
- Added new test utilities to support the cache-and-review UI tests.

### DIFF
--- a/app/src/main/java/com/google/jetpackcamera/AppModule.kt
+++ b/app/src/main/java/com/google/jetpackcamera/AppModule.kt
@@ -15,7 +15,9 @@
  */
 package com.google.jetpackcamera
 
+import com.google.jetpackcamera.core.common.DefaultCaptureModeOverride
 import com.google.jetpackcamera.core.common.DefaultSaveMode
+import com.google.jetpackcamera.model.CaptureMode
 import com.google.jetpackcamera.model.SaveMode
 import dagger.Module
 import dagger.Provides
@@ -25,6 +27,12 @@ import dagger.hilt.components.SingletonComponent
 @Module
 @InstallIn(SingletonComponent::class)
 object AppModule {
+    /**
+     * provides the default [CaptureMode] to override by the app
+     */
+    @Provides
+    @DefaultCaptureModeOverride
+    fun providesDefaultCaptureModeOverride(): CaptureMode = CaptureMode.STANDARD
 
     /**
      * provides the default [SaveMode] to be used by the app
--- a/feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewViewModel.kt
+++ b/feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewViewModel.kt
@@ -178,11 +178,13 @@ class PreviewViewModel @Inject constructor(
     private fun CameraAppSettings.applyExternalCaptureMode(
         externalCaptureMode: ExternalCaptureMode
     ): CameraAppSettings {
-        val captureMode = externalCaptureMode.toCaptureMode()
-        return if (captureMode == this.captureMode) {
+        val requiredCaptureModeOverride = externalCaptureMode.toCaptureMode()
+        return if (requiredCaptureModeOverride == null ||
+            requiredCaptureModeOverride == this.captureMode
+        ) {
             this
         } else {
-            this.copy(captureMode = captureMode)
+            this.copy(captureMode = requiredCaptureModeOverride)
         }
     }
 
@@ -324,7 +326,13 @@ class PreviewViewModel @Inject constructor(
                             cameraState,
                             externalCaptureMode
                         ),
-                        hdrUiState = hdrUiState
+                        hdrUiState = hdrUiState,
+
+                        imageWellUiState = ImageWellUiState.from(
+                            trackedUiState.recentCapturedMedia,
+                            cameraState.videoRecordingState
+                        )
+
                     )
                 }
             }.collect {}
@@ -411,12 +419,8 @@ class PreviewViewModel @Inject constructor(
 
     fun updateLastCapturedMedia() {
         viewModelScope.launch {
-            val lastCapturedMediaDescriptor = mediaRepository.getLastCapturedMedia()
-            _captureUiState.update { old ->
-                (old as? CaptureUiState.Ready)?.copy(
-                    imageWellUiState =
-                    ImageWellUiState.from(lastCapturedMediaDescriptor)
-                ) ?: old
+            trackedPreviewUiState.update { old ->
+                old.copy(recentCapturedMedia = mediaRepository.getLastCapturedMedia())
             }
         }
     }
@@ -425,7 +429,7 @@ class PreviewViewModel @Inject constructor(
         ExternalCaptureMode.ImageCapture -> CaptureMode.IMAGE_ONLY
         ExternalCaptureMode.MultipleImageCapture -> CaptureMode.IMAGE_ONLY
         ExternalCaptureMode.VideoCapture -> CaptureMode.VIDEO_ONLY
-        ExternalCaptureMode.Standard -> CaptureMode.STANDARD
+        ExternalCaptureMode.Standard -> null
     }
 
     /**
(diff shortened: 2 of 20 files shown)


.github/workflows/PullRequestWorkflow.yaml
```diff
@@ -145,6 +145,59 @@ jobs:
             */build/reports/androidTests/**/${{ matrix.device.name }}
             */build/outputs/androidTest-results/**/${{ matrix.device.name }}
 
+  android-test-stable-cache:
```
temcguir: Adding a whole new workflow just for this feature is a bit overkill. Many of the tests being run don't depend on taking a picture or video, so there's not much value in running them again.

I would much prefer if this could just be set as an intent argument to enable/disable the cache-and-review workflow. Then we could just parameterize any of the tests that would use this workflow to run with cache-and-review and without.

app/src/main/java/com/google/jetpackcamera/AppModule.kt
```diff
@@ -0,0 +1,33 @@
...
+import dagger.hilt.components.SingletonComponent
+
+@Module
+@InstallIn(SingletonComponent::class)
+object AppModule {
+
+    // This is the only place that needs to know about the flavor-specific BuildConfig
+
+    @Provides
+    fun providesSaveMode(): SaveMode =
```
temcguir: This could be `providesDefaultSaveMode`, where it will get the saveMode only if it's not provided by an intent argument. Then we can get rid of the whole build flavor / BuildConfig and just provide `Immediate` as the default save mode. For those tests that will actually use this workflow, we can just use an intent to enable the cache-and-review flow. If someone wanted to fork the app, they could provide their own defaultSaveMode to be whatever they want.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewViewModel.kt
```diff
@@ -124,12 +124,11 @@ private const val IMAGE_CAPTURE_TRACE = "JCA Image Capture"
 class PreviewViewModel @Inject constructor(
     private val cameraSystem: CameraSystem,
     private val savedStateHandle: SavedStateHandle,
+    private val saveMode: SaveMode,
```
temcguir: With changes suggested in AppModule, this would become `val @DefaultSaveMode defaultSaveMode: SaveMode`.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewViewModel.kt
```diff
@@ -124,12 +124,11 @@ private const val IMAGE_CAPTURE_TRACE = "JCA Image Capture"
...
     private val settingsRepository: SettingsRepository,
     private val constraintsRepository: ConstraintsRepository,
     private val mediaRepository: MediaRepository
 ) : ViewModel() {
-    private val saveMode: SaveMode = SaveMode.Immediate
```
temcguir, line 131: With changes suggested in AppModule, this would become `private val saveMode: SaveMode = defaultSaveMode` for now.

Once the intent argument is added it would become:
`private val saveMode: SaveMode = savedStateHandle.getRequestedSaveMode() ?: defaultSaveMode`

.github/workflows/PullRequestWorkflow.yaml
Kimblebee: ok. one question I'd have down the line is what kind of intent argument could we use to specify a maximum quantity of captures before navigating to postcapture? right now this feature just supports single images/videos, but could be reworked later to support a list of Uri's if necessary

temcguir: I think the current multi-capture intent argument could work where it's a list of URIs. Once the same number of items as the number of URIs has been captured, then it could automatically go to post-capture. They'll be saved to the provided URIs in order they are taken. The post-capture screen in this case would probably need to be a ViewPager so you could review all of the images/videos, however.

So you would have one intent extra which is a lsit of URIs, and another intent argument which is a boolean to do cache-and-review.

I think the multi-capture scenario needs to be reworked a little anyways. The ImageWell should only allow you to view the images you've captured during that session.

app/src/androidTest/java/com/google/jetpackcamera/utils/UiTestUtil.kt
```diff
@@ -72,6 +72,12 @@ val compatMainActivityExtras: Bundle?
     }
 
 val debugExtra: Bundle = Bundle().apply { putBoolean("KEY_DEBUG_MODE", true) }
+val cacheExtra: Bundle = Bundle().apply { putBoolean("KEY_REVIEW_AFTER_CAPTURE", true) }
+val debugCacheExtra: Bundle = Bundle().apply {
+    putBoolean("KEY_REVIEW_AFTER_CAPTURE", true)
```
temcguir: Looks like you have this in here twice.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/navigation/PreviewNavigation.kt
```diff
@@ -35,12 +35,16 @@ import com.google.jetpackcamera.feature.preview.PreviewScreen
...
+import com.google.jetpackcamera.feature.preview.navigation.PreviewRoute.ARG_SAVE_MODE
 import com.google.jetpackcamera.model.CaptureEvent
 import com.google.jetpackcamera.model.DebugSettings
 import com.google.jetpackcamera.model.ExternalCaptureMode
+import com.google.jetpackcamera.model.SaveMode
 
 object PreviewRoute {
     internal const val ARG_EXTERNAL_CAPTURE_MODE: String = "externalCaptureMode"
+
+    internal const val ARG_SAVE_MODE: String = "saveMode"
```
temcguir: nit: since this is a boolean, maybe `ARG_REVIEW_AFTER_CAPTURE` would be a better name for the argument.

app/src/androidTest/java/com/google/jetpackcamera/utils/UiTestUtil.kt
Kimblebee: turns out i didn't need that bundle, i've removed it

README.md
```diff
@@ -648,6 +649,22 @@ the user interface and does not impact the captured images or videos.
...
+
+JCA supports two different media saving workflows, controlled by an intent extra. This allows developers to test and validate different user experiences for capturing and saving media.
+
+*   **Immediate Save (Default):** This is the default behavior. When a photo or video is captured, it is immediately saved to the device's MediaStore.
+*   **Cache-and-Review:** In this mode, captured media is saved to a temporary cache. After capture, the user is taken to a post-capture review screen where they can view the media and decide to either save it to the MediaStore or delete it. **Note: When JCA is launched via an `ACTION_IMAGE_CAPTURE` or `ACTION_VIDEO_CAPTURE` intent, the cache-and-review mode is bypassed, and media is saved directly to the provided output URI.**
+
+#### How to Enable Cache-and-Review
+
+You can enable the cache-and-review workflow by launching the app with the following extra:
+`intent.putExtra("should_cache_and_review", true)`
```
temcguir: nit: This doesn't look right. I think it should be "KEY_REVIEW_AFTER_CAPTURE".

.github/workflows/PullRequestWorkflow.yaml
Kimblebee: sounds good, i'll write up some tasks for these.

PR #440, opened 2025-11-17
Imagewell UI refactor

Kimblebee:
tidies up the imagewell composable.
returns the ImageWell component to the capture screen, and updates the animations whenever a new image is loaded

### DIFF
--- a/ui/uistateadapter/capture/src/test/java/com/google/jetpackcamera/ui/uistateadapter/capture/ImageWellUiStateAdapterTest.kt
+++ b/ui/uistateadapter/capture/src/test/java/com/google/jetpackcamera/ui/uistateadapter/capture/ImageWellUiStateAdapterTest.kt
@@ -15,6 +15,7 @@
  */
 package com.google.jetpackcamera.ui.uistateadapter.capture
 
+import android.graphics.Bitmap
 import android.net.Uri
 import com.google.jetpackcamera.core.camera.VideoRecordingState
 import com.google.jetpackcamera.data.media.MediaDescriptor
@@ -24,12 +25,13 @@ import org.junit.Test
 import org.mockito.Mockito.mock
 
 class ImageWellUiStateAdapterTest {
-    private val testThumbnailUri: Uri = mock(Uri::class.java)
+    private val testUri: Uri = mock(Uri::class.java)
+    val testBitmap: Bitmap = mock(Bitmap::class.java)
 
     @Test
-    fun from_mediaContentAndVideoInactive_returnsLastCapture() {
+    fun from_mediaContentAndVideoInactive_withThumbnail_returnsLastCapture() {
         // Given
-        val mediaDescriptor = MediaDescriptor.Content.Image(testThumbnailUri, null, false)
+        val mediaDescriptor = MediaDescriptor.Content.Image(testUri, testBitmap, false)
         val videoRecordingState = VideoRecordingState.Inactive()
 
         // When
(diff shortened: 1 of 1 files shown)


ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/CaptureLayout.kt
```diff
@@ -132,7 +135,7 @@ private fun VerticalMaterialControls(
                                     .height(120.dp),
                                 contentAlignment = Alignment.Center
                             ) {
-                                //  item to be placed left of capture button goes here
+                                imageWell(Modifier)
```
temcguir, line 138: This seems like something we should show if the `SaveMode` is `Immediate`, but not if it's `Cached`. Maybe there should be a check of whether the `ImageWellUiState` is `Unavailable` and it's controlled by the `SaveMode` in the ViewModel or UiStateAdapter.

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/CaptureLayout.kt
```diff
@@ -221,6 +224,13 @@ private fun CaptureLayoutPreview() {
                     .background(Color.Cyan)
             )
         },
+        imageWell = { modifier ->
+            Box(
+                modifier = modifier
+                    .size(48.dp)
+                    .background(Color.Cyan)
```
temcguir, line 231: Should this color really be Cyan or was that being used for debugging?

Kimblebee: this is just a placeholder shape for the Preview of the layout. this is an `@Preview` composable

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/CaptureLayout.kt
Kimblebee: I think access to recently saved media _could_ be accessible from either the cache flow or immediate. My thought is that someone could still make use of it by seeing their most recent capture, whether it was taken 2 minutes or 2 days ago.

I believe it would make more sense to hide the  imageWell if it only displayed the most recent saved media from that app's active **session**.

This PR is based on #434 and isn't attached to the capture-cache-review commits, so I'll refrain from adding checks for now.

I also think that maintaining an imagewell on the preview screen would effectively render the "successful image/video capture" snackbar redundant, so we may also need to consider checks for that..

temcguir: Yeah that makes sense. I think you can leave as-is for now, but it's worth discussing (in the design doc) when the ImageWell should be visible.

ui/components/capture/src/main/res/values/strings.xml
```diff
@@ -139,4 +139,7 @@
     <string name="quick_settings_more_text">More</string>
     <string name="quick_settings_more_description">Settings icon</string>
 
+    <!-- ImageWell -->
+    <string name="image_well_content_description">Click to </string>
```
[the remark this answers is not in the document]
temcguir: Please resolve

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/ImageWell.kt
```diff
@@ -15,95 +15,74 @@
...
-
-                is MediaDescriptor.None -> null
+    Box(
+        modifier = modifier
+            .testTag(IMAGE_WELL_TAG)
+            .size(IconButtonDefaults.largeContainerSize())
+            .padding(18.dp)
+            .border(2.dp, Color.White, shape)
+            .clip(shape)
+            .clickable(onClick = onClick, enabled = enabled)
```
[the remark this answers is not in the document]
temcguir: Please resolve

temcguir (approved):

PR #444, opened 2025-11-19
Introduce DefaultCaptureModeOverride

davidjiagoogle:
Add a Qualifier annotation in CommonModule named DefaultCaptureModeOverride. This annotation can be used by apps that deploys JCA as a library to override the default captureMode without making changes to the view model

### DIFF
--- a/feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewViewModel.kt
+++ b/feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewViewModel.kt
@@ -321,7 +321,13 @@ class PreviewViewModel @Inject constructor(
                             cameraState,
                             externalCaptureMode
                         ),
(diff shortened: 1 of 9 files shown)


feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewViewModel.kt
```diff
@@ -174,7 +174,7 @@ class PreviewViewModel @Inject constructor(
         externalCaptureMode: ExternalCaptureMode
     ): CameraAppSettings {
         val captureMode = externalCaptureMode.toCaptureMode()
```
temcguir: Perhaps we should rename this val to something like "requiredCaptureModeOverride" and rename the `toCaptureMode` extension function to be `toRequiredCaptureMode`. This will make the code a little clearer.

temcguir (approved):

PR #448, opened 2025-11-21
centralize filepath creation

Kimblebee:
Extracted the filename and output directory generation logic from the `LocalMediaRepository`, `PostCaptureViewModel`, and `CameraCoreUtil` into a new `FilePathGenerator` interface.

This change makes the project friendlier to forks by allowing developers to provide their own implementation of `FilePathGenerator` to customize file naming and save locations without modifying the core application logic.

Key changes:
- Created `FilePathGenerator` interface in `core:common`.
- Implemented `JcaFilePathGenerator` in the `app` module as the default implementation.
- Injected `FilePathGenerator` into `LocalMediaRepository` and `CameraXCameraSystem`.
- Updated tests to provide a `FakeFilePathGenerator` and adjusted mock stubs.

### DIFF
--- a/core/common/src/main/java/com/google/jetpackcamera/core/common/FakeFilePathGenerator.kt
+++ b/core/common/src/main/java/com/google/jetpackcamera/core/common/FakeFilePathGenerator.kt
@@ -16,7 +16,6 @@
 package com.google.jetpackcamera.core.common
 
 import android.os.Environment
-import com.google.jetpackcamera.core.common.FilePathGenerator.Companion.constructFilename
 import java.io.File
 import java.util.Date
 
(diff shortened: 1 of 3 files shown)


core/common/src/main/java/com/google/jetpackcamera/core/common/FakeFilePathGenerator.kt
```diff
@@ -0,0 +1,57 @@
...
+
+    override fun generateVideoFilename(suffixText: String?, fileExtension: String?): String {
+        return constructFilename(
+            "JCA-test-recording",
+            createTimestamp().toString(),
+            suffixText,
+            fileExtension
+        )
+    }
+}
```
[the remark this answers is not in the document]
Kimblebee: the JcaFilePathGenerator can be swapped with whatever a forking developer wants. the two classes are similar  because this is the original JCA repository.

core/common/src/main/java/com/google/jetpackcamera/core/common/FilePathGenerator.kt
```diff
@@ -0,0 +1,85 @@
...
+     *
+     * @param suffixText additional text to append to the end of the generated filename, before the
+     * file extension.
+     * @param fileExtension the extension to be appended at the end of the generated filename
+     */
+    fun generateVideoFilename(suffixText: String? = null, fileExtension: String? = ".mp4"): String
+
+    companion object {
+        /**
+         * constructs a filename based on the provided values
```
temcguir: Please expand on this documentation. It looks like this is constructing a filename by concatenating the given filename fragments with "-" between them. It looks like this is just here for convenience? There is no requirement that filenames follow this pattern, right?

Kimblebee: you're correct. There is no requirement for filenames to follow such a pattern. This kind of function should honestly just live inside of the implementing class, so that anyone may decide how the filename should be constructed.

PR #453, opened 2025-12-03
Create the core/camera/postProcess module

davidjiagoogle:
Introduce ImagePostProcessor and its feature key. A map of key value pairs is provided via PostProcessModule. Features should implement ImagePostProcessor with its own feature key. ImagePostProcessor.postProcessImage() performs a post-processing operations on the image after it has been successfully captured and saved.

### DIFF
--- a/core/camera/src/androidTest/java/com/google/jetpackcamera/core/camera/CameraXCameraSystemTest.kt
+++ b/core/camera/src/androidTest/java/com/google/jetpackcamera/core/camera/CameraXCameraSystemTest.kt
@@ -143,7 +143,7 @@ class CameraXCameraSystemTest {
     }
 
     @Test
-    fun captureImage_withPostProcessor_postProcessNotCalled(): Unit = runBlocking {
+    fun captureImage_withInvalidUri_postProcessNotCalled(): Unit = runBlocking {
         // Arrange.
         val imagePostProcessor = FakeImagePostProcessor()
         val cameraSystem =
@@ -163,7 +163,7 @@ class CameraXCameraSystemTest {
     }
 
     @Test
-    fun canCaptureImage_withPostProcessor_postProcessFailed(): Unit = runBlocking {
+    fun captureImage_withFailingPostProcessor_imageStillSaved(): Unit = runBlocking {
         // Arrange.
         val imagePostProcessor = FakeImagePostProcessor(shouldError = true)
         val cameraSystem =


core/camera/postprocess/src/main/java/com/google/jetpackcamera/core/camera/postprocess/PostProcessModule.kt
```diff
@@ -0,0 +1,53 @@
...
+                Map.Entry<
+                    ImagePostProcessorFeatureKey,
+                    @JvmSuppressWildcards Provider<ImagePostProcessor>
+                    >
+                >
+        ): Map<
+            ImagePostProcessorFeatureKey,
+            @JvmSuppressWildcards Provider<ImagePostProcessor>
+            > =
+            entries.associate { it.key to it.value }
```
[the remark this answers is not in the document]
davidjiagoogle: Out of this PR's scope

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraSystem.kt
```diff
@@ -615,6 +619,14 @@ constructor(
             closeable?.close()
         }.also { outputFileResults ->
             outputFileResults.savedUri?.let {
+                for ((key, value) in imagePostProcessors) {
+                    try {
+                        value.get().postProcessImage(it)
+                        Log.d(TAG, "Post processing image with $key")
+                    } catch (e: Exception) {
```
temcguir: Catching and logging the exception here means that the user won't be aware when post-processing fails. It's probably better to throw the exception so the failure can be reported to the UI.

core/camera/postprocess/src/main/java/com/google/jetpackcamera/core/camera/postprocess/ImagePostProcessor.kt
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * An interface for performing post-processing on a captured image.
+ *
+ * Each post-processor should be associated with a unique [ImagePostProcessorFeatureKey].
+ */
+interface ImagePostProcessor {
+    /**
+     * Performs a post-processing operation on the image located at the given URI after an image has
+     * been successfully captured and saved.
```
temcguir: Let's add a line here that care should be taken to ensure that blocking I/O does not happen during this function call. There should be no expectation about which thread this is called from.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraSystem.kt
```diff
@@ -116,7 +118,9 @@ constructor(
     availabilityCheckers:
     Map<LowLightBoostFeatureKey, @JvmSuppressWildcards Provider<LowLightBoostAvailabilityChecker>>,
     effectProviders:
-    Map<LowLightBoostFeatureKey, @JvmSuppressWildcards Provider<LowLightBoostEffectProvider>>
+    Map<LowLightBoostFeatureKey, @JvmSuppressWildcards Provider<LowLightBoostEffectProvider>>,
+    val imagePostProcessors:
+    Map<ImagePostProcessorFeatureKey, @JvmSuppressWildcards Provider<ImagePostProcessor>>
 ) : CameraSystem {
```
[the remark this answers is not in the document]
davidjiagoogle: I tried to use CameraXCameraSystemTest to test the takePicture() function with a real CameraXCameraSystem, but the imageCaptureUseCase.takePicture() call hangs indefinitely.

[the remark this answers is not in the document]
davidjiagoogle: There doesn't seem to be a good way to test this right now

core/camera/postprocess/src/main/java/com/google/jetpackcamera/core/camera/postprocess/PostProcessModule.kt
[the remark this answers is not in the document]
temcguir: This looks like a legitimate copy/paste error? It's not providing a `ImageProcessorAvailabilityChecker` like in LLB, but an `ImageProcessor`, so the suggestion makes sense.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraSystem.kt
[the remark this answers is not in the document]
temcguir: I've created #455 which fixes the issue and adds an image capture test. You can build your tests on top of that. PTAL.

core/camera/src/androidTest/java/com/google/jetpackcamera/core/camera/CameraXCameraSystemTest.kt
```diff
@@ -104,7 +112,9 @@ class CameraXCameraSystemTest {
     @Test
     fun canCaptureImage(): Unit = runBlocking {
         // Arrange.
-        val cameraSystem = createAndInitCameraXCameraSystem()
+        val imagePostProcessor = FakeImagePostProcessor()
+        val cameraSystem =
+            createAndInitCameraXCameraSystem(fakeImagePostProcessor = imagePostProcessor)
```
temcguir: You should put this in a separate test so that we're testing image capture with and without the post-processor. You should also have a test that includes a post-processor that fails (throws an exception).

@gemini-code-assist can suggest the function names for these tests and how they should be broken down.

temcguir: You're still missing tests for:
1. No post-processor
2. Failing post-processor

core/camera/src/androidTest/java/com/google/jetpackcamera/core/camera/CameraXCameraSystemTest.kt
```diff
@@ -143,7 +143,7 @@ class CameraXCameraSystemTest {
     }
 
     @Test
-    fun captureImage_withPostProcessor_postProcessNotCalledIfFailed(): Unit = runBlocking {
+    fun captureImage_withPostProcessor_postProcessNotCalled(): Unit = runBlocking {
```
temcguir: How about "captureImage_withInvalidUri_postProcessNotCalled".

core/camera/src/androidTest/java/com/google/jetpackcamera/core/camera/CameraXCameraSystemTest.kt
```diff
@@ -162,6 +162,30 @@ class CameraXCameraSystemTest {
         assertThat(imagePostProcessor.postProcessImageCalled).isFalse()
     }
 
+    @Test
+    fun canCaptureImage_withPostProcessor_postProcessFailed(): Unit = runBlocking {
```
temcguir: How about "captureImage_withFailingPostProcessor_imageStillSaved"

PR #456, opened 2025-12-05
Adopt CameraX Feature Group API

### LINKED ISSUE #454: Camera session failure with HDR + 60 FPS
Tested devices: Pixel 9, Pixel 9 Pro XL

JCA allows enabling both HDR and 60 FPS at the same time. However, it seems to lead to session failure (black screen, record failure etc.) on Pixel 9 devices while Pixel 10 XL works fine.

Ideally, we should allow these settings to be enabled together only if device supports them (e.g. Pixel 10 XL). This can be checked with the CameraX Feature Groups API.

https://github.com/user-attachments/assets/bbb7dd9d-d9ed-4fbb-a8af-33dac15139c5

Zeronfinity:
# Description

This PR adopts the new **CameraX Feature Group API** to validate feature combinations and dynamically update system constraints. This ensures that the app only allows supported combinations of settings (e.g., HDR, FPS, Stabilization), preventing runtime failures on devices that do not support specific combinations (like HDR + 60 FPS on Pixel 9).

## Key Changes
*   **CameraX Update:** Updated CameraX dependencies to `1.6.0-SNAPSHOT` to access the stable Feature Groups API.
*   **Feature Mapping:** Added `FeatureGroupData` to map JCA internal models (`DynamicRange`, `VideoQuality`, `StabilizationMode`, etc.) to CameraX's `GroupableFeature`.
*   **Dynamic Constraints:** Implemented `updateSystemConstraintsByFeatureGroups` in `CameraXCameraSystem`. This method asynchronously validates the current session settings against the device capabilities and filters out unsupported options for other settings (e.g., if 60 FPS is selected, it might disable HDR if they can't be used together).
*   **Refactoring:** Migrated from `UseCaseGroup` to `SessionConfig` for single camera session creation to align with the new API usage and allow for `isSessionConfigSupported` checks.

## Bug Fixes & Enhancements
*   Fixes a crash/black screen issue on Pixel 9 devices when enabling both HDR and 60 FPS.
*   Stream mode is now selectable based on whether supported, rather being always disabled, when HDR is enabled.
*   Feature combinations can now be supported more often (due to the additional guaranteed combinations from feature combination query API), e.g. UHD recording with single stream mode in Samsung S23U.

### System constraints not updating for multiple updates from setting screen
The system constraints are currently updated only when the camera session is initialized or re-created (e.g., upon returning to the preview screen). This means that if a user sets feature options (like UHD and  60 FPS) directly from the settings menu *without* navigating back to the preview screen, the system constraints is not re-evaluated.
Consequently, it allows users to select a combination of features that are not mutually supported by the device, which can lead to issues when they eventually navigate back to the preview screen.  

**TODO:** Addressing this limitation requires further architectural refactoring to trigger constraint updates dynamically upon any individual setting change. Or, the constraints repository should be updated with the info of all possible constraints at initialization, rather than only the settings state navigable from the current settings state.

### Features incompatible with CameraX Feature Group API
CameraX depends on the Camera2 feature combination query (FCQ) API which guarantees support only a limited set of features. So, the scenarios using features incompatible with the FCQ API do not get much benefits from this PR.

### Asynchronous system constraints update
System constraints are updated asynchronously, creating a window where a user might enable an unsupported combination of features before the update completes.  

**Future Work:** This could be addressed by waiting for the constraints update to finish and restricting setting changes until it is complete.

## Tests
*   Added `FeatureGroupDataTest` (unit tests) to verify mapping logic between JCA models and CameraX `GroupableFeature`.
*   Added `setMultipleFeatures_systemConstraintsUpdatedAndFeaturesSetIfSupported` to `CameraXCameraSystemTest` (integration test) to verify that the camera system dynamically updates constraints and sets features correctly based on CameraX feature group compatibility.

## Additional Context
*   https://android-developers.googleblog.com/2025/10/beyond-single-features-guaranteeing.html

## Related Issues
*   Fixes #457
*   Fixes #454

### DIFF
--- a/core/camera/src/androidTest/java/com/google/jetpackcamera/core/camera/CameraXCameraSystemTest.kt
+++ b/core/camera/src/androidTest/java/com/google/jetpackcamera/core/camera/CameraXCameraSystemTest.kt
@@ -18,6 +18,7 @@ package com.google.jetpackcamera.core.camera
 import android.app.Application
 import android.content.ContentResolver
 import android.net.Uri
+import android.util.Log
 import androidx.test.ext.junit.runners.AndroidJUnit4
 import androidx.test.filters.LargeTest
 import androidx.test.platform.app.InstrumentationRegistry
@@ -37,11 +38,15 @@ import com.google.jetpackcamera.model.CaptureMode
 import com.google.jetpackcamera.model.DynamicRange
 import com.google.jetpackcamera.model.FlashMode
 import com.google.jetpackcamera.model.Illuminant
+import com.google.jetpackcamera.model.ImageOutputFormat
 import com.google.jetpackcamera.model.LensFacing
 import com.google.jetpackcamera.model.SaveLocation
 import com.google.jetpackcamera.model.StabilizationMode
+import com.google.jetpackcamera.model.StreamConfig
 import com.google.jetpackcamera.model.VideoQuality
 import com.google.jetpackcamera.settings.model.CameraAppSettings
+import com.google.jetpackcamera.settings.model.CameraConstraints
+import com.google.jetpackcamera.settings.model.CameraSystemConstraints
 import com.google.jetpackcamera.settings.model.DEFAULT_CAMERA_APP_SETTINGS
 import com.google.jetpackcamera.settings.model.forCurrentLens
 import java.io.File
@@ -85,6 +90,7 @@ class CameraXCameraSystemTest {
         private const val GENERAL_TIMEOUT_MS = 3_000L
         private const val RECORDING_TIMEOUT_MS = 10_000L
         private const val RECORDING_START_DURATION_MS = 500L
+        private const val TAG = "CameraXCameraSystemTest"
     }
 
     @get:Rule
@@ -347,68 +353,205 @@ class CameraXCameraSystemTest {
     }
 
     @Test
-    fun setMultipleFeatures_systemConstraintsUpdatedAndFeaturesSettableIfSupported() = runBlocking {
-        // Arrange.
-        val cameraSystem = createAndInitCameraXCameraSystem()
-        val systemConstraintsFlow = cameraSystem.getSystemConstraints()
+    fun setMultipleFeatures_systemConstraintsUpdatedAndFeaturesSetIfSupported() = runBlocking {
+        // TODO: Add STREAM_CONFIG_SINGLE to the featuresToTest list. This currently leads to flaky
+        //  crashes due to some camera effect related surface not being cleaned up properly somehow.
+        //  This doesn't seem to be related to the primary purpose of this test, so simply excluding
+        //  it for now.
+        val featuresToTest = listOf(
+            Feature.DYNAMIC_RANGE_HLG10,
+            Feature.FPS_60,
+            Feature.VIDEO_QUALITY_UHD
+        )
 
-        // Each camera run/update should lead to a new systemConstraints update
-        var systemConstraintsUpdate = systemConstraintsFlow.observeNextUpdate(this)
-        cameraSystem.startCameraAndWaitUntilRunning()
+        // TODO: Run a subset of permutations instead of all when `featuresToTest` increases.
+        featuresToTest.permutations().forEach { orderedFeatures ->
+            Log.d(TAG, "Testing $orderedFeatures")
 
-        var currentConstraints = systemConstraintsUpdate.awaitUntil()
-        val lensFacing = cameraSystem.getCurrentSettings().value?.cameraLensFacing
+            // Setup
+            val cameraSystem = createAndInitCameraXCameraSystem()
 
-        // Act: For each of the features — HDR, 60 FPS, UHD recording, await previous constraints
-        //  update and set the feature if the constraints supports it.
+            // Initial run: each camera run/update should lead to a new systemConstraints update
+            var currentConstraints = cameraSystem.getSystemConstraints().observeNextUpdate(
+                this
+            ).let {
+                cameraSystem.startCameraAndWaitUntilRunning()
+                it.awaitUntil()
+            }
 
-        if (
-            currentConstraints
-                .perLensConstraints[lensFacing]
-                ?.supportedDynamicRanges
-                ?.contains(DynamicRange.HLG10) == true
-        ) {
-            systemConstraintsUpdate = systemConstraintsFlow.observeNextUpdate(this)
-            cameraSystem.setDynamicRange(DynamicRange.HLG10)
-            currentConstraints = systemConstraintsUpdate.awaitUntil()
+            val lensFacing =
+                requireNotNull(cameraSystem.getCurrentSettings().value?.cameraLensFacing)
+
+            orderedFeatures.forEach { feature ->
+                currentConstraints = when (feature) {
+                    Feature.DYNAMIC_RANGE_HLG10 -> feature.tryApplyFeature(
+                        scope = this,
+                        expectedValue = DynamicRange.HLG10,
+                        lensFacing = lensFacing,
+                        cameraSystemConstraints = currentConstraints,
+                        constraintsState = cameraSystem.getSystemConstraints(),
+                        cameraSystem = cameraSystem,
+                        setFeature = { cameraSystem.setDynamicRange(DynamicRange.HLG10) },
+                        getNewFeatureValue = { it?.dynamicRange }
+                    ) { constraints ->
+                        constraints
+                            ?.supportedDynamicRanges
+                            ?.contains(DynamicRange.HLG10) == true
+                    }
+
+                    Feature.FPS_60 -> feature.tryApplyFeature(
+                        scope = this,
+                        expectedValue = 60,
+                        lensFacing = lensFacing,
+                        cameraSystemConstraints = currentConstraints,
+                        constraintsState = cameraSystem.getSystemConstraints(),
+                        cameraSystem = cameraSystem,
+                        setFeature = { cameraSystem.setTargetFrameRate(60) },
+                        getNewFeatureValue = { it?.targetFrameRate }
+                    ) { constraints ->
+                        constraints
+                            ?.supportedFixedFrameRates
+                            ?.contains(60) == true
+                    }
+
+                    Feature.VIDEO_QUALITY_UHD -> feature.tryApplyFeature(
+                        scope = this,
+                        expectedValue = VideoQuality.UHD,
+                        lensFacing = lensFacing,
+                        cameraSystemConstraints = currentConstraints,
+                        constraintsState = cameraSystem.getSystemConstraints(),
+                        cameraSystem = cameraSystem,
+                        setFeature = { cameraSystem.setVideoQuality(VideoQuality.UHD) },
+                        getNewFeatureValue = { it?.videoQuality }
+                    ) { constraints ->
+                        constraints
+                            ?.supportedVideoQualitiesMap
+                            ?.get(cameraSystem.getCurrentSettings().value?.dynamicRange)
+                            ?.contains(VideoQuality.UHD) == true
+                    }
+
+                    Feature.STABILIZATION_MODE_ON -> feature.tryApplyFeature(
+                        scope = this,
+                        expectedValue = StabilizationMode.ON,
+                        lensFacing = lensFacing,
+                        cameraSystemConstraints = currentConstraints,
+                        constraintsState = cameraSystem.getSystemConstraints(),
+                        cameraSystem = cameraSystem,
+                        setFeature = { cameraSystem.setStabilizationMode(StabilizationMode.ON) },
+                        getNewFeatureValue = { it?.stabilizationMode }
+                    ) { constraints ->
+                        constraints
+                            ?.supportedStabilizationModes
+                            ?.contains(StabilizationMode.ON) == true
+                    }
+
+                    Feature.IMAGE_FORMAT_JPEG_ULTRA_HDR -> feature.tryApplyFeature(
+                        scope = this,
+                        expectedValue = ImageOutputFormat.JPEG_ULTRA_HDR,
+                        lensFacing = lensFacing,
+                        cameraSystemConstraints = currentConstraints,
+                        constraintsState = cameraSystem.getSystemConstraints(),
+                        cameraSystem = cameraSystem,
+                        setFeature = {
+                            cameraSystem.setImageFormat(
+                                ImageOutputFormat.JPEG_ULTRA_HDR
+                            )
+                        },
+                        getNewFeatureValue = { it?.imageFormat }
+                    ) { constraints ->
+                        constraints
+                            ?.supportedImageFormatsMap
+                            ?.get(cameraSystem.getCurrentSettings().value?.streamConfig)
+                            ?.contains(ImageOutputFormat.JPEG_ULTRA_HDR) == true
+                    }
+
+                    Feature.STREAM_CONFIG_SINGLE -> feature.tryApplyFeature(
+                        scope = this,
+                        expectedValue = StreamConfig.SINGLE_STREAM,
+                        lensFacing = lensFacing,
+                        cameraSystemConstraints = currentConstraints,
+                        constraintsState = cameraSystem.getSystemConstraints(),
+                        cameraSystem = cameraSystem,
+                        setFeature = { cameraSystem.setStreamConfig(StreamConfig.SINGLE_STREAM) },
+                        getNewFeatureValue = { it?.streamConfig }
+                    ) { constraints ->
+                        constraints
+                            ?.supportedStreamConfigs
+                            ?.contains(StreamConfig.SINGLE_STREAM) == true
+                    }
+                }
+            }
         }
-
-        if (
-            currentConstraints
-                .perLensConstraints[lensFacing]
-                ?.supportedFixedFrameRates
-                ?.contains(60) == true
-        ) {
-            systemConstraintsUpdate = systemConstraintsFlow.observeNextUpdate(this)
-            cameraSystem.setTargetFrameRate(60)
-            currentConstraints = systemConstraintsUpdate.awaitUntil()
-        }
-
-        if (
-            currentConstraints
-                .perLensConstraints[lensFacing]
-                ?.supportedVideoQualitiesMap
-                ?.get(cameraSystem.getCurrentSettings().value?.dynamicRange)
-                ?.contains(VideoQuality.UHD) == true
-        ) {
-            systemConstraintsUpdate = systemConstraintsFlow.observeNextUpdate(this)
-            cameraSystem.setVideoQuality(VideoQuality.UHD)
-            systemConstraintsUpdate.awaitUntil()
-        }
-
-        // Wait to ensure the async updateSystemConstraintsByFeatureGroups has time to run
-        // and potentially crash if there's an issue.
-        // As currentConstraints has already been updated, we just ensure we reach here without crash.
-
-        // Assert.
-        // If the test reaches here without crashing, it passes.
-        // This ensures that the feature group logic doesn't cause runtime exceptions
-        // even when high-end features are requested.
-        return@runBlocking
     }
 
     private fun <T> StateFlow<T?>.observeNextUpdate(scope: CoroutineScope): Deferred<T> {
         return scope.async { drop(1).filterNotNull().first() }
+    }
+
+    private suspend fun <T> Feature.tryApplyFeature(
+        scope: CoroutineScope,
+        expectedValue: T,
+        lensFacing: LensFacing,
+        cameraSystemConstraints: CameraSystemConstraints,
+        constraintsState: StateFlow<CameraSystemConstraints?>,
+        cameraSystem: CameraSystem,
+        setFeature: suspend () -> Unit,
+        getNewFeatureValue: (CameraAppSettings?) -> T?,
+        isSupported: (CameraConstraints?) -> Boolean
+    ): CameraSystemConstraints {
+        // Check support
+        if (!isSupported(cameraSystemConstraints.perLensConstraints[lensFacing])) {
+            Log.d(TAG, "Skipping $this: Not supported by current constraints.")
+            return cameraSystemConstraints
+        }
+
+        Log.d(TAG, "Applying $this...")
+
+        // Prepare observer
+        val nextUpdate = constraintsState.observeNextUpdate(scope)
+
+        setFeature()
+
+        // Wait to verify constraints is updated
+        val newConstraints = nextUpdate.awaitUntil()
+
+        // Verify feature is set according to current settings
+        assertThat(getNewFeatureValue(cameraSystem.getCurrentSettings().value)).isEqualTo(
+            expectedValue
+        )
+
+        return newConstraints
+    }
+
+    private fun <T> List<T>.permutations(): List<List<T>> {
+        if (isEmpty()) {
+            // Base case: an empty list has one permutation (the empty list itself)
+            return listOf(emptyList())
+        }
+
+        val result = mutableListOf<List<T>>()
+        val head = first() // Take the first element
+        val tail = drop(1) // Get the rest of the list
+
+        // Recursively get permutations of the tail
+        tail.permutations().forEach { permOfTail ->
+            // Insert the head element at all possible positions in each permutation of the tail
+            for (i in 0..permOfTail.size) {
+                val newPerm = permOfTail.toMutableList()
+                newPerm.add(i, head)
+                result.add(newPerm)
+            }
+        }
+        return result
+    }
+
+    private enum class Feature {
+        DYNAMIC_RANGE_HLG10,
+        FPS_60,
+        VIDEO_QUALITY_UHD,
+        STABILIZATION_MODE_ON,
+        IMAGE_FORMAT_JPEG_ULTRA_HDR,
+        STREAM_CONFIG_SINGLE
     }
 
     suspend fun <T> Deferred<T>.awaitUntil(timeout: Duration = 2.seconds): T {
--- a/core/camera/src/androidTest/java/com/google/jetpackcamera/core/camera/CameraXCameraSystemTest.kt
+++ b/core/camera/src/androidTest/java/com/google/jetpackcamera/core/camera/CameraXCameraSystemTest.kt
@@ -34,26 +34,33 @@ import com.google.jetpackcamera.core.camera.utils.APP_REQUIRED_PERMISSIONS
 import com.google.jetpackcamera.core.camera.utils.provideUpdatingSurface
 import com.google.jetpackcamera.core.common.testing.FakeFilePathGenerator
 import com.google.jetpackcamera.model.CaptureMode
+import com.google.jetpackcamera.model.DynamicRange
 import com.google.jetpackcamera.model.FlashMode
 import com.google.jetpackcamera.model.Illuminant
 import com.google.jetpackcamera.model.LensFacing
 import com.google.jetpackcamera.model.SaveLocation
 import com.google.jetpackcamera.model.StabilizationMode
+import com.google.jetpackcamera.model.VideoQuality
 import com.google.jetpackcamera.settings.model.CameraAppSettings
 import com.google.jetpackcamera.settings.model.DEFAULT_CAMERA_APP_SETTINGS
 import com.google.jetpackcamera.settings.model.forCurrentLens
 import java.io.File
 import java.util.AbstractMap
 import javax.inject.Provider
+import kotlin.time.Duration
+import kotlin.time.Duration.Companion.seconds
 import kotlin.time.DurationUnit
 import kotlin.time.toDuration
 import kotlinx.coroutines.CompletableDeferred
 import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Deferred
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.async
 import kotlinx.coroutines.cancelAndJoin
 import kotlinx.coroutines.channels.ReceiveChannel
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.drop
 import kotlinx.coroutines.flow.filterNotNull
 import kotlinx.coroutines.flow.first
 import kotlinx.coroutines.flow.map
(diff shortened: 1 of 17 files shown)


core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraSystem.kt
```diff
@@ -500,37 +531,342 @@ constructor(
...
+        cameraAppSettings: CameraAppSettings,
+        initialCameraConstraints: CameraConstraints,
+        cameraInfo: CameraInfo
+    ): Set<DynamicRange> {
+        Log.d(TAG, "filterGroupableDynamicRanges")
+
+        return initialCameraConstraints.supportedDynamicRanges.filter {
+            cameraAppSettings.copyDynamicRange(it, initialSystemConstraints)
+                .supportsGrouping(cameraInfo)
+        }.toSet()
```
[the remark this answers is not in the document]
Zeronfinity: False positive.

core/camera/src/test/java/com/google/jetpackcamera/core/camera/FeatureGroupDataTest.kt
```diff
@@ -0,0 +1,291 @@
...
+@RunWith(RobolectricTestRunner::class)
+class FeatureGroupDataTest {
+    @Test
+    fun requiresFeatureGroupQuery_returnsTrue_whenTwoOrMoreExplicit() {
+        val collection = listOf(
+            ExplicitlyGroupable(GroupableFeatures.SD_RECORDING),
+            ExplicitlyGroupable(GroupableFeature.IMAGE_ULTRA_HDR),
+            InexplicitlyGroupable
+        )
+        assertTrue(collection.requiresFeatureGroupQuery())
```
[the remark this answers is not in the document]
Zeronfinity: Done

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraSession.kt
```diff
@@ -560,30 +519,90 @@ internal fun applyDeviceRotation(deviceRotation: DeviceRotation, useCaseGroup: U
...
+                    }
+                }
+            )
+        }
+    }
+    if (cameraEffect == null &&
+        sessionSettings.streamConfig == StreamConfig.SINGLE_STREAM
+    ) {
+        cameraEffect = SingleSurfaceForcingEffect(sessionScope)
+    }
```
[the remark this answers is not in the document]
Zeronfinity: Let's minimize changes.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraSystem.kt
```diff
@@ -500,37 +531,342 @@ constructor(
...
+                    )
+//                    unsupportedStabilizationFpsMap = unsupportedStabilizationFpsMap
+                )
+
+        systemConstraints = systemConstraints.copy(perLensConstraints = updatedPerLensConstraints)
+
+        constraintsRepository.updateSystemConstraints(systemConstraints)
+
+        Log.d(TAG, "updateSystemConstraints: updated systemConstraints = $systemConstraints")
+    }
```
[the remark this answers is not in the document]
Zeronfinity: Added test.

core/camera/src/test/java/com/google/jetpackcamera/core/camera/FeatureGroupDataTest.kt
```diff
@@ -0,0 +1,291 @@
...
+import com.google.jetpackcamera.model.DynamicRange
+import com.google.jetpackcamera.model.ImageOutputFormat
+import com.google.jetpackcamera.model.LowLightBoostPriority
+import com.google.jetpackcamera.model.StabilizationMode
+import com.google.jetpackcamera.model.StreamConfig
+import com.google.jetpackcamera.model.VideoQuality
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertFalse
+import org.junit.Assert.assertNull
+import org.junit.Assert.assertTrue
```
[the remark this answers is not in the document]
Zeronfinity: Done

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraSystem.kt
```diff
@@ -500,37 +531,342 @@ constructor(
...
+                )
+
+            createSessionConfig(
+                cameraConstraints = cameraConstraints,
+                initialTransientSettings = transientSettings,
+                videoCaptureUseCase = videoCaptureUseCase,
+                sessionSettings = sessionSettings,
+                sessionScope = CoroutineScope(defaultDispatcher + Job())
+            )
+        }
```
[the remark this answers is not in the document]
Zeronfinity: When `coroutineScope` is used, some issues are observed. It seems like the coroutineScope block never completes for StreamConfig.SINGLE_STREAM. Probably some code flow downstream is keeping the coroutineScope busy forever, so this approach is taken.

Modified the code to cancel the job after query is completed and added comment there to explain the situation.

core/camera/src/androidTest/java/com/google/jetpackcamera/core/camera/CameraXCameraSystemTest.kt
```diff
@@ -161,6 +170,74 @@ class CameraXCameraSystemTest {
...
+
+        // Wait to ensure the async updateSystemConstraintsByFeatureGroups has time to run
+        // and potentially crash if there's an issue.
+        systemConstraints.awaitUntil()
+
+        // Assert.
+        // If the test reaches here without crashing, it passes.
+        // This ensures that the feature group logic doesn't cause runtime exceptions
+        // even when high-end features are requested.
+        return@runBlocking
```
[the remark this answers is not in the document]
Zeronfinity: Done

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraSystem.kt
```diff
@@ -500,37 +531,350 @@ constructor(
...
+                backgroundDispatcher = defaultDispatcher,
+                screenFlashEvents = screenFlashEvents,
+                filePathGenerator = filePathGenerator,
+                focusMeteringEvents = focusMeteringEvents,
+                videoCaptureControlEvents = videoCaptureControlEvents,
+                currentCameraState = currentCameraState,
+                surfaceRequests = _surfaceRequest,
+                transientSettings = MutableStateFlow(transientSettings).asStateFlow(),
+                lowLightBoostEffectProvider = lowLightBoostEffectProvider
+            )
```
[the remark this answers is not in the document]
Zeronfinity: Done

core/camera/src/androidTest/java/com/google/jetpackcamera/core/camera/CameraXCameraSystemTest.kt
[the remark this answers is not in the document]
Zeronfinity: Done

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraSystem.kt
```diff
@@ -500,37 +537,370 @@ constructor(
...
+                backgroundDispatcher = defaultDispatcher,
+                screenFlashEvents = Channel(),
+                filePathGenerator = filePathGenerator,
+                focusMeteringEvents = Channel(),
+                videoCaptureControlEvents = Channel(),
+                currentCameraState = MutableStateFlow(CameraState()),
+                surfaceRequests = MutableStateFlow(null),
+                transientSettings = MutableStateFlow(transientSettings).asStateFlow(),
+                lowLightBoostEffectProvider = lowLightBoostEffectProvider
+            )
```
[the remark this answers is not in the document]
Zeronfinity: Done

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraSession.kt
```diff
@@ -596,26 +613,68 @@ internal fun createUseCaseGroup(
...
+                        it.feature.featureType == GroupableFeature.FEATURE_TYPE_RECORDING_QUALITY ->
+                            captureMode != CaptureMode.IMAGE_ONLY
+                        else -> true
+                    }
+                    if (shouldAdd) {
+                        add(it.feature)
+                    }
+                }
+                else -> {} // No-op.
+            }
```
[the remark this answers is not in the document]
Zeronfinity: We may need to add extra condition for other cases in future and converting from if to when again then will lead to more changes redundantly.

gradle/libs.versions.toml
```diff
@@ -1,6 +1,6 @@
 [versions]
 # Used directly in build.gradle.kts files
-compileSdk = "35"
+compileSdk = "36"
```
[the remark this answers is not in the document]
Zeronfinity: Doesn't seem relevant.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/FeatureGroupData.kt
```diff
@@ -0,0 +1,164 @@
...
+
+sealed interface FeatureGroupData<out T> {
+    /** Corresponds to a specific [GroupableFeature] object. */
+    data class ExplicitlyGroupable(val feature: GroupableFeature) : FeatureGroupData<Nothing>
+
+    /**
+     * Does not correspond to a specific [GroupableFeature] object, but inexplicitly groupable as
+     * it is equivalent to the base value CameraX feature groups API will use.
+     */
+    object InexplicitlyGroupable : FeatureGroupData<Nothing>
```
temcguir: nit: "Implicit" is more commonly used as an antonym with "Explicit". Could this be changed to "ImplicitlyGroupable"?

core/camera/src/main/java/com/google/jetpackcamera/core/camera/FeatureGroupData.kt
```diff
@@ -0,0 +1,164 @@
...
+import com.google.jetpackcamera.model.StabilizationMode.ON
+import com.google.jetpackcamera.model.StabilizationMode.OPTICAL
+import com.google.jetpackcamera.model.VideoQuality
+import com.google.jetpackcamera.model.VideoQuality.FHD
+import com.google.jetpackcamera.model.VideoQuality.HD
+import com.google.jetpackcamera.model.VideoQuality.SD
+import com.google.jetpackcamera.model.VideoQuality.UHD
+import com.google.jetpackcamera.model.VideoQuality.UNSPECIFIED
+
+sealed interface FeatureGroupData<out T> {
```
temcguir: nit: Maybe instead of "FeatureGroupData", this could be named "FeatureGroupability"?

temcguir: Can you add a quick kdoc to the top level to explain what this used for?

core/camera/src/main/java/com/google/jetpackcamera/core/camera/FeatureGroupData.kt
```diff
@@ -0,0 +1,164 @@
...
+ * Converts this [StabilizationMode] to a [GroupableFeatureData].
+ *
+ * This allows the stabilization mode to be used in CameraX feature group compatibility checks.
+ */
+fun StabilizationMode.toFeatureGroupData(): FeatureGroupData<StabilizationMode> {
+    return when (this) {
+        OFF -> InexplicitlyGroupable
+        ON -> ExplicitlyGroupable(GroupableFeature.PREVIEW_STABILIZATION)
+        HIGH_QUALITY -> ExplicitlyGroupable(GroupableFeatures.VIDEO_STABILIZATION)
+        AUTO, // AUTO should be resolved to concrete value before this API is called
```
temcguir: Throw an `IllegalStateException` rather than returning `Ungroupable` for `AUTO`. It's a programming error to try to call `toFeatureGroupData` without first resolving it to a concrete stabilization mode.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraSystem.kt
```diff
@@ -500,37 +553,360 @@ constructor(
...
+     *
+     * This function checks various combinations of settings (dynamic range, frame rate,
+     * stabilization, etc.) against the device's capabilities using the CameraX feature groups API.
+     * It filters out unsupported options from the system constraints, ensuring that the UI only
+     * presents valid combinations to the user.
+     *
+     * This update happens asynchronously after the initial camera session is started to avoid
+     * blocking the UI thread.
+     */
+    private suspend fun updateSystemConstraintsByFeatureGroups() {
```
temcguir, line 580: Is there any way we can move this to a seaparate file. CameraXCameraSystem is in dire need of refactoring, but I'd prefer not to add so many lines of code if possible.

Eventually I would like to move all of the constraints logic out, which may even help with consistency in the Settings screen.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraSystem.kt
```diff
@@ -944,23 +1370,39 @@ constructor(
...
-                ?.tryApplyImageFormatConstraints()
-                ?.tryApplyConcurrentCameraModeConstraints()
-                ?.tryApplyCaptureModeConstraints()
-                ?.tryApplyVideoQualityConstraints()
+            old?.copyStreamConfig(streamConfig = streamConfig)
         }
     }
 
+    /** Creates a copy of this [CameraAppSettings] with the provided [StreamConfig] applied. */
+    private fun CameraAppSettings.copyStreamConfig(
```
temcguir: Since CameraAppSettings is immutable, I think `copy` can be implied. Maybe rename this to `applyStreamConfig()`.

The same naming convention should be used with the other `copy` functions you've added.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraSession.kt
```diff
@@ -659,20 +718,25 @@ private fun createImageUseCase(
...
+    captureMode: CaptureMode,
+    backgroundDispatcher: CoroutineDispatcher,
+    targetFrameRate: Int? = null,
+    stabilizationMode: StabilizationMode? = null,
+    dynamicRange: DynamicRange? = null,
+    videoQuality: VideoQuality? = null
+): VideoCapture<Recorder>? {
+    if (captureMode != CaptureMode.STANDARD && captureMode != CaptureMode.VIDEO_ONLY) {
+        return null
+    }
```
temcguir, line 754: I suggest removing `captureMode` from this function and returning a non-nullable `VideoCapture<Recorder>`. The `captureMode` check should happen at the call site (similar to `createImageUseCase`) to avoid unnecessary argument passing and implicit null returns.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraSession.kt
```diff
@@ -743,8 +810,10 @@ private fun createPreviewUseCase(
     )
 }.build()
     .apply {
-        setSurfaceProvider { surfaceRequest ->
-            surfaceRequests.update { surfaceRequest }
+        withContext(Dispatchers.Main) {
+            setSurfaceProvider { surfaceRequest ->
+                surfaceRequests.update { surfaceRequest }
+            }
```
temcguir, line 845: I'm assuming this `withContext` was added so that this can be called from within the featuregroup check? Do we need to set a surface provider just to check for capabilities?

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraSystem.kt
```diff
@@ -500,37 +553,360 @@ constructor(
...
+                    videoQuality.takeIfFeatureGroupInvalid(sessionSettings)
+                )
+
+            createSessionConfig(
+                cameraConstraints = cameraConstraints,
+                initialTransientSettings = transientSettings,
+                videoCaptureUseCase = videoCaptureUseCase,
+                sessionSettings = sessionSettings,
+                sessionScope = CoroutineScope(defaultDispatcher + job)
+            )
```
temcguir: This seems quite expensive. Any time we want to call supportsGrouping, which happens when we change a setting, we set up and creating an entire pipeline (including use cases, effects, etc.). Maybe it's unavoidable, but do we really need to do this every time we change settings? Are there enough combinations of settings that we couldn't just do this once up front during initialization? Maybe there is, but I would much prefer exploring that approach if it meant we could avoid this type of check whenever we change a setting.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraSystem.kt
```diff
@@ -97,7 +97,7 @@ interface CameraSystem {
 
     fun getCurrentSettings(): StateFlow<CameraAppSettings?>
 
-    fun setFlashMode(flashMode: FlashMode)
+    suspend fun setFlashMode(flashMode: FlashMode)
```
temcguir: I'm not sure why we have so many setters here that are `suspend`, but most (all?) of them don't need to be, at least before this PR. We should be careful turning synchronous functions into asynchronous suspend functions. If one of these setters suspends for long enough, and another setter is called while it's suspending, you can run into a case where your suspend code will be run multiple times in order to ensure consistency (this is how CAS loops such as `MutableStateFlow.update` work).

Ideally, I would like for these setters to remain synchronous (non-suspend). They atomically update the CameraAppSettings, which also acts as our "prescriptive" UI settings, so it's important that they update quickly. If they have to suspend, we may need an intermediate state for CameraAppSettings that temporarily disables other settings in the UI while we're suspended.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraSystem.kt
```diff
@@ -911,7 +1329,7 @@ constructor(
     override fun getScreenFlashEvents() = screenFlashEvents
     override fun getCurrentSettings() = currentSettings.asStateFlow()
 
-    override fun setFlashMode(flashMode: FlashMode) {
+    override suspend fun setFlashMode(flashMode: FlashMode) {
```
temcguir: With this change, you're now calling a `suspend` function (`tryApplyDynamicRangeConstraints`) within a `MutableStateFlow.update{}` lambda, which internally uses a CAS loop. This could run multiple times depending on if the CameraAppSettings change during execution of the lambda, which is more likely if it takes longer, such as with a suspend function. Generally, you want to avoid any heavy computation within an `update` lambda.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraSession.kt
Zeronfinity: I am fine with changing it back to that, but note that there are multiple call sites for this and the `captureMode` checking logic is duplicate in all of those places then (3 places in total with this PR). I had moved `captureMode` inside the function mainly to avoid this duplicate checking in all 3 places. WDYT?

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraSession.kt
Zeronfinity: Short answer: not really.

As explained below, I implemented it this way for a few reasons. Regardless, this should be a JCA decision and I am happy to change this to however you prefer (or at least leave some TODO comments for later in case we decide to do something a bit more complicated).

1. I was trying to avoid different code flows for capability query and actual binding as much as possible.

Needing too many different code flow paths often leads to code smell and we've had developers complaining about similar situation recently as well. So, I tried to keep functions like `createPreviewUseCase` as generic as possible without having different code flows in these functions based on whether the intention is capability query or binding.

2. In future, we may use the user-provided surface in query directly (only if users have already provided the surface).

In some cases, the framework-layer FCQ API wants to have concrete surfaces, not deferred ones. Right now, CameraX creates such surfaces for queries, even when surface provider is already providing one. However, this may lead to different results in query and binding (in rare cases though), and there's also an optimization opportunity here. So, we may end up using the user-provided surface in future (only if already provided, it won't be a hard requirement). But this is kinda moot point for this PR since we are using `defaultCameraSessionContext` during query here which probably doesn't set up the `surfaceRequests` flow correctly anyway.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraSystem.kt
Zeronfinity: Removed the suspend modifier.

This was needed mainly because of doing a query call to know if HDR and JPEG_R supported together in `tryApplyDynamicRangeConstraints`. Cached the query call result when updating system constraints with feature groups results. So, all feature group results are calculated asynchronously now and used only when the result is available.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraSystem.kt
Zeronfinity: Done

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraSystem.kt
Zeronfinity: Removed the suspend modifier.

This was needed mainly because of doing a query call to know if HDR and JPEG_R supported together in `tryApplyDynamicRangeConstraints`. Cached the query call result when updating system constraints with feature groups results. So, all feature group results are calculated asynchronously now and used only when the result is available.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraSystem.kt
Zeronfinity: I initially hesitated with the apply... naming because the standard Kotlin apply scope function returns the same instance, which might be confusing for readers who aren't aware that CameraAppSettings is immutable.

However, given that the codebase already consistently uses the tryApply... pattern for similar configuration derivations, I agree that apply... also makes sense for consistency. so I’ve updated the naming as suggested.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/FeatureGroupData.kt
Zeronfinity: Sounds good. Done!

Zeronfinity: Done

core/camera/src/main/java/com/google/jetpackcamera/core/camera/FeatureGroupData.kt
Zeronfinity: Done

core/camera/src/main/java/com/google/jetpackcamera/core/camera/FeatureGroupData.kt
Zeronfinity: Done

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraSystem.kt
Zeronfinity: That is a valid concern regarding the overhead. The CameraX `isSessionConfigSupported` API for querying feature groups require a full SessionConfig because support can be influenced by session-wide variables like CameraEffect, use case resolutions, aspect ratios etc.

To ensure accuracy without assuming which scenarios might conflict in the JCA codebase (specially since I am not too well-versed with all scenarios in JCA and future plans), I have followed the framework's contract strictly for now. I agree that pre-calculation can be a great optimization since the number of factors that can change the result may not be too high for JCA, so I've added a TODO (in `FeatureGroupHandler.isGroupingSupported`) suggesting a persistence-based approach for the team to explore in future.

Does that work for this PR?

core/camera/src/main/java/com/google/jetpackcamera/core/camera/FeatureGroupHandler.kt
```diff
@@ -0,0 +1,474 @@
...
+                        initialCameraConstraints,
+                        cameraInfo
+                    ),
+                    supportedStreamConfigs = filterStreamConfig(
+                        currentSettings,
+                        initialSystemConstraints,
+                        initialCameraConstraints,
+                        cameraInfo
+                    )
+//                    unsupportedStabilizationFpsMap = unsupportedStabilizationFpsMap
```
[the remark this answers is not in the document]
Zeronfinity: It's part of a TODO comment mentioned above, so let's keep it here to help future devs.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/FeatureGroupHandler.kt
```diff
@@ -0,0 +1,474 @@
...
+//                        }
+//
+//                        if (!cameraAppSettings.copyStabilizationMode(stabilizationMode)
+//                                .copyTargetFrameRate(fps).isGroupingSupported(cameraInfo)
+//                        ) {
+//                            put(stabilizationMode, fpsList)
+//                        }
+//                    }
+//                }
+//        }
```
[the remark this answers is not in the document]
Zeronfinity: Fixed

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraSystem.kt
```diff
@@ -1025,10 +1208,30 @@ constructor(
...
+
+        return copy(
+            captureMode = captureMode,
+            dynamicRange =
+            if (isHdrOn) DynamicRange.HLG10 else DynamicRange.SDR,
+            imageFormat =
+            if (isHdrOn) ImageOutputFormat.JPEG_ULTRA_HDR else ImageOutputFormat.JPEG
+        )
+            .tryApplyDynamicRangeConstraints()
+            .tryApplyImageFormatConstraints()
```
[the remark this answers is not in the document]
Zeronfinity: Fixed

core/camera/src/androidTest/java/com/google/jetpackcamera/core/camera/CameraXCameraSystemTest.kt
```diff
@@ -344,6 +364,138 @@ class CameraXCameraSystemTest {
...
+        //  crashes due to some camera effect related surface not being cleaned up properly somehow.
+        //  This doesn't seem to be related to the primary purpose of this test, so simply excluding
+        //  it for now.
+        val featuresToTest = listOf(
+            DYNAMIC_RANGE_HLG10,
+            FPS_60,
+            VIDEO_QUALITY_UHD
+        )
+
+        featuresToTest.permutations().forEach { orderedFeatures ->
```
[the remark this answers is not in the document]
Zeronfinity: Left a TODO comment

core/camera/src/androidTest/java/com/google/jetpackcamera/core/camera/CameraXCameraSystemTest.kt
```diff
@@ -344,6 +364,138 @@ class CameraXCameraSystemTest {
...
+                        constraints
+                            ?.supportedStreamConfigs
+                            ?.contains(StreamConfig.SINGLE_STREAM) == true
+                    }
+                }
+            }
+        }
+    }
+
+    suspend fun <T> Deferred<T>.awaitUntil(timeout: Duration = 2.seconds): T {
```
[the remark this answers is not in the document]
Zeronfinity: Done

core/camera/src/androidTest/java/com/google/jetpackcamera/core/camera/CameraXCameraSystemTest.kt
```diff
@@ -465,6 +618,71 @@ class CameraXCameraSystemTest {
             )
         )
     }
+
+    suspend fun <T> Feature.tryApplyFeature(
```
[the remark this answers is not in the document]
Zeronfinity: Done

core/camera/src/androidTest/java/com/google/jetpackcamera/core/camera/CameraXCameraSystemTest.kt
```diff
@@ -465,6 +618,71 @@ class CameraXCameraSystemTest {
...
+
+        // Verify feature is set according to current settings
+        assertThat(getNewFeatureValue(cameraSystem.getCurrentSettings().value)).isEqualTo(
+            expectedValue
+        )
+
+        return newConstraints
+    }
+
+    fun <T> List<T>.permutations(): List<List<T>> {
```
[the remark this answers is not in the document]
Zeronfinity: Done

core/camera/src/androidTest/java/com/google/jetpackcamera/core/camera/CameraXCameraSystemTest.kt
```diff
@@ -465,6 +618,71 @@ class CameraXCameraSystemTest {
...
+            for (i in 0..permOfTail.size) {
+                val newPerm = permOfTail.toMutableList()
+                newPerm.add(i, head)
+                result.add(newPerm)
+            }
+        }
+        return result
+    }
+
+    enum class Feature {
```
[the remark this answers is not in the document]
Zeronfinity: Done

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraSystem.kt
```diff
@@ -1006,17 +1173,33 @@ constructor(
...
+            old?.applyStabilizationMode(stabilizationMode = stabilizationMode)
         }
     }
 
+    /** Returns a new [CameraAppSettings] with the provided [StabilizationMode] applied. */
+    internal fun CameraAppSettings.applyStabilizationMode(
+        stabilizationMode: StabilizationMode
+    ): CameraAppSettings {
+        return copy(stabilizationMode = stabilizationMode)
+    }
```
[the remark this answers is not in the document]
Zeronfinity: This requires cleaning up tryApplyStabilizationConstraints carefully and not too important right now, so let's leave it for now.

Zeronfinity:
Closing this PR in favor of #522. We have moved this work to a new PR based on a branch directly on the upstream repository. A new PR was created since Github doesn't support changing the source repository of existing PR.

PR #459, opened 2025-12-30
Add ContentResolver to ImagePostProcessor.processImage()

davidjiagoogle:
As titled. Any implementations of such ImagePostProcessor likely needs access to the contentResolver that is used during the image capture process. Adding it as an input parameter to processImage()


temcguir:
Please resolve the @gemini-code-assist comments.

temcguir:
The implementation LGTM, but can you add some tests that use the new ContentResolver argument? Maybe @gemini-code-assist could suggest some tests that could be added.

temcguir (approved):

PR #461, opened 2026-01-14
Update Capture Button visuals

temcguir:
This PR updates the appearance and behavior of the `CaptureButton` component to better match design mocks and improve preview visibility.

### Key Changes
* **Visual Updates:**
  * Added support for multiple background styles via `LocalShutterBackgroundStyle` (`WHITE_20` and `BLACK_60`).
  * Adjusted the size of the capture button, ring stroke width, and nucleus to align with mocks.
  * Fixed alignment of buttons in previews to ensure they are not cut off.

image

* **Accessibility and Localization:**
  * Made the `CaptureButton` focusable and added appropriate roles and content descriptions.
  * Moved content descriptions to `strings.xml` for proper localization support.
  * Enabled automated accessibility checks in tests.

* **Refactoring and Testing:**
  * Refactored pressed state previews to use `CompositionLocalProvider(LocalInitialPressedState provides true)`, avoiding manual construction of complex UI states and ensuring accurate rendering in previews.
  * Added screenshot tests for various states (Idle, Pressed, Disabled, Recording) across different capture modes and background styles.
  * Consolidated previews in `CaptureButtonScreenshotTest.kt` for better maintainability.


ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/CaptureButtonComponents.kt
```diff
@@ -637,20 +698,80 @@ private fun CaptureButtonNucleus(
...
-    val animatedColor by animateColorAsState(
-        targetValue = when (val uiState = currentUiState.value) {
-            is CaptureButtonUiState.Available.Idle -> when (uiState.captureMode) {
-                CaptureMode.STANDARD -> imageCaptureModeColor
-                CaptureMode.IMAGE_ONLY -> imageCaptureModeColor
-                CaptureMode.VIDEO_ONLY -> recordingColor
+    val isPressableImageMode = currentUiState.value.let {
+        it is CaptureButtonUiState.Available.Idle &&
+            (it.captureMode == CaptureMode.IMAGE_ONLY || it.captureMode == CaptureMode.STANDARD)
+    }
```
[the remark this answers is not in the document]
temcguir: @gemini-code-assist I think it's debatable whether this is more readable. It's certainly longer (including an explicit `else -> false`). I think I prefer folding the `is` check into the boolean statement. Can you give a strong reason to prefer one way over the other?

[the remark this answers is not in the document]
temcguir: > Exhaustiveness and Future-Proofing: The when expression, especially when used with sealed classes like CaptureButtonUiState, encourages exhaustive checking. The else -> false explicitly states that any other CaptureButtonUiState or CaptureMode combination will result in false. If new CaptureButtonUiState types are introduced in the future, the compiler will prompt for an update to the when expression, ensuring that the logic for isPressableImageMode is always explicitly handled. In contrast, the let block might silently fail or produce unexpected behavior if new states are added and not explicitly considered.

@gemini-code-assist, I don't think that's correct. The inclusion of an `else` statement in the `when` block will implicitly handle any new `CaptureButtonUiState` types. We would need to remove the `else` statement, and exhaustively list all of the types and return false if we wanted this to be true. In that case, it seems like more work than it's worth. 

> Clarity of Intent: The when expression clearly separates the different states and their corresponding conditions. It reads as "when the UI state is X, then check Y; otherwise, it's Z." This can be easier to parse mentally than a single boolean expression with multiple && and || operators, especially as the conditions grow.

That's fair. However, because the number of conditions is small now, I don't think we need to change it. If the condition grows, that would be a good time to change it to a `when` block.

> Consistency: If other parts of the codebase use when expressions for similar state-based logic, using it here maintains consistency, which also contributes to overall readability.

No argument that we like to use `when` blocks, but this is a case where it is concise enough on its own that I don't think its necessary.

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/CaptureButtonComponents.kt
```diff
@@ -697,176 +808,173 @@ private fun CaptureButtonNucleus(
...
-    CaptureButton(
-        onImageCapture = {},
-        onStartRecording = {},
-        onStopRecording = {},
-        onLockVideoRecording = {},
-        onIncrementZoom = {},
+    PreviewCaptureButton(
         captureButtonUiState = CaptureButtonUiState.Unavailable
     )
 }
```
Kimblebee: nit: this `@Preview` function should be moved below `PreviewCaptureButton`

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/CaptureButtonComponents.kt
```diff
@@ -637,20 +698,80 @@ private fun CaptureButtonNucleus(
...
+    }
+
+    val transition =
+        updateTransition(targetState = nucleusState, label = "Nucleus Color Transition")
+    val animatedColor by transition.animateColor(
+        label = "Nucleus Color",
+        transitionSpec = {
+            when {
+                NucleusState.Disabled isTransitioningTo NucleusState.Idle -> tween(
+                    durationMillis = 300
```
Kimblebee: these `durationMillis` values could be constants?

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/CaptureButtonComponents.kt
```diff
@@ -637,20 +698,80 @@ private fun CaptureButtonNucleus(
...
+            },
+        label = "Press Size Transition"
+    )
+
+    val centerShapeSize by pressTransition.animateDp(
+        transitionSpec = {
+            if (targetState) {
+                snap()
+            } else {
+                tween(durationMillis = 200)
```
Kimblebee: another constant `durationMillis`?

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/CaptureButtonComponents.kt
```diff
@@ -333,13 +367,19 @@ private fun CaptureButton(
...
+
     val animatedColor by animateColorAsState(
         targetValue = if (isVisuallyDisabled) {
             LocalContentColor.current.copy(alpha = 0.38f)
         } else {
             LocalContentColor.current
         },
-        animationSpec = tween(durationMillis = if (isVisuallyDisabled) 1000 else 300),
+        animationSpec = if (isVisuallyDisabled) {
+            tween(durationMillis = 1000)
```
Kimblebee: more magic numbers to be moved to constants?

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/CaptureButtonComponents.kt
```diff
@@ -637,20 +698,80 @@ private fun CaptureButtonNucleus(
...
+                )
+                NucleusState.Pressed isTransitioningTo NucleusState.Idle -> tween(
+                    durationMillis = 100
+                )
+                else -> snap()
             }
+        }
+    ) { state ->
+        when (state) {
+            NucleusState.Disabled -> LocalContentColor.current.copy(alpha = 0.38f)
```
Kimblebee: repeated alpha value `.38f` to be moved to a constant?

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/CaptureButtonComponents.kt
```diff
@@ -474,10 +520,15 @@ fun CaptureButtonRing(
...
     color: Color,
-    borderWidth: Float = 4f,
+    borderWidth: Float = 3f,
     contents: (@Composable () -> Unit)? = null
 ) {
     Box(modifier = modifier, contentAlignment = Alignment.Center) {
+        Box(
+            modifier = Modifier
+                .size(captureButtonSize.dp)
+                .background(Color.Black.copy(alpha = 0.5f), CircleShape)
```
Kimblebee: should this unique alpha value be a constant?

Kimblebee (commented):
some duplicate comments regarding readibility, otherwise I think the PR looks great

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/CaptureButtonComponents.kt
```diff
@@ -683,34 +915,42 @@ private fun CaptureButtonNucleus(
...
-                        if (isPressed &&
-                            currentUiState.value ==
-                            CaptureButtonUiState.Enabled.Idle(CaptureMode.IMAGE_ONLY)
-                        ) {
-                            .5f // transparency to indicate click ONLY on IMAGE_ONLY
-                        } else {
-                            1f // solid alpha the rest of the time
-                        }
-                    )
+                    .clip(androidx.compose.foundation.shape.RoundedCornerShape(cornerRadius))
```
Kimblebee: can import this at the top of the file

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/CaptureScreenComponents.kt
```diff
@@ -478,7 +478,7 @@ fun PreviewDisplay(
 
     surfaceRequest?.let {
         BoxWithConstraints(
-            modifier
+            Modifier
```
Kimblebee: why don't we use the passed modifier value here? and the fresh modifier in the inner box? Or is this intentional

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/CaptureButtonComponents.kt
```diff
@@ -582,27 +718,40 @@ private fun LockSwitchCaptureButtonNucleus(
...
+            Box(
                 modifier = Modifier
-                    .size(switchHeight * .75f)
                     .align(Alignment.CenterStart)
                     .padding(start = 8.dp)
                     .offset(x = -(switchWidth - pressedNucleusSize))
-                    .clickable(indication = null, interactionSource = null) {
-                        onToggleSwitchPosition()
+                    .size(32.dp)
+                    .semantics { contentDescription = "Lock Video Recording" }
```
Kimblebee: hardcoded string

Kimblebee: should add `role = Role.Button` for this custom button.

that or we could turn this into a button/iconbutton. instead of just an Icon

gradle/libs.versions.toml
```diff
@@ -86,8 +88,11 @@ camera-video = { module = "androidx.camera:camera-video", version.ref = "android
 camera-compose = { module = "androidx.camera:camera-compose", version.ref = "androidxCamera" }
 compose-bom = { module = "androidx.compose:compose-bom", version.ref = "composeBom" }
 compose-junit = { module = "androidx.compose.ui:ui-test-junit4" }
+compose-accessibility = { module = "androidx.compose.ui:ui-test-junit4-accessibility", version = "1.8.0-beta03" }
+accessibility-test-framework = { module = "com.google.android.apps.common.testing.accessibility.framework:accessibility-test-framework", version = "4.1.1" }
```
Kimblebee: should define the versions for these under `[versions]` and reference them here instead

ui/components/capture/build.gradle.kts
```diff
@@ -116,3 +124,8 @@ dependencies {
 kapt {
     correctErrorTypes = true
 }
+configurations.all {
+    resolutionStrategy {
+        exclude(group = "com.google.protobuf", module = "protobuf-lite")
+    }
+}
```
Kimblebee, line 140: can we include a comment why this is necessary

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/CaptureButtonComponents.kt
temcguir: Done. I extracted the string into `strings.xml` as `capture_button_lock_video_recording` and dynamically referenced it here using `stringResource()`.

temcguir: Done. I've appended `role = Role.Button` to the semantics tag for this.

gradle/libs.versions.toml
temcguir: Done. I removed the hardcoded version for `compose-accessibility` so it now delegates properly to the `compose-bom`. Then I created an `accessibilityTestFramework` variable in the `[versions]` block with `version.ref` for the `accessibility-test-framework` dependency.

ui/components/capture/build.gradle.kts
temcguir: Done! I've added a comment above the `configurations.all` block. It is necessary because the `accessibility-test-framework` brings in `protobuf-lite`, which conflicts directly with the `protobuf-javalite` dependency that `androidx.datastore` relies on, throwing a `DuplicateClassException`. Excluding `protobuf-lite` here forces it to use the expected resolution strategy.

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/CaptureButtonComponents.kt
temcguir: Done. I replaced the fully-qualified `androidx.compose.foundation.shape.RoundedCornerShape` with just `RoundedCornerShape` and added the appropriate import at the top of the file.

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/CaptureScreenComponents.kt
temcguir: Nice catch. That was unintentional. I've updated it so the passed `modifier` applies to the root layout as expected, and a fresh `Modifier` is used for the inner `Box`.

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/CaptureButtonComponents.kt
Kimblebee, line 1: Since `switchPosition` is updated at high frequency during drag gestures, we should pass it in as a lambda instead of the computed `Dp` offset. Pass the position as a lambda provider to defer state reads and skip recomposition. Inside `LockSwitchCaptureButtonNucleus`, apply the offset to `CaptureButtonNucleus` using a lambda-based `Modifier.offset` instead of passing the computed `Dp` offset value as a parameter:                      
                                                                                                                                                       
    1. **Update the signature of `LockSwitchCaptureButtonNucleus`** to accept a lambda:                                                                
    ```kotlin                                                                                                                                          
    private fun LockSwitchCaptureButtonNucleus(                                                                                                        
        ...                                                                                                                                            
        switchPositionProvider: () -> Float, // Changed from Float                                                                                     
        ...                                                                                                                                            
    )                                                                                                                                                  
                                                                                                                                                       
  2. Update the call inside LockSwitchCaptureButtonNucleus to use lambda-based offset and remove offsetX parameter from CaptureButtonNucleus:          
                                                                                                                                                       
            CaptureButtonNucleus(                                                                                                                      
                modifier = Modifier.offset {                                                                                                           
                    IntOffset(                                                                                                                         
                        x = (-(switchWidth - pressedNucleusSize).toPx() * switchPositionProvider()).roundToInt(),
                        y = 0
                    )        
                },   
                captureButtonSize = captureButtonSize,
                captureButtonUiState = captureButtonUiState,
                pressedVideoCaptureScale = LOCK_SWITCH_PRESSED_NUCLEUS_SCALE,
                isTapping = isTapping,
                isVisuallyDisabled = isVisuallyDisabled
            )
             
  3. Update the caller in CaptureButton (line 592) to pass the state as a lambda:
  
                LockSwitchCaptureButtonNucleus(
                    ...
                    switchPositionProvider = { switchPosition },
                    ...
                )      
                 
  4. Update the calls in the preview composables (around L1117 and L1148) to pass lambda providers:
  
    // In LockSwitchLockedAtThresholdPressedRecordingPreview:
                LockSwitchCaptureButtonNucleus(
                    ...
                    switchPositionProvider = { MINIMUM_LOCK_THRESHOLD },
                    ...
                )      
                 
    // In LockSwitchLockedPressedRecordingPreview:
                LockSwitchCaptureButtonNucleus(   
                    ...
                    switchPositionProvider = { 1f },
                    ...
                )

Kimblebee, line 515: the `onRelease` callbacks should be mutually exclusive.
```suggestion
if (shouldBeLocked()) {
onLockVideoRecording(true)
onRelease(true)
}

else {
onRelease(false)
}

switchPosition = LOCK_SWITCH_POSITION_OFF
}
```

Kimblebee (approved):

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/CaptureButtonComponents.kt
temcguir: Done. Thanks for the suggestion.

temcguir: I've updated the component signatures and implemented the lambda-based offset using `Modifier.offset { IntOffset(...) }` as suggested.

PR #463, opened 2026-01-20
Preview stabilisation fix

davidjiagoogle:
A bug was discovered where preview usecase retains the "ON" value for stabilization after video stabilization is set to true for video recording than switching to image capture mode.

This was caused by the when statement in createPreviewUseCase() not calling setPreviewStabilizationEnabled(false) in the StabilizationMode.OFF condition branch.

I made it so setPreviewStabilizationEnabled(false) and setOpticalStabilizationModeEnabled(false) are called in all 3 other branches


temcguir:
It looks like we were missing some tests for stabilization mode from the beginning. Maybe some tests in CameraXCameraSystemTest would be good. If you look at [this line](https://github.com/google/jetpack-camera-app/blob/6c9df1d1783d844c871ec0bd98547d5d02157169/core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraSession.kt#L1375), you'll see that the `CameraState` always gets the stabilization mode from the `TotalCaptureResult`, so we should be able to cycle through all of the supported stabilization modes and test that each of them gets applied correctly. @gemini-code-assist, can you suggest tests for this? The tests should 1. Check that the stabilization mode is supported, 2. Apply the stabilization mode, and 3. Wait on the CameraState to verify that it eventually changes to the requested stabilization mode.

temcguir:
@gemini-code-assist, `CameraState` already has a `stabilizationMode` property, and `CameraSystem` already has a function which returns a `StateFlow<CameraState>` called `getCurrentCameraState()` and has a `setStabilizationMode` function. The supported stabilization modes can be retrieved via `SettableConstraintsRepository` that can be passed into `CameraXCameraSystemTest.createAndInitCameraXCameraSystem())`. There shouldn't be any need to add helper functions to non-test code as you suggsted.

core/camera/src/androidTest/java/com/google/jetpackcamera/core/camera/CameraXCameraSystemTest.kt
```diff
@@ -250,16 +250,16 @@ class CameraXCameraSystemTest {
         val appSettings = DEFAULT_CAMERA_APP_SETTINGS
         val cameraConstraints =
             constraintsRepository.systemConstraints.value?.forCurrentLens(appSettings)
```
temcguir: This will still skip the test. You need to move the initialization before this line.

core/camera/src/androidTest/java/com/google/jetpackcamera/core/camera/CameraXCameraSystemTest.kt
```diff
@@ -225,6 +227,70 @@ class CameraXCameraSystemTest {
...
+        val recordingComplete = CompletableDeferred<Unit>()
+        cameraSystem.startRecording {
+            when (it) {
+                is OnVideoRecorded -> {
+                    recordingComplete.complete(Unit)
+                }
+
+                is OnVideoRecordError -> recordingComplete.completeExceptionally(it.error)
+            }
+        }
```
temcguir, line 280: We shouldn't need to create a recording to test this.

core/camera/src/androidTest/java/com/google/jetpackcamera/core/camera/CameraXCameraSystemTest.kt
```diff
@@ -225,6 +227,70 @@ class CameraXCameraSystemTest {
...
+        stabilizationCheck.awaitValue(true)
+
+        cameraSystem.stopVideoRecording()
+
+        stabilizationCheck.awaitValue(true)
+
+        // Clean-up.
+        recordingComplete.await()
+        stabilizationCheck.cancel()
+    }
```
temcguir, line 292: You should be able to replace all of this code with:
```
        val stabilizationCheck: ReceiveChannel<StabilizationMode> = cameraSystem.getCurrentCameraState()
            .map { it.stabilizationMode }
            .produceIn(this)
        cameraSystem.startCameraAndWaitUntilRunning()

        // Act.
        cameraSystem.setStabilizationMode(stabilizationMode)

        // Assert.
        stabilizationCheck.awaitValue(stabilizationMode)

        // Clean-up.
        stabilizationCheck.cancel()
```

and it should pass. However, it doesn't seem to pass for HIGH_QUALITY and OPTICAL, with or without your other changes. I think there is something else wrong.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraSession.kt
```diff
@@ -748,13 +748,24 @@ private fun createPreviewUseCase(
...
+            setPreviewStabilizationEnabled(false)
+            setOpticalStabilizationModeEnabled(true)
+        }
+        StabilizationMode.OFF -> {
+            setPreviewStabilizationEnabled(false)
+            setOpticalStabilizationModeEnabled(false)
+        }
+        StabilizationMode.HIGH_QUALITY -> {
+            // No-op. Handled by VideoCapture use case.
+        }
```
temcguir, line 768: It shouldn't be necessary to set all of these. This is creating a new preview use case, so the absence of setting preview stabilization to `true` should be sufficient. Also, setting optical stabilization to `false` when preview stabilization is `true` is not the right thing to do. There is nothing that says optical stabilization needs to be disabled when preview stabilization is enabled. We should leave it to the camera on whether to enable or disable optical stabilization.

PR #470, opened 2026-02-06
add documentation to :ui:components:capture

Kimblebee:
Also deleted `Theme.kt` and `MultipleEventsCutter.kt` from the module.


ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/theme/Theme.kt
```diff
@@ -25,6 +25,15 @@ import androidx.compose.material3.lightColorScheme
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.platform.LocalContext
 
+/**
+ * A composable that provides a Material Design theme for previews within the Jetpack Camera App.
+ *
+ * This theme adapts to system dark mode settings and supports dynamic colors on Android 12+.
```
temcguir: nit: This is technically true, but I don't think we'll need to support the light theme in the future. No need to change anything now, except for maybe adding a TODO to remove the ability to change the theme.

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/CaptureButtonComponents.kt
```diff
@@ -155,6 +155,18 @@ private fun CaptureKeyHandler(
...
 
+/**
+ * A capture button that can be used for both image and video capture.
+ *
+ * @param modifier the modifier for this component
+ * @param onImageCapture the callback for an image capture event
+ * @param onStartRecording the callback for a start recording event
+ * @param onStopRecording the callback for a stop recording event
+ * @param onLockVideoRecording The callback for a lock video recording event. The boolean parameter indicates if the recording should be locked.
+ * @param onIncrementZoom The callback for a zoom increment event, providing the zoom increment value.
```
temcguir, line 179: nit: explain why there is a zoom option here. It seems out of place without explanation.

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/EventDebouncer.kt
```diff
@@ -16,14 +16,20 @@
 package com.google.jetpackcamera.ui.components.capture
 
 /**
- * A helper class that prevents multiple clicks.
+ * A helper class that debounces events, preventing multiple rapid-fire executions.
  */
-internal class MultipleEventsCutter {
+internal class EventDebouncer {
```
temcguir: nit: I don't think this class is used currently. It can probably be deleted.

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/ZoomState.kt
temcguir: It looks like this class is missing documentation.

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/CaptureButtonComponents.kt
```diff
@@ -155,6 +155,18 @@ private fun CaptureKeyHandler(
     }
 }
 
+/**
+ * A capture button that can be used for both image and video capture.
```
temcguir: This description seems a little sparse. What are the different modes it can be in? What about the lock behavior?

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/CaptureScreenComponents.kt
```diff
@@ -671,6 +733,14 @@ fun VideoQualityIcon(videoQuality: VideoQuality, modifier: Modifier = Modifier)
     }
 }
 
+/**
+ * A composable that displays the flip camera button.
```
temcguir: There are a few composables that are documented like this. I think you should add a little more about the behavior of each composable. This description is already basically captured by virtue of the signature:
```
@Composable
fun FlipCameraButton(
```

temcguir (approved):

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/theme/Theme.kt
Kimblebee: This was actually only used in the composables with the `@Preview` annotation. This should honestly just be deleted.

PR #476, opened 2026-02-13
Refactor: Decouple DataStore from Settings Module

Kimblebee:
Extracted Proto DataStore implementation into a new `:data:settings-datastore` module. This improves modularity and separation of concerns by isolating the persistence layer. 

`:data:settings` now solely defines interfaces and handles constraints.


davidjiagoogle (approved):

davidjiagoogle:
In https://github.com/google/jetpack-camera-app/pull/525, datastore dependency is looking to be completely removed and replaced with sharedprefs. This PR achieves similar result by splitting datastore dep into a decoupled module and offloading the data storing implementation to the app layer if it doesn't want to use DataStore. We should discuss which approach we want to use

PR #482, opened 2026-02-20
Moving APIs in PreviewViewModel to Controller classes

davidjiagoogle:
Interfaces QuickSettingsController, CaptureController, CaptureScreenController, SnackBarController, ZoomController, and DebugController are created. They contain APIs that were originally implemented in the view models. A default implementation of their APIs is made for each interface so that view model can be simplified

PreviewViewModel now only contains instances of these controller and has no APIs itself.


ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/quicksettings/QuickSettingsScreen.kt
```diff
@@ -339,17 +330,30 @@ fun ExpandedQuickSettingsUiPreview_WithHdr() {
...
-            toggleQuickSettings = { },
             onNavigateToSettings = { },
-            onCaptureModeClick = { },
-            onSetFocusedSetting = {}
+            quickSettingsController = MockQuickSettingsController()
         )
     }
 }
+
+class MockQuickSettingsController : QuickSettingsController {
```
temcguir: nit: Call this "NoOpQuickSettingsController". We should follow this naming convention for any of the `@Preview` composables..

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/controller/Utils.kt
```diff
@@ -0,0 +1,77 @@
...
+object Utils {
+    /**
+     * Posts the given [MediaDescriptor] to the [MediaRepository] as the current media.
+     *
+     * @param viewModelScope The [CoroutineScope] for launching the coroutine.
+     * @param mediaRepository The repository to update.
+     * @param mediaDescriptor The media to set as current.
+     */
+    fun postCurrentMediaToMediaRepository(
+        viewModelScope: CoroutineScope,
```
temcguir: Instead of passing in a viewModelScope, you should just make this a suspend function and have the `viewModelScope.launch` within the ViewModel wrapping this suspend function.

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/controller/CaptureScreenController.kt
```diff
@@ -0,0 +1,62 @@
...
+     * Transfers the media from the image well to the repository.
+     */
+    fun imageWellToRepository()
+
+    /**
+     * Pauses or resumes video recording.
+     *
+     * @param shouldBePaused Whether the recording should be paused.
+     */
+    fun setPaused(shouldBePaused: Boolean)
```
temcguir: This should be in CaptureController.

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/controller/CaptureScreenController.kt
```diff
@@ -0,0 +1,62 @@
...
+     * @param y The y-coordinate of the tap.
+     */
+    fun tapToFocus(x: Float, y: Float)
+
+    /**
+     * Enables or disables audio recording.
+     *
+     * @param shouldEnableAudio Whether audio should be enabled.
+     */
+    fun setAudioEnabled(shouldEnableAudio: Boolean)
```
temcguir: This should be in CaptureController.

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/controller/CaptureScreenController.kt
```diff
@@ -0,0 +1,62 @@
...
+
+    /**
+     * Updates the UI with the last captured media.
+     */
+    fun updateLastCapturedMedia()
+
+    /**
+     * Transfers the media from the image well to the repository.
+     */
+    fun imageWellToRepository()
```
temcguir: Can these be moved to an ImageWellController?

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/controller/CaptureScreenController.kt
```diff
@@ -0,0 +1,62 @@
...
+ * limitations under the License.
+ */
+package com.google.jetpackcamera.ui.components.capture.controller
+
+import com.google.jetpackcamera.model.DeviceRotation
+
+/**
+ * Interface for controlling actions on the capture screen.
+ */
+interface CaptureScreenController {
```
temcguir: If we move the functions out of this class that I've recommended, then we're left with setDisplayRotation() and tapToFocus(). I think this interface could then be renamed to "ViewfinderController".

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/controller/CaptureControllerImpl.kt
```diff
@@ -0,0 +1,288 @@
...
+ */
+class CaptureControllerImpl(
+    private val trackedCaptureUiState: MutableStateFlow<TrackedCaptureUiState>,
+    private val viewModelScope: CoroutineScope,
+    private val cameraSystem: CameraSystem,
+    private val mediaRepository: MediaRepository,
+    private val saveMode: SaveMode,
+    private val externalCaptureMode: ExternalCaptureMode,
+    private val externalCapturesCallback: () -> Pair<SaveLocation, IntProgress?>,
+    private val captureEvents: Channel<CaptureEvent>,
```
temcguir: Does this need to be an input to the constructor? Can it just be a val on CaptureController? (likely as a ReceiveChannel in the interface).

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/controller/CaptureControllerImpl.kt
```diff
@@ -0,0 +1,288 @@
...
+ * @property saveMode Mode for saving captured media.
+ * @property externalCaptureMode Mode for external capture requests.
+ * @property externalCapturesCallback Callback for getting external capture information.
+ * @property captureEvents Channel for sending capture-related events.
+ * @property captureScreenController Controller for UI-related capture screen actions.
+ * @property snackBarController Controller for showing snackbars.
+ */
+class CaptureControllerImpl(
+    private val trackedCaptureUiState: MutableStateFlow<TrackedCaptureUiState>,
+    private val viewModelScope: CoroutineScope,
```
temcguir: It's generally considered bad practice to pass in CoroutineScope to constructors. At a minimum this should be a CoroutineContext from which we create a scope within the controller. See go/android-api-guidelines-async#coroutines-constructor-context for more info.

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/controller/CaptureScreenController.kt
```diff
@@ -0,0 +1,62 @@
...
+     */
+    fun setDisplayRotation(deviceRotation: DeviceRotation)
+
+    /**
+     * Initiates a tap-to-focus action at the given coordinates.
+     *
+     * @param x The x-coordinate of the tap.
+     * @param y The y-coordinate of the tap.
+     */
+    fun tapToFocus(x: Float, y: Float)
```
temcguir: These can be moved into CameraController.

ui/controller/impl/src/main/java/com/google/jetpackcamera/ui/controller/impl/CaptureControllerImpl.kt
```diff
@@ -0,0 +1,288 @@
...
+ * @property externalCapturesCallback Callback for getting external capture information.
+ * @property captureEvents Channel for sending capture-related events.
+ * @property captureScreenController Controller for UI-related capture screen actions.
+ * @property snackBarController Controller for showing snackbars.
+ */
+class CaptureControllerImpl(
+    private val trackedCaptureUiState: MutableStateFlow<TrackedCaptureUiState>,
+    private val viewModelScope: CoroutineScope,
+    private val cameraSystem: CameraSystem,
+    private val mediaRepository: MediaRepository,
```
temcguir, line 77: It would be good if we could remove the dependency on MediaRepository here. We should only have a MediaRepository dependency in the ViewModel. Maybe this can be replaced with a callback that can be supplied at creation time which will call the MediaRepository within the ViewModel instead of within the controller.

temcguir (changes requested):
This is a good start. There are definitely a few things that could be cleaned up later (such as the zoom controller). If you can address the comments I've added we could merge this and make smaller clean-up PRs to address those later.

ui/controller/impl/src/main/java/com/google/jetpackcamera/ui/controller/impl/SnackBarControllerImpl.kt
```diff
@@ -20,18 +20,28 @@ import com.google.jetpackcamera.ui.uistate.DisableRationale
...
+ * @param snackBarUiState The [MutableStateFlow] of the [SnackBarUiState.Enabled] to update.
+ * @param coroutineContext The [CoroutineContext] to use for launching coroutines.
+ */
 class SnackBarControllerImpl(
-    private val viewModelScope: CoroutineScope,
-    private val snackBarUiState: MutableStateFlow<SnackBarUiState.Enabled>
+    private val snackBarUiState: MutableStateFlow<SnackBarUiState.Enabled>,
+    coroutineContext: CoroutineContext
 ) : SnackBarController {
+    private val job = Job(parent = coroutineContext[Job])
```
temcguir, line 46: Technically, we need `close()`, `cancel()`, and optionally, `join()` functions on the `SnackBarControllerImpl` (not on the `SnackBarController` interface), so that any launched work within this controller can be canceled. When using a ViewModelScope's context, it won't be a big deal, but if it's used outside of a ViewModelScope, such as in tests, it may be necessary. See go/android-api-guidelines-async#coroutines-close-cancel.

ui/controller/impl/src/main/java/com/google/jetpackcamera/ui/controller/impl/CameraControllerImpl.kt
```diff
@@ -37,21 +37,22 @@ private const val TAG = "CameraControllerImpl"
...
     private val initializationDeferred: Deferred<Unit>,
     private val captureUiState: StateFlow<CaptureUiState>,
-    private val viewModelScope: CoroutineScope,
-    private val cameraSystem: CameraSystem
+    private val cameraSystem: CameraSystem,
+    coroutineContext: CoroutineContext
 ) : CameraController {
     private var runningCameraJob: Job? = null
-
+    private val job = Job(parent = coroutineContext[Job])
```
temcguir, line 54: Technically, we need `close()`, `cancel()`, and optionally, `join()` functions on the `CameraControllerImpl` (not on the `CameraController` interface), so that any launched work within this controller can be canceled. When using a ViewModelScope's context, it won't be a big deal, but if it's used outside of a ViewModelScope, such as in tests, it may be necessary. See go/android-api-guidelines-async#coroutines-close-cancel.

ui/controller/impl/src/main/java/com/google/jetpackcamera/ui/controller/impl/CaptureControllerImpl.kt
```diff
@@ -56,33 +57,35 @@ private const val IMAGE_CAPTURE_TRACE = "JCA Image Capture"
...
     private val captureScreenController: CaptureScreenController,
-    private val snackBarController: SnackBarController?
+    private val snackBarController: SnackBarController?,
+    coroutineContext: CoroutineContext
 ) : CaptureController {
 
     private val traceCookie = atomic(0)
     private val videoCaptureStartedCount = atomic(0)
     private var recordingJob: Job? = null
+    private val job = Job(parent = coroutineContext[Job])
```
temcguir, line 90: Technically, we need `close()`, `cancel()`, and optionally, `join()` functions on the `CaptureControllerImpl` (not on the `CaptureController` interface), so that any launched work within this controller can be canceled. When using a ViewModelScope's context, it won't be a big deal, but if it's used outside of a ViewModelScope, such as in tests, it may be necessary. See go/android-api-guidelines-async#coroutines-close-cancel.

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/controller/ImageWellControllerImpl.kt
```diff
@@ -1,4 +1,46 @@
...
+ * Implementation of [ImageWellController] that handles image well actions.
+ *
+ * @param mediaRepository The [MediaRepository] for accessing media.
+ * @param coroutineContext The [CoroutineContext] for launching coroutines.
+ */
+class ImageWellControllerImpl(
+    private val mediaRepository: MediaRepository,
+    coroutineContext: CoroutineContext
+) : ImageWellController {
+    private val job = Job(parent = coroutineContext[Job])
```
temcguir: Technically, we need `close()`, `cancel()`, and optionally, `join()` functions on the `ImageWellControllerImpl` (not on the `ImageWellController` interface), so that any launched work within this controller can be canceled. When using a ViewModelScope's context, it won't be a big deal, but if it's used outside of a ViewModelScope, such as in tests, it may be necessary. See go/android-api-guidelines-async#coroutines-close-cancel.

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/quicksettings/controller/QuickSettingsControllerImpl.kt
```diff
@@ -36,17 +38,20 @@ import kotlinx.coroutines.launch
...
  */
 class QuickSettingsControllerImpl(
     private val trackedCaptureUiState: MutableStateFlow<TrackedCaptureUiState>,
-    private val viewModelScope: CoroutineScope,
     private val cameraSystem: CameraSystem,
-    private val externalCaptureMode: ExternalCaptureMode
+    private val externalCaptureMode: ExternalCaptureMode,
+    coroutineContext: CoroutineContext
 ) : QuickSettingsController {
+    private val job = Job(parent = coroutineContext[Job])
```
temcguir: Technically, we need `close()`, `cancel()`, and optionally, `join()` functions on the `QuickSettingsControllerImpl` (not on the `QuickSettingsController` interface), so that any launched work within this controller can be canceled. When using a ViewModelScope's context, it won't be a big deal, but if it's used outside of a ViewModelScope, such as in tests, it may be necessary. See go/android-api-guidelines-async#coroutines-close-cancel.

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/controller/CaptureScreenControllerImpl.kt
```diff
@@ -0,0 +1,49 @@
...
+
+    override fun setDisplayRotation(deviceRotation: DeviceRotation) {
+        scope.launch {
+            cameraSystem.setDeviceRotation(deviceRotation)
+        }
+    }
+
+    override fun updateLastCapturedMedia() {
+        updateLastCapturedMediaCallback()
+    }
```
temcguir: `setDisplayRotation` should be moved. to `CameraController`, and `updateLastCapturedMedia` should be moved to `ImageWellController`. Then you can delete `CaptureScreenController`.

temcguir (changes requested):

ui/components/capture/build.gradle.kts
```diff
@@ -100,6 +100,8 @@ dependencies {
     androidTestImplementation(libs.androidx.junit)
     androidTestImplementation(libs.androidx.espresso.core)
 
+    implementation(libs.kotlinx.atomicfu)
+
     implementation(project(":ui:uistate"))
     implementation(project(":ui:uistate:capture"))
     implementation(project(":core:camera"))
```
temcguir, line 106: Sorry I just realized this when looking at removing Hilt dependencies from the non-app libraries.

This dependency on `:core:camera` highlights an architectural issue that makes testing our UI components difficult.

Right now, `:ui:components` contains not just Composables, but also the Controller implementations, which pull in heavyweight dependencies like `:core:camera` and Hilt. This forces our UI tests for simple components to deal with a larger, more complex dependency graph just to provide fakes.

To improve testability, I propose we separate our UI into three distinct modules:

1. :ui:api (renamed from :ui:uistate): Contains the UI state data classes and Controller *interfaces*.
2. :ui:components (this module): Contains only the `@Composable` functions.
3. :ui:controller (new module): Contains the Controller *implementations*.

This decouples our Composables from their production logic, allowing us to write lean UI tests against simple, fake controller implementations without needing the entire dependency graph.

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/controller/CaptureControllerImpl.kt
```diff
@@ -0,0 +1,326 @@
...
+        )
+    }
+
+    suspend fun join() {
+        job.join()
+    }
+
+    fun cancel() {
+        job.cancel()
+    }
```
temcguir: The current API with separate `cancel()` and `join()` methods allows for a deadlock if `join()` is called first. We can make this API safer by combining them.

I recommend replacing `cancel()` and `join()` with a single `cancelScope()` function that returns the `Job`. This makes the incorrect usage a compile-time error and guides developers to the safe `controller.cancelScope().join()` pattern:

```kotlin
/**
 * Initiates the cancellation of this controller's scope and returns its Job.
 * To wait for cancellation to complete, call .join() on the returned Job.
 */
fun cancelScope(): Job {
    scope.cancel()
    return scope.coroutineContext.job
}
```

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/controller/CaptureControllerImpl.kt
```diff
@@ -0,0 +1,326 @@
...
+
+    fun close() {
+        job.cancel()
+    }
```
temcguir: `close()` is supposed to have a different behavior form `cancel()`: it's supposed to ensure that no new coroutines get launched in the scope, but it's supposed to ensure that all of the existing child jobs complete normally. It's actually a bit of a more complicated change.

Since we don't yet have a use for this close() method, let's just leave it out for now, or if you're interested, we can discuss how to implement it offline.

ui/components/capture/build.gradle.kts
davidjiagoogle: I would prefer keeping the existing ui:components & ui:uistate modules, and additionally create:
1. ui:controller (Controller interfaces, minimal deps)
2. ui:controller:impl (Controller impls, dep to camera:core, and ui:uistates for the time being)

temcguir: That works as well.

ui/controller/impl/build.gradle.kts
```diff
@@ -0,0 +1,85 @@
+/*
+ * Copyright (C) 2023 The Android Open Source Project
```
temcguir: nit: 2026

ui/controller/build.gradle.kts
```diff
@@ -0,0 +1,74 @@
+/*
+ * Copyright (C) 2023 The Android Open Source Project
```
temcguir: nit: 2026

ui/controller/build.gradle.kts
```diff
@@ -0,0 +1,74 @@
...
+    implementation(project(":data:media"))
+    implementation(project(":ui:uistate"))
+    implementation(project(":ui:uistate:capture"))
+    implementation(project(":core:model"))
+}
+
+// Allow references to generated code
+kapt {
+    correctErrorTypes = true
+}
```
temcguir: Is this needed?

temcguir (commented):
LGTM once all the nits and gemini comments are addressed.

temcguir (approved):

PR #486, opened 2026-03-12
getSupportedMimeTypes() API

davidjiagoogle:
Add a function to CameraSystem that return a list of supported MIME types, given a lens facing.

For the CameraXCameraSystem implementation, invoking this function requires initialize() to finish first because it relies on CameraConstraints to determine the MIME type capabilities.

### DIFF
--- a/core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraSystem.kt
+++ b/core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraSystem.kt
@@ -66,14 +66,17 @@ import com.google.jetpackcamera.model.LowLightBoostState
 import com.google.jetpackcamera.model.SaveLocation
 import com.google.jetpackcamera.model.StabilizationMode
 import com.google.jetpackcamera.model.StreamConfig
+import com.google.jetpackcamera.model.TARGET_FPS_15
+import com.google.jetpackcamera.model.TARGET_FPS_30
+import com.google.jetpackcamera.model.TARGET_FPS_60
+import com.google.jetpackcamera.model.TARGET_FPS_AUTO
 import com.google.jetpackcamera.model.TestPattern
+import com.google.jetpackcamera.model.UNLIMITED_VIDEO_DURATION
 import com.google.jetpackcamera.model.VideoQuality
 import com.google.jetpackcamera.model.ZoomStrategy
 import com.google.jetpackcamera.settings.SettableConstraintsRepository
 import com.google.jetpackcamera.settings.model.CameraAppSettings
 import com.google.jetpackcamera.settings.model.CameraConstraints
-import com.google.jetpackcamera.settings.model.CameraConstraints.Companion.FPS_15
-import com.google.jetpackcamera.settings.model.CameraConstraints.Companion.FPS_60
 import com.google.jetpackcamera.settings.model.CameraSystemConstraints
 import com.google.jetpackcamera.settings.model.forCurrentLens
 import dagger.hilt.android.scopes.ViewModelScoped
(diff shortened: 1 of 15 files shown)


core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraSystem.kt
```diff
@@ -1034,6 +1034,40 @@ constructor(
...
+     * Returns a set of MIME types supported by the underlying camera hardware capabilities and JCA.
+     *
+     * Empty set is returned if there are is available [CameraInfo].
+     * The [CameraConstraints] of the default lens is used to determine the supported MIME types.
+     * image/jpeg is supported if the device supports JPEG format for the default stream.
+     * video/mp4 is supported if any dynamic ranges are supported.
+     *
+     * @return A [Set] of strings representing the supported MIME types.
+     */
+    override suspend fun getSupportedMimeTypes(): Set<String> {
```
temcguir: Instead of this being part of `CameraSystem`, can it just be an extension function on `SystemConstraints`?

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraSystem.kt
```diff
@@ -1034,6 +1034,40 @@ constructor(
...
+        val result = mutableSetOf<String>()
+        currentSettings.value?.let { cameraAppSettings ->
+            val lensFacing = cameraAppSettings.cameraLensFacing
```
temcguir: If this functionis being called before opening the camera screen, how do we know this is the correct lens facing? I think this API needs to take into account both lens facing directions (and pass them as part of the API). For instance, maybe `getSupportedMimeTypes()` needs to return a `Map<LensFacing, Set<String>>`.

Alternatively, we could just return the entire `SystemConstraints` and then have an extension function on `CameraConstraints` called `getSupportedMimeTypes`.

Either of these would provide 2 benefits: you could detect whether the device only has a single lens facing (the keys in the map show the available cameras) and you could get the supported mime types.

data/settings/src/test/java/com/google/jetpackcamera/settings/ConstraintsTest.kt
```diff
@@ -0,0 +1,147 @@
...
+import com.google.jetpackcamera.model.DynamicRange
+import com.google.jetpackcamera.model.ImageOutputFormat
+import com.google.jetpackcamera.model.LensFacing
+import com.google.jetpackcamera.model.StreamConfig
+import com.google.jetpackcamera.settings.model.CameraConstraints
+import com.google.jetpackcamera.settings.model.CameraSystemConstraints
+import com.google.jetpackcamera.settings.model.getSupportedMimeTypes
+import org.junit.Test
+
+class ConstraintsTest {
```
temcguir, line 31: This is missing a test runner.

data/settings/src/main/java/com/google/jetpackcamera/settings/model/Constraints.kt
```diff
@@ -54,6 +54,40 @@ inline fun <reified T> CameraSystemConstraints.forDevice(
...
+        result.put(lensFacing, mimeTypes)
+        val constraints = perLensConstraints[lensFacing] ?: continue
+        constraints.supportedImageFormatsMap[StreamConfig.SINGLE_STREAM]?.let {
+            if (it.contains(
+                    ImageOutputFormat.JPEG
+                )
+            ) {
+                mimeTypes.add("image/jpeg")
+            }
+            if (constraints.supportedDynamicRanges.isNotEmpty()) {
```
temcguir: Maybe a better check here would be `supportedVideoQualitiesMap[DynamicRange.SDR]?.isNotEmpty() == true`?

temcguir (approved):

PR #488, opened 2026-03-24
Replace Material Icons with Google Fonts Material Symbols

davidjiagoogle:
This PR replaces all usage of the standard Compose Material Icons with the equivalent Material Symbols downloaded as XML
     vectors from Google Fonts. This migration reduces the dependency on the large material-icons-extended library and switches the app to the newer, variable-font-based symbol set.


feature/permissions/src/main/java/com/google/jetpackcamera/permissions/PermissionsEnums.kt
temcguir: Now that all icons are sourced from drawable resources instead of Compose `ImageVector`s, we can simplify this interface. You can remove the `getImageVector()` property and change `getDrawableResId()` to return a non-nullable `Int`. This allows you to vastly simplify the `getPainter()` function:

```kotlin
    @Composable
    fun getPainter(): Painter = painterResource(getDrawableResId())

    // ...

    @DrawableRes
    fun getDrawableResId(): Int
```
You'll also need to update the enum entries below to remove `getImageVector()` and return non-nullable `Int`s.

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/quicksettings/QuickSettingsEnums.kt
```diff
@@ -68,68 +41,68 @@ interface QuickSettingsEnum {
...
+
+        override fun getDrawableResId() = R.drawable.ic_nightlight
         override fun getTextResId() = R.string.quick_settings_flash_llb
         override fun getDescriptionResId() = R.string.quick_settings_flash_llb_description
     },
     LOW_LIGHT_BOOST_ACTIVE {
-        override fun getDrawableResId() = null
-        override fun getImageVector() = Icons.Filled.Nightlight
+
+        override fun getDrawableResId() = R.drawable.ic_nightlight
```
temcguir: Because only the outlined variant of the nightlight icon was imported, we lost the visual distinction between the active and inactive states for Low Light Boost. Both `LOW_LIGHT_BOOST_INACTIVE` and `LOW_LIGHT_BOOST_ACTIVE` are now returning the same drawable.

Please download the filled version of the nightlight icon (`ic_nightlight_fill1_24.xml`) using the `fill1` URL mentioned in the general feedback on `ic_mic.xml`, and restore it for the `LOW_LIGHT_BOOST_ACTIVE` state to match the original behavior. `LOW_LIGHT_BOOST_INACTIVE` should be updated to use the renamed `ic_nightlight_24.xml` drawable.

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/ToggleSwitch.kt
```diff
@@ -86,8 +81,8 @@ fun ToggleSwitch(
     trackColor: Color = Color.White.copy(alpha = TRACK_COLOR_ALPHA),
     thumbColor: Color = MaterialTheme.colorScheme.primaryFixedDim,
     disableColor: Color = MaterialTheme.colorScheme.onSurface,
-    leftIcon: ImageVector? = null,
-    rightIcon: ImageVector? = null,
+    leftIcon: Int? = null,
+    rightIcon: Int? = null,
```
temcguir: `nit:` Since `leftIcon` and `rightIcon` now expect drawable resource IDs, consider annotating them with `@DrawableRes` to ensure type safety and proper linting:

```kotlin
    @DrawableRes leftIcon: Int? = null,
    @DrawableRes rightIcon: Int? = null,
```

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/ToggleSwitch.kt
temcguir: `nit:` Please update the KDoc to reflect that these parameters now accept a drawable resource ID (`Int`) rather than an `ImageVector`.

```kotlin
 * @param leftIcon The drawable resource ID for the 'off' state.
 * @param rightIcon The drawable resource ID for the 'on' state.
```

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/CaptureScreenComponents.kt
```diff
@@ -323,16 +313,16 @@ fun CaptureModeToggleButton(
         leftIcon = if (uiState.selectedCaptureMode ==
             CaptureMode.IMAGE_ONLY
         ) {
-            Icons.Filled.CameraAlt
+            R.drawable.ic_camera_alt
         } else {
-            Icons.Outlined.CameraAlt
+            R.drawable.ic_camera_alt
```
temcguir: It appears the migration to local drawables accidentally removed the visual distinction for the active vs. inactive states of the capture mode toggle. Currently, the `if`/`else` branches return the exact same outlined drawable regardless of whether the mode is active.

To fix this, you'll need to download the filled variants of these icons (`ic_camera_alt_fill1_24.xml` and `ic_videocam_fill1_24.xml`) using the `fill1` URLs, add them to the `res/drawable` folder, and update the active branches to use them.

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/quicksettings/QuickSettingsEnums.kt
```diff
@@ -17,48 +17,21 @@ package com.google.jetpackcamera.ui.components.capture.quicksettings
...
-        require((iconResId == null).xor(iconVector == null)) {
-            "UI Item should have exactly one of iconResId or iconVector set."
-        }
-        return iconResId?.let { painterResource(it) }
-            ?: iconVector?.let {
-                rememberVectorPainter(
-                    it
-                )
-            }!! // !! allowed because we've checked null
+        return iconResId?.let { painterResource(it) }!!
```
temcguir: Similar to `PermissionsEnums`, we can simplify this interface now that we are exclusively using drawable resources. If you change `getDrawableResId()` to return `Int` instead of `Int?`, you can safely remove the `?.let` and `!!` operators:

```kotlin
    @Composable
    fun getPainter(): Painter = painterResource(getDrawableResId())

    @DrawableRes
    fun getDrawableResId(): Int
```

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/ToggleSwitch.kt
```diff
@@ -135,8 +130,8 @@ fun ToggleSwitch(
     )
 
     // --- 4. Icon Painters ---
-    val offIconPainter = leftIcon?.let { rememberVectorPainter(image = leftIcon) }
-    val onIconPainter = rightIcon?.let { rememberVectorPainter(image = rightIcon) }
+    val offIconPainter = leftIcon?.let { painterResource(id = leftIcon) }
+    val onIconPainter = rightIcon?.let { painterResource(id = rightIcon) }
```
temcguir: `nit:` While `leftIcon` and `rightIcon` are smartly cast to non-null inside the `let` block, it is more idiomatic in Kotlin to use `it` (the implicit name of the non-nullable parameter passed to the lambda) instead of referencing the outer nullable variable:

```kotlin
    val offIconPainter = leftIcon?.let { painterResource(id = it) }
    val onIconPainter = rightIcon?.let { painterResource(id = it) }
```

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/ToggleSwitch.kt
```diff
@@ -458,8 +453,8 @@ private fun Capture_Switch_Off_Enabled() {
         content =
         {
             ToggleSwitch(
-                leftIcon = Icons.Filled.CameraAlt,
-                rightIcon = Icons.Outlined.Videocam,
+                leftIcon = R.drawable.ic_camera_alt,
+                rightIcon = R.drawable.ic_videocam,
```
temcguir: Because the filled variants of the camera icons were missed in the migration, these previews currently pass the same outlined icons for both sides of the switch. 

Once you rename the existing files and import the filled variants (`ic_camera_alt_fill1_24` and `ic_videocam_fill1_24`), please update these previews to accurately reflect the toggled state by mixing the filled and outlined variants appropriately.

temcguir (changes requested):
As we caught a few unintended visual changes in this PR (like the loss of outlined vs. filled states), it highlights a great opportunity to add screenshot testing to the project.

If we implement a visual regression testing framework, like [Compose Preview Screenshot Testing](https://developer.android.com/studio/preview/compose-screenshot-testing), these kinds of regressions would be caught automatically by our CI. 

Since we already have `@Preview` setups for many of our components (like `ToggleSwitch` and `PermissionTemplate`), we could generate "golden" reference images. Any future refactors, like migrating icon libraries, would instantly fail the test if the pixel output changed. 

Setting that up is probably out of scope for this PR, but I highly recommend we consider adding it to our issue tracker.

ui/components/capture/src/main/res/drawable/ic_flip_camera_android.xml
Kimblebee, line 1: we might want to double check the correctness of the icon used in design

ui/components/capture/src/main/res/drawable/ic_flash_on_outline.xml
Kimblebee, line 1: this should be filled icon

ui/components/capture/src/main/res/drawable/ic_flash_auto_outline.xml
Kimblebee, line 1: this should be filled icon

Kimblebee (commented):
some icons were incorrect.

also i would suggest that you include "filled" or "outline" as part of the drawable naming conventions. Not all of them, but the ones that have the two vaiants. something like `ic_mic_filled.xml`

ui/components/capture/src/main/res/drawable/ic_flash_auto_outline.xml
davidjiagoogle: only flash_auto_filled is used. Deleted this file

ui/components/capture/src/main/res/drawable/ic_flash_on_outline.xml
davidjiagoogle: deleting this also

ui/components/capture/src/main/res/drawable/ic_flip_camera_android.xml
davidjiagoogle: Image

Seems correct

run-spotless.skill
temcguir, line 1: Did you mean to check this in?

.gemini/skills/run-spotless/scripts/run_spotless.sh
temcguir, line 1: Did you mean to check this in?

.gemini/skills/run-spotless/SKILL.md
temcguir, line 1: Did you mean to check this in?

temcguir (approved):

PR #491, opened 2026-04-02
multipleEventCutter

davidjiagoogle:
Re-implement MultipleEventsCutter that blocks capture button clicks during the capture process (either video or image) instead of using a hardcoded time duration. This prevents breaking tests while ensuring the button cannot be double-clicked before the underlying camera system call finishes or errors out.

### DIFF
--- a/ui/controller/impl/src/main/java/com/google/jetpackcamera/ui/controller/impl/CaptureControllerImpl.kt
+++ b/ui/controller/impl/src/main/java/com/google/jetpackcamera/ui/controller/impl/CaptureControllerImpl.kt
@@ -89,6 +89,7 @@ class CaptureControllerImpl(
     private var recordingJob: Job? = null
     private val job = Job(parent = coroutineContext[Job])
     private val scope = CoroutineScope(coroutineContext + job)
+    private val multipleEventsCutter = MultipleEventsCutter()
 
     override fun captureImage(contentResolver: ContentResolver) {
         if (externalCaptureMode == ExternalCaptureMode.VideoCapture) {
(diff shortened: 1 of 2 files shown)


ui/controller/impl/src/main/java/com/google/jetpackcamera/ui/controller/impl/MultipleEventsCutter.kt
```diff
@@ -0,0 +1,41 @@
...
+internal class MultipleEventsCutter {
+    private val isProcessing = atomic(false)
+
+    /**
+     * Executes the [event] if no other event is currently being processed.
+     */
+    fun processEvent(event: () -> Unit) {
+        if (isProcessing.compareAndSet(expect = false, update = true)) {
+            event()
+        }
```
temcguir, line 32: Silently dropping events here is the root cause of the test flakiness. The testing framework assumes the click was successful, but the underlying system ignores it. We should remove this class entirely and rely on the UI being disabled while a capture is in progress.

ui/controller/impl/src/main/java/com/google/jetpackcamera/ui/controller/impl/CaptureControllerImpl.kt
```diff
@@ -102,56 +103,66 @@ class CaptureControllerImpl(
...
-                    if (saveLocation !is SaveLocation.Cache) {
-                        imageWellController.updateLastCapturedMedia()
-                    } else {
-                        savedUri?.let {
-                            scope.launch {
-                                postCurrentMediaToMediaRepository(
-                                    mediaRepository,
-                                    MediaDescriptor.Content.Image(it, null, true)
-                                )
+        multipleEventsCutter.processEvent {
```
temcguir, line 106: Instead of wrapping this in `processEvent`, `cameraSystem.takePicture()` and `cameraSystem.startVideoRecording()` should internally update an 'in-progress' state on `CameraState`. This will flow down to the UI state, disable the capture button, and naturally prevent a second click.

temcguir (changes requested):
Moving `MultipleEventsCutter` into the Controller layer is going to reintroduce the same test flakiness that caused us to remove the `MultipleEventsCutter` from the UI layer in the first place.

Because `MultipleEventsCutter` silently swallows events, our automated tests will occasionally fire a `performClick()` that gets dropped. The test will then time out or fail (sometimes) because it expects an image/video to be captured but nothing happened.

Instead of an imperative debounce workaround, we should fix this reactively. We should tie the capture button's enabled state directly to the camera's processing pipeline:
1. Emit an 'in-progress' state from `CameraSystem` / `CameraState` the moment `takePicture` or `startVideoRecording` is invoked.
2. Have the UI state adapter (`CaptureButtonUiStateAdapter`) map this 'in-progress' state to `CaptureButtonUiState.Unavailable` (or the Disabled state introduced in PR #425).
3. PR #425 already updates the `CaptureButton` composable to be semantically disabled (using `Modifier.semantics { disabled() }`) when appropriate, and updates our tests to use `waitForCaptureButton()` which explicitly checks `isEnabled()`.

If we implement the reactive state in `CameraState`, it will integrate with the groundwork laid out in PR #425, preventing double-clicks naturally, giving the user proper visual feedback that a capture is occurring, and tying the button's interactability to the actual reality of the camera hardware without breaking our tests.

PR #492, opened 2026-04-07
scrollable setting dialog

Kimblebee:
Adds a scrollable  property to settings screen alert dialogs, such that overflowing text can still be accessible.

The id of the scrollable container is: `dialog_contents_container`

### DIFF
--- a/feature/settings/src/main/java/com/google/jetpackcamera/settings/ui/TestTags.kt
+++ b/feature/settings/src/main/java/com/google/jetpackcamera/settings/ui/TestTags.kt
@@ -27,7 +27,8 @@ package com.google.jetpackcamera.settings.ui
 
 const val SETTINGS_TITLE = "SettingsTitle"
 const val BACK_BUTTON = "BackButton"
-const val CLOSE_BUTTON = "CloseButton"
+const val BTN_CLOSE_POPUP = "btn_close_popup"
+const val CONTAINER_DIALOG_CONTENTS = "container_dialog_contents"
 
 // unsupported rationale tags
 const val DEVICE_UNSUPPORTED_TAG = "DeviceUnsupportedTag"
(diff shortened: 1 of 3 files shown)


feature/settings/src/main/java/com/google/jetpackcamera/settings/ui/TestTags.kt
```diff
@@ -27,7 +27,8 @@ package com.google.jetpackcamera.settings.ui
 
 const val SETTINGS_TITLE = "SettingsTitle"
 const val BACK_BUTTON = "BackButton"
-const val CLOSE_BUTTON = "CloseButton"
+const val BTN_CLOSE_POPUP = "btn_close_popup"
+const val CONTAINER_DIALOG_CONTENTS = "container_dialog_contents"
```
temcguir: Do we need to rename the close button in this PR? This will require that ITS changes their scripts to adapt to the new close button tag and it looks to be unrelated to this change. If we want to change the test tag style, we should do it all at once so that ITS can be updated once as well.

temcguir: It looks like the CONTAINER_DIALOG_CONTENTS doesn't follow either of the styles for test tag naming that we use.

temcguir (changes requested):

feature/settings/src/main/java/com/google/jetpackcamera/settings/ui/TestTags.kt
Kimblebee: sounds good, reverted

Kimblebee: updated, trying to follow the styleguide since it is a new string

temcguir (approved):

PR #493, opened 2026-04-07
Camera Feature Policy API

Kimblebee:
### Summary
Introduces `CameraAppConfig` in `:core:settings` to allow developer options, embedding applications, and test harnesses to configure default camera settings and control UI option visibility.

This decouples the baseline camera session state (`defaultValue`) from Compose UI presentation (`OptionVisibility`), enabling controls to be shown normally, hidden entirely, or restricted to a specific whitelist of options with guaranteed domain safety and graceful fallbacks.

---

### Key Changes

* **Configuration & Models (`core:settings`):**
  * Added `CameraAppConfig` with nullable properties: `captureMode`, `aspectRatio`, `flashMode`, `imageFormat`, and `dynamicRange`.
  * Added generic, non-null bounded `SettingConfig<T : Any>(defaultValue, visibility)`.
  * Defined the `OptionVisibility<T : Any>` sealed hierarchy:
    * `OptionVisibility.Visible`: All device-supported options are selectable (default).
    * `OptionVisibility.Hidden`: Setting is hidden/unavailable in the UI; camera runs with `defaultValue`.
    * `OptionVisibility.Only(enabledOptions)`: Whitelists specific options in the UI (minimum 2 options; supports `vararg` construction).
  * Enforced domain safety constraints:
    * `defaultValue` must be included in `OptionVisibility.Only`.
    * `flashMode` must default to `OFF` when `Hidden` and must include `OFF` when restricted.
    * `imageFormat` must default to `JPEG` when `Hidden`.
    * `dynamicRange` must default to `SDR` when `Hidden`.
  * Added `CameraAppConfig.toCameraAppSettings(defaultSettings)` to map overrides cleanly over baseline settings.

* **Camera System & Repository Wiring (`core:camera`, `data:camera`):**
  * Injected `CameraAppConfig` into `CameraXCameraSystemRepository`.
  * Applied `CameraAppConfig` defaults during lazy camera initialization before use cases bind.
  * Preserved un-overridden settings from user preferences rather than hardcoded fallbacks.

* **UI State & Adapters (`ui:uistateadapter:capture`):**
  * Integrated `OptionVisibility` policies into:
    * `CaptureModeUiStateAdapter`: Respects hidden capture mode, hides/disables toggle buttons or quick-settings rows appropriately, and provides clear disabled rationales.
    * `FlashModeUiStateAdapter`: Filters selectable flash modes or marks as `Unavailable` when hidden or unsupported.
    * `HdrUiStateAdapter`: Disables HDR options when dynamic range or image format are restricted. Fixed an issue where Low Light Boost disabled all HDR options instead of keeping "Off" selectable.
    * `CaptureUiStateAdapter`: Wires `CameraAppConfig` into compound UI state generation with `appConfig: CameraAppConfig? = null` default parameter.

* **Dependency Injection & App Integration (`app`):**
  * Provided `CameraAppConfig` via `AppModule.providesCameraAppConfig()`.
  * Added `@Volatile @VisibleForTesting var testCameraAppConfig: CameraAppConfig?` to enable isolated, deterministic device test overrides.

* **Testing:**
  * **Unit Tests:**
    * `:core:settings`: Comprehensive validation of `CameraAppConfig`, `SettingConfig`, and `OptionVisibility` edge cases in `CameraAppConfigTest`.
    * `:data:camera`: Verified lazy initialization and default overrides in `CameraXCameraSystemRepositoryTest`.
    * `:ui:uistateadapter:capture`: Tested UI state adaptation, visibility restrictions, and fallback behavior across `CaptureModeUiStateAdapterTest`, `FlashModeUiStateAdapterTest`, `HdrUiStateAdapterTest`, and `CaptureUiStateAdapterTest`.
    * `:feature:preview`: Verified ViewModel integration with `CameraAppConfig` in `PreviewViewModelTest`.
  * **Device Instrumentation Tests (`app`):**
    * Added end-to-end device tests validating Compose UI behavior under various `CameraAppConfig` configurations:
      * `CaptureModeAppConfigDeviceTest` (including photo and video capture execution)
      * `FlashModeAppConfigDeviceTest` (including graceful fallback on devices without flash units)
      * `AspectRatioAppConfigDeviceTest`
      * `HdrAppConfigDeviceTest`
    * Isolated `CaptureModeSettingsTest.hdr_supports_video_only()` using `testCameraAppConfig` for deterministic behavior across dual-HDR and single-HDR devices.

---

### Usage Example

```kotlin
val config = CameraAppConfig(
    captureMode = SettingConfig(
        defaultValue = CaptureMode.IMAGE_ONLY,
        visibility = OptionVisibility.Hidden
    ),
    flashMode = SettingConfig(
        defaultValue = FlashMode.OFF,
        visibility = OptionVisibility.Only(FlashMode.OFF, FlashMode.AUTO)
    ),
    dynamicRange = SettingConfig(
        defaultValue = DynamicRange.SDR,
        visibility = OptionVisibility.Hidden
    )
)
```

---

### Limitations & Future Work

1. **Audio Configuration:** `audioEnabled` configuration is deferred to a follow-up PR.
2. **Aspect Ratio Preservation Across Mode Switches:** Photo mode currently enforces 3:4 for external capture compatibility. We should consider if we want to preserve developer-configured aspect ratios (e.g., 1:1) for thsee scenarios


data/settings/api/src/main/java/com/google/jetpackcamera/settings/api/RestrictionConfig.kt
```diff
@@ -0,0 +1,136 @@
...
+data class SettingConfig<T>(
+    val defaultValue: T,
+    val uiRestriction: OptionRestrictionConfig<T> = OptionRestrictionConfig.NotRestricted()
+) {
+    init {
+        // Validate that if options are enabled for this setting, the default value
+        // is always included in the set of enabled options.
+        (uiRestriction as? OptionRestrictionConfig.OptionsEnabled)?.let {
+            require(
+                uiRestriction.enabledOptions.size >= 2 &&
```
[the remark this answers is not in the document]
Kimblebee: enabled option of one means the setting is effectively unusable to a user, so we can consider it bloat for now and just remove it from visibility.

app/src/main/java/com/google/jetpackcamera/MainActivity.kt
```diff
@@ -217,6 +218,9 @@ class MainActivity : ComponentActivity() {
     private val shouldReviewAfterCapture: Boolean
         get() = intent?.shouldReviewAfterCapture == true
 
+    private val useDeveloperConfig: Boolean
```
temcguir, line 196: What is the rationale for gatekeeping the developer configuration behind an Intent extra (`KEY_USE_DEVELOPER_CONFIG`)?

Since `DeveloperAppConfig` is provided via Hilt (`@Provides DeveloperAppConfig`), the dependency injection binding itself should define the configuration. If an integrating app (or a test) does not want any restrictions, it can simply bind the default unrestricted configuration (`DeveloperAppConfig()`).

Requiring an intent boolean introduces unnecessary indirection and forces host apps embedding JCA Composables (`JcaApp`) to plumb an arbitrary flag through navigation arguments and `SavedStateHandle` just to make their injected configuration take effect. Could we remove this gate and have the injected configuration be respected directly?

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewViewModel.kt
```diff
@@ -118,6 +121,8 @@ class PreviewViewModel @Inject constructor(
     private val externalUris: List<Uri> = savedStateHandle.getCaptureUris()
     private lateinit var externalUriProgress: IntProgress
 
+    private val useDeveloperConfig: Boolean = savedStateHandle.getUseDeveloperConfig()
```
temcguir, line 120: Related to my comment on `MainActivity.kt`: since `appConfig: DeveloperAppConfig` is already injected directly into `PreviewViewModel` via Hilt, we shouldn't need to gate its usage behind `savedStateHandle.getUseDeveloperConfig()`.

Removing this check allows `appConfig` to be consumed directly and eliminates the need to plumb the boolean flag through navigation arguments and `SavedStateHandle`.

data/settings/api/src/main/java/com/google/jetpackcamera/settings/api/RestrictionConfig.kt
```diff
@@ -0,0 +1,114 @@
...
+        defaultSettings: CameraAppSettings = DEFAULT_CAMERA_APP_SETTINGS
+    ): CameraAppSettings {
+        return defaultSettings.copy(
+            aspectRatio = this.aspectRatio.defaultValue,
+            flashMode = this.flashMode.defaultValue,
+            captureMode = this.captureMode.defaultValue,
+            imageFormat = this.imageOutputFormat.defaultValue,
+            dynamicRange = this.videoDynamicRange.defaultValue
+        )
+    }
```
temcguir, line 71: `toCameraAppSettings()` unconditionally copies all 5 settings onto `defaultSettings`.

Because `DeveloperAppConfig` assigns default constructor values to every property (such as `aspectRatio = SettingConfig(DEFAULT_CAMERA_APP_SETTINGS.aspectRatio)`), calling `toCameraAppSettings(defaultSettings)` will forcibly overwrite all settings with hardcoded defaults. For example, if a developer only configures `imageOutputFormat = JPEG_ULTRA_HDR`, passing a custom `defaultSettings` will unintentionally reset `aspectRatio`, `flashMode`, `captureMode`, and `dynamicRange`.

If this conversion function is kept, it should act as a sparse overlay (only overriding properties the developer explicitly configured) rather than unconditionally clobbering the baseline `defaultSettings`.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewViewModel.kt
```diff
@@ -131,15 +136,22 @@ class PreviewViewModel @Inject constructor(
...
     private var initializationDeferred: Deferred<Unit> = viewModelScope.async {
+        val defaultSettings = settingsRepository.defaultCameraAppSettings.first()
         cameraSystemRepository.cameraSystem.initialize(
-            cameraAppSettings = settingsRepository.defaultCameraAppSettings.first()
+            cameraAppSettings =
+            if (useDeveloperConfig) {
+                appConfig.toCameraAppSettings(defaultSettings)
+            } else {
+                defaultSettings
+            }
```
temcguir: Related to `toCameraAppSettings()` in `RestrictionConfig.kt`: calling `appConfig.toCameraAppSettings(defaultSettings)` here wipes out the user's persistent preferences loaded from `settingsRepository.defaultCameraAppSettings.first()`.

For example, if a user has previously saved a preference for 3:4 aspect ratio in DataStore, launching with an `appConfig` that only configures Ultra HDR will reset the user's aspect ratio back to 16:9.

Instead of having the ViewModel perform ad-hoc settings reconciliation here, developer baseline defaults should be integrated into the settings layer (e.g. `SettingsRepository` / `SettingsDataSource`) so unconfigured user preferences from DataStore are preserved.

ui/uistateadapter/capture/src/main/java/com/google/jetpackcamera/ui/uistateadapter/capture/compound/CaptureUiStateAdapter.kt
```diff
@@ -84,8 +88,11 @@ fun captureUiState(
...
         val roundedCameraState = cameraState.copy(videoRecordingState = roundedVideoRecordingState)
 
+        val captureModeRestriction =
+            appConfig?.captureMode?.uiVisibility ?: OptionAvailabilityConfig.NotRestricted
         val captureModeUiState = CaptureModeUiState.from(
             systemConstraints,
+            captureModeRestriction,
             cameraAppSettings,
             externalCaptureMode
         )
```
temcguir, line 101: `DeveloperAppConfig` defines restrictions for `aspectRatio`, `flashMode`, `imageOutputFormat`, and `videoDynamicRange`, but here in `CaptureUiStateAdapter.kt` only `captureModeRestriction` is currently forwarded to UI state adapters.

Could you add a brief note or TODO confirming that connecting `flashMode`, `aspectRatio`, and `imageOutputFormat` restrictions to `FlashModeUiStateAdapter`, `AspectRatioUiStateAdapter`, and `HdrUiStateAdapter` will happen in follow-up PRs?

data/settings/api/src/main/java/com/google/jetpackcamera/settings/api/RestrictionConfig.kt
```diff
@@ -0,0 +1,114 @@
...
+data class DeveloperAppConfig(
+    val captureMode: SettingConfig<CaptureMode> =
+        SettingConfig(DEFAULT_CAMERA_APP_SETTINGS.captureMode),
+    val aspectRatio: SettingConfig<AspectRatio> =
+        SettingConfig(DEFAULT_CAMERA_APP_SETTINGS.aspectRatio),
+    val flashMode: SettingConfig<FlashMode> = SettingConfig(DEFAULT_CAMERA_APP_SETTINGS.flashMode),
+    val imageOutputFormat: SettingConfig<ImageOutputFormat> =
+        SettingConfig(DEFAULT_CAMERA_APP_SETTINGS.imageFormat),
+    val videoDynamicRange: SettingConfig<DynamicRange> =
+        SettingConfig(DEFAULT_CAMERA_APP_SETTINGS.dynamicRange)
```
temcguir, line 39: In your design doc, `audio: SettingConfig<Boolean>` (or `audioEnabled`) was listed as one of the configurable settings.

Was audio intentionally omitted from this initial PR, or should we include `audio: SettingConfig<Boolean>` alongside `captureMode`, `aspectRatio`, and `flashMode`?

data/settings/api/src/main/java/com/google/jetpackcamera/settings/api/RestrictionConfig.kt
```diff
@@ -0,0 +1,114 @@
...
+import com.google.jetpackcamera.model.FlashMode
+import com.google.jetpackcamera.model.ImageOutputFormat
+import com.google.jetpackcamera.settings.model.CameraAppSettings
+import com.google.jetpackcamera.settings.model.DEFAULT_CAMERA_APP_SETTINGS
+
+/**
+ * Defines a configuration for the Jetpack Camera App that can be used by developers
+ * to override the default app settings.
+ */
+data class DeveloperAppConfig(
```
temcguir, line 30: `DeveloperAppConfig` feels a bit generic and doesn't clearly communicate what the API is doing. In standard Jetpack libraries, "Developer" is typically redundant (since the entire SDK is for developers), and "App" doesn't reflect that this configures camera features and presentation rules.

What do you think about naming this **`CameraFeaturePolicy`** (paired with `FeatureConfig<T>` or `FeaturePolicy<T>`)?

This feels much more aligned with its actual purpose:
1. **"Camera"**: Anchors it to the domain rather than generic app configs.
2. **"Feature"**: It governs specific camera features (Capture Mode, Flash, Ultra HDR, Dynamic Range, Aspect Ratio).
3. **"Policy"**: Accurately describes what it is: an application-level policy dictating session defaults and UI availability rules.

Also, could we add `@param` tags to the KDoc documenting constructor properties and the flash mode invariants enforced in `init` (`FlashMode.OFF` requirement)?

ui/uistateadapter/capture/src/main/java/com/google/jetpackcamera/ui/uistateadapter/capture/CaptureModeUiStateAdapter.kt
```diff
@@ -56,36 +57,35 @@ fun CaptureModeToggleUiState.Companion.from(
...
-                is SingleSelectableUiState.SelectableUi -> item.value == CaptureMode.VIDEO_ONLY
-                is SingleSelectableUiState.Disabled -> item.value == CaptureMode.VIDEO_ONLY
-            }
+        val imageOnlyState = availableCaptureModes.first { it.value == CaptureMode.IMAGE_ONLY }
+        val videoOnlyState = availableCaptureModes.first { it.value == CaptureMode.VIDEO_ONLY }
+        if (imageOnlyState is SingleSelectableUiState.Disabled ||
+            videoOnlyState is SingleSelectableUiState.Disabled
+        ) {
+            CaptureModeToggleUiState.Unavailable
+        } else {
```
temcguir, line 82: When either `IMAGE_ONLY` or `VIDEO_ONLY` is disabled (e.g. by hardware constraints, concurrent camera, or an external capture mode), `CaptureModeToggleUiState` resolves to `Unavailable`.

This makes complete sense because a binary toggle button requires at least two enabled options to switch between. 

Could we add a brief comment or KDoc here clarifying this contract for library consumers? Specifically:
1. Document that `CaptureModeToggleUiState` is intentionally scoped only to binary quick-toggle widgets (which require 2 valid selectable states).
2. Clarify that integrating apps or custom UI layouts wanting to inspect or display disabled states and their `disabledReason` should observe `CaptureModeUiState` directly (via `CaptureModeUiState.Available(options)` and `SingleSelectableUiState.Disabled.disabledReason`), rather than relying on `CaptureModeToggleUiState`.

settings.gradle.kts
```diff
@@ -72,3 +72,4 @@ include(":core:settings:datastore-prefs")
 include(":core:settings:datastore-prefs:testing")
 include(":ui:debug")
 include(":ui:debug:testing")
+include(":data:settings:api")
```
temcguir, line 75: Creating `:data:settings:api` to host the developer options API introduces an architectural layering violation:

1. **`:ui:uistateadapter:capture` is a Public Library API, Not a Feature**:
   `:ui:uistateadapter:capture` is part of JCA's public library surface, providing reusable adapters for client applications to bind camera state into UI components. In this PR, its public functions (`CaptureModeUiState.Companion.from` and `CaptureModeToggleUiState.Companion.from`) accept `OptionAvailabilityConfig<CaptureMode>` directly in their public method signatures.
2. **The Data Layer Must Not Leak into Public APIs**:
   In clean architecture, `:data:*` modules represent internal implementation details (DataStore persistence, network/device sources, repository implementations). By placing `OptionAvailabilityConfig` and `DeveloperAppConfig` under `:data:settings:api`, our public UI library is forced to expose a `:data:*` package to external client applications, compelling every integrator to depend on the data layer.
3. **Natural Alignment with `:core:settings`**:
   Notice that all other domain models used in `CaptureModeUiState.Companion.from` (specifically `CameraAppSettings`, `CameraConstraints`, and `CameraSystemConstraints`) already live in `:core:settings` (`com.google.jetpackcamera.settings.model.*`).
4. **Recommendation**:
   Move `CameraFeaturePolicy` (currently `DeveloperAppConfig`), `SettingConfig`, and `OptionAvailabilityConfig` into **`:core:settings`** alongside `CameraAppSettings` and `Constraints`. This keeps our public contracts clean, eliminates the extraneous `:data:settings:api` subproject, and ensures external apps never need to depend on our internal data persistence modules.

data/settings/api/src/main/java/com/google/jetpackcamera/settings/api/RestrictionConfig.kt
temcguir, line 1: Per Kotlin style guidelines and Android API conventions, when a file contains a primary top-level class, the file name should match the class name.

Currently, the file is named `RestrictionConfig.kt`, but the top-level declaration is `DeveloperAppConfig` (or `CameraFeaturePolicy`). Could we rename the file to match the root class (e.g. `CameraFeaturePolicy.kt` or `DeveloperAppConfig.kt`)?

data/settings/api/src/main/java/com/google/jetpackcamera/settings/api/RestrictionConfig.kt
temcguir, line 39: There is a naming mismatch between the properties in `DeveloperAppConfig` and the corresponding properties in `CameraAppSettings`:
* `DeveloperAppConfig.imageOutputFormat` vs `CameraAppSettings.imageFormat`
* `DeveloperAppConfig.videoDynamicRange` vs `CameraAppSettings.dynamicRange`

Notice that in `toCameraAppSettings()`, this forces an arbitrary cross-mapping:
```kotlin
imageFormat = this.imageOutputFormat.defaultValue,
dynamicRange = this.videoDynamicRange.defaultValue
```

To maintain consistency across our public API surface and internal models, could we align these names to match `CameraAppSettings` (i.e. `imageFormat` and `dynamicRange`)?

data/settings/api/src/main/java/com/google/jetpackcamera/settings/api/RestrictionConfig.kt
```diff
@@ -0,0 +1,114 @@
...
+
+    /** The entire setting is unavailable and hidden from the UI. */
+    data object Hidden : OptionAvailabilityConfig<Nothing>
+
+    /** ONLY the options in this set are allowed, if supported by the device. */
+    data class OptionsEnabled<T>(val enabledOptions: Set<T>) : OptionAvailabilityConfig<T> {
+        init {
+            require(enabledOptions.size >= 2) {
+                "enabledOptions must contain at least 2 options. Use Hidden to lock a single option and hide the control."
+            }
```
temcguir, line 111: Throwing `IllegalArgumentException` at runtime when `enabledOptions.size < 2` creates an easy crash trap for integrating developers.

For example: if a client application computes allowed options dynamically (e.g. filtering a set of modes against remote feature flags or device capabilities):
```kotlin
val allowedModes = remoteConfig.supportedModes.filter { deviceSupports(it) }.toSet()
// If allowedModes happens to resolve to 1 item on a particular device:
OptionAvailabilityConfig.OptionsEnabled(allowedModes) // CRASH: IllegalArgumentException!
```

Forcing developers to write manual boilerplate (`if (options.size >= 2) OptionsEnabled(options) else Hidden`) is error-prone. 

Could we either:
1. Provide a factory function (e.g. `OptionAvailabilityConfig.from(setOf(...))`) that automatically falls back to `Hidden` (or a single locked option) if `size < 2`?
2. Or allow a single option in `OptionsEnabled`, rendering the UI control in an unclickable/locked state rather than throwing an unhandled runtime exception?

ui/uistateadapter/capture/src/main/java/com/google/jetpackcamera/ui/uistateadapter/capture/CaptureModeUiStateAdapter.kt
```diff
@@ -260,17 +233,30 @@ private fun getCaptureModeDisabledReason(
...
+    externalCaptureMode: ExternalCaptureMode,
+    isHdrOn: Boolean
 ): DisabledReason {
     when (disabledCaptureMode) {
         CaptureMode.IMAGE_ONLY -> {
             if (externalCaptureMode == ExternalCaptureMode.VideoCapture) {
                 return DisabledReason
                     .IMAGE_CAPTURE_EXTERNAL_UNSUPPORTED
             }
+            when (restrictionConfig) {
```
temcguir, line 249: In `getCaptureModeDisabledReason`, there are checks for `OptionAvailabilityConfig.Hidden`:
```kotlin
when (restrictionConfig) {
    is OptionAvailabilityConfig.Hidden ->
        return DisabledReason.IMAGE_CAPTURE_RESTRICTED
```

However, both entry points that call this function already short-circuit when `Hidden`:
1. In `CaptureModeUiState.Companion.from` (lines 110-112):
   ```kotlin
   if (restrictionConfig is OptionAvailabilityConfig.Hidden) {
       return CaptureModeUiState.Unavailable
   }
   ```
2. In `CaptureModeToggleUiState.Companion.from` (lines 64-67):
   ```kotlin
   if (... || restrictionConfig is OptionAvailabilityConfig.Hidden) {
       CaptureModeToggleUiState.Unavailable
   }
   ```

Because both entry points return `Unavailable` immediately when `Hidden`, `getAvailableCaptureModes` and `getCaptureModeDisabledReason` are never called with `OptionAvailabilityConfig.Hidden`. 

These `is OptionAvailabilityConfig.Hidden` branches in `getCaptureModeDisabledReason` are unreachable dead code and can be safely removed.

data/settings/api/src/test/java/com/google/jetpackcamera/settings/api/RestrictionConfigTest.kt
```diff
@@ -0,0 +1,170 @@
...
+
+        assertThat(appSettings.aspectRatio).isEqualTo(AspectRatio.NINE_SIXTEEN)
+        assertThat(appSettings.flashMode).isEqualTo(FlashMode.ON)
+        assertThat(appSettings.captureMode).isEqualTo(CaptureMode.VIDEO_ONLY)
+        assertThat(appSettings.imageFormat).isEqualTo(ImageOutputFormat.JPEG)
+        assertThat(appSettings.dynamicRange).isEqualTo(DynamicRange.SDR)
+    }
+
+    @Test
+    fun toCameraAppSettings_withCustomDefaults_preservesUnoverriddenSettings() {
```
temcguir, line 157: The test `toCameraAppSettings_withCustomDefaults_preservesUnoverriddenSettings` passes, but it masks a significant bug because it only checks `maxVideoDurationMillis`, a property that does not exist in `DeveloperAppConfig`.

If we test an un-overridden property that *is* declared in `DeveloperAppConfig` (such as `aspectRatio` or `flashMode`):
```kotlin
val customDefaults = DEFAULT_CAMERA_APP_SETTINGS.copy(
    aspectRatio = AspectRatio.THREE_FOUR
)
val developerConfig = DeveloperAppConfig(
    imageOutputFormat = SettingConfig(ImageOutputFormat.JPEG_ULTRA_HDR)
)

val appSettings = developerConfig.toCameraAppSettings(customDefaults)
assertThat(appSettings.aspectRatio).isEqualTo(AspectRatio.THREE_FOUR) // FAILS! Reset to NINE_SIXTEEN
```

This test will fail because `DeveloperAppConfig`'s constructor assigns `aspectRatio = SettingConfig(DEFAULT_CAMERA_APP_SETTINGS.aspectRatio)`, which unconditionally overwrites the user's custom default `AspectRatio.THREE_FOUR` back to `NINE_SIXTEEN`.

Could we add a test asserting that custom defaults for `aspectRatio` and `flashMode` are preserved when the developer config only touches `imageOutputFormat`?

ui/uistateadapter/capture/src/main/java/com/google/jetpackcamera/ui/uistateadapter/capture/CaptureModeUiStateAdapter.kt
temcguir, line 249: Because both `from()` entry points (lines 65 and 110) already return `Unavailable` when `Hidden`, the `Hidden` branches here and on lines 301 and 332 are unreachable.

Since `NotRestricted` is a no-op, we can simplify this check across all three branches (`IMAGE_ONLY`, `VIDEO_ONLY`, and `STANDARD`) to:
```kotlin
if (restrictionConfig is OptionAvailabilityConfig.OptionsEnabled &&
    disabledCaptureMode !in restrictionConfig.enabledOptions
) {
    return DisabledReason.IMAGE_CAPTURE_RESTRICTED // (or VIDEO / HYBRID)
}
```

data/settings/api/src/main/java/com/google/jetpackcamera/settings/api/RestrictionConfig.kt
```diff
@@ -0,0 +1,114 @@
...
+
+/**
+ * Represents UI option availability applied to a setting.
+ */
+sealed interface OptionAvailabilityConfig<out T> {
+    /** All device-supported options are available. */
+    data object NotRestricted : OptionAvailabilityConfig<Nothing>
+
+    /** The entire setting is unavailable and hidden from the UI. */
+    data object Hidden : OptionAvailabilityConfig<Nothing>
```
temcguir, line 104: Regarding the lens-flip demotion trap under Limitations & Future Work in the PR description:

As you noted, when a setting like Ultra HDR has `uiVisibility = Hidden`, flipping to an unsupported front lens demotes it to JPEG, and the user has no UI affordance to re-enable it when switching back to the rear lens.

Separating `desiredSettings` (the developer's baseline policy) from `effectiveSettings` (the active sensor's constrained state) will be critical for making `Hidden` policies robust in production. Do we already have an issue or follow-up PR tracking this camera session lifecycle redesign?

temcguir (changes requested):

ui/uistateadapter/capture/src/main/java/com/google/jetpackcamera/ui/uistateadapter/capture/compound/CaptureUiStateAdapter.kt
Kimblebee: theyre all implemented now, the original commit for that mustve been lost in a stash

core/settings/src/main/java/com/google/jetpackcamera/settings/model/CameraFeaturePolicy.kt
```diff
@@ -0,0 +1,234 @@
...
+                    DynamicRange.SDR in visibility.enabledOptions
+                ) {
+                    "DynamicRange.SDR must always be included in enabledOptions for dynamicRange."
+                }
+
+                is OptionVisibility.Hidden -> require(
+                    config.defaultValue == DynamicRange.SDR
+                ) {
+                    "When dynamicRange is Hidden, defaultValue must be DynamicRange.SDR."
+                }
```
temcguir: Requiring `defaultValue == ImageOutputFormat.JPEG` (and `DynamicRange.SDR`) when `Hidden` blocks a primary developer options use case: client apps that want to default or lock capture to Ultra HDR (or HDR video) without exposing a user toggle.

While requiring `JPEG` in `enabledOptions` for `OptionVisibility.Only` makes sense as a device fallback, when an option is `Hidden`, developers should be allowed to set `defaultValue = ImageOutputFormat.JPEG_ULTRA_HDR` or `DynamicRange.HLG10`. If the active camera hardware does not support Ultra HDR or HLG10, `CameraXCameraSystem.tryApplyImageFormatConstraints()` already gracefully demotes to JPEG/SDR at runtime.

Can we drop the `require(config.defaultValue == JPEG/SDR)` check when `Hidden`? (We should also remove/update the corresponding throwing tests in `CameraFeaturePolicyTest`).

data/settings/src/main/java/com/google/jetpackcamera/settings/LocalSettingsRepository.kt
```diff
@@ -98,4 +106,28 @@ class LocalSettingsRepository @Inject constructor(
...
+    private fun applyFeaturePolicy(
+        storedSettings: CameraAppSettings,
+        policy: CameraFeaturePolicy
+    ): CameraAppSettings {
+        return storedSettings.copy(
+            aspectRatio = enforceRestrictions(storedSettings.aspectRatio, policy.aspectRatio),
+            flashMode = enforceRestrictions(storedSettings.flashMode, policy.flashMode),
+            imageFormat = enforceRestrictions(storedSettings.imageFormat, policy.imageFormat),
+            dynamicRange = enforceRestrictions(storedSettings.dynamicRange, policy.dynamicRange),
+            captureMode = policy.captureMode?.defaultValue ?: storedSettings.captureMode
```
temcguir, line 119: Should this use `enforceRestrictions(storedSettings.captureMode, policy.captureMode)` like the other settings?

```kotlin
captureMode = enforceRestrictions(storedSettings.captureMode, policy.captureMode)
```

Using `policy.captureMode?.defaultValue ?: storedSettings.captureMode` unconditionally resets `captureMode` to `defaultValue` on every emission, even when `storedSettings.captureMode` is an allowed option under `OptionVisibility.Only`.

ui/uistateadapter/capture/src/main/java/com/google/jetpackcamera/ui/uistateadapter/capture/compound/QuickSettingsUiStateAdapter.kt
```diff
@@ -35,6 +35,9 @@ import com.google.jetpackcamera.ui.uistate.capture.compound.QuickSettingsUiState
  * @param hdrUiState The UI state for the HDR setting.
  * @return A [QuickSettingsUiState.Available] instance containing the consolidated states.
  */
+// TODO: Consider returning QuickSettingsUiState.Unavailable if all option rows (aspectRatio,
+//       captureMode, flashMode, hdr) are Unavailable due to policy restrictions or device constraints,
+//       preventing an empty bottom sheet from being presented to the user.
```
temcguir, line 40: Rather than leaving this as a TODO, can we implement the fix now? 

If all option rows (`aspectRatio`, `captureMode`, `flashMode`, `hdr`) evaluate to `Unavailable`, `QuickSettingsUiState.Companion.from` should return `QuickSettingsUiState.Unavailable`.

core/settings/src/main/java/com/google/jetpackcamera/settings/model/CameraFeaturePolicy.kt
```diff
@@ -0,0 +1,234 @@
...
+                    ImageOutputFormat.JPEG in visibility.enabledOptions
+                ) {
+                    "ImageOutputFormat.JPEG must always be included in enabledOptions for imageFormat."
+                }
+
+                is OptionVisibility.Hidden -> require(
+                    config.defaultValue == ImageOutputFormat.JPEG
+                ) {
+                    "When imageFormat is Hidden, defaultValue must be ImageOutputFormat.JPEG."
+                }
```
temcguir, line 101: Requiring `defaultValue == ImageOutputFormat.JPEG` when `Hidden` blocks a primary developer options use case: client apps that want to default or lock capture to Ultra HDR without exposing a user toggle.

While requiring `JPEG` in `enabledOptions` for `OptionVisibility.Only` makes sense as a device fallback, when an option is `Hidden`, developers should be allowed to set `defaultValue = ImageOutputFormat.JPEG_ULTRA_HDR`. If the active camera hardware does not support Ultra HDR, `CameraXCameraSystem.tryApplyImageFormatConstraints()` already gracefully demotes to JPEG at runtime.

Can we drop this check when `Hidden`?

core/settings/src/main/java/com/google/jetpackcamera/settings/model/CameraFeaturePolicy.kt
temcguir, line 119: Same here for dynamic range: dropping this check allows client apps to default or lock to HDR video (`DynamicRange.HLG10`) with the toggle hidden.

core/settings/src/test/java/com/google/jetpackcamera/settings/model/CameraFeaturePolicyTest.kt
```diff
@@ -0,0 +1,254 @@
...
+                defaultValue = FlashMode.OFF,
+                visibility = OptionVisibility.Hidden
+            )
+        )
+        assertThat(config.flashMode?.defaultValue).isEqualTo(FlashMode.OFF)
+        assertThat(config.flashMode?.visibility).isEqualTo(OptionVisibility.Hidden)
+    }
+
+    @Test
+    fun cameraFeaturePolicy_whenImageFormatHiddenAndNotJpeg_throwsException() {
```
temcguir, line 135: If we drop the requirement in `CameraFeaturePolicy` that `imageFormat` must be `JPEG` when `Hidden`, this test (and the corresponding test for `dynamicRange` on lines 159-165) should be updated to assert success instead of expecting an `IllegalArgumentException`.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewScreen.kt
```diff
@@ -600,6 +600,8 @@ private fun ContentScreen(
         }
     }
 
+    // TODO: Consider checking quickSettingsState.value is QuickSettingsUiState.Available in
+    //       isQuickSettingsVisible to hide the button when all Quick Settings rows are unavailable.
```
temcguir, line 604: In `isQuickSettingsVisible`, we should also verify `quickSettingsState.value is QuickSettingsUiState.Available` so that the toggle button is hidden when all Quick Settings rows are unavailable (paired with returning `Unavailable` in `QuickSettingsUiStateAdapter`).

ui/uistateadapter/capture/src/main/java/com/google/jetpackcamera/ui/uistateadapter/capture/CaptureModeUiStateAdapter.kt
```diff
@@ -56,37 +56,38 @@ fun CaptureModeToggleUiState.Companion.from(
     systemConstraints: CameraSystemConstraints,
     cameraAppSettings: CameraAppSettings,
     cameraState: CameraState,
-    externalCaptureMode: ExternalCaptureMode
-): CaptureModeToggleUiState =
-    if (cameraState.videoRecordingState !is VideoRecordingState.Inactive ||
-        cameraAppSettings.captureMode == CaptureMode.STANDARD
+    externalCaptureMode: ExternalCaptureMode,
+    visibilityConfig: OptionVisibility<CaptureMode>? = null
+): CaptureModeToggleUiState {
```
temcguir, line 61: `CaptureModeToggleUiState.Companion.from` still uses `visibilityConfig: OptionVisibility<CaptureMode>? = null` and is missing its `@param` documentation in the KDoc. 

Could we add the `@param` tag and align the parameter name with `CaptureModeUiState.Companion.from` (`optionVisibility` vs `visibilityConfig`)?

ui/uistateadapter/capture/src/main/java/com/google/jetpackcamera/ui/uistateadapter/capture/CaptureModeUiStateAdapter.kt
```diff
@@ -98,59 +99,88 @@ fun CaptureModeToggleUiState.Companion.from(
...
- * and a list of all available modes, each represented as a [SingleSelectableUiState].
+ * and a list of all available modes, each represented as a [SingleSelectableUiState], or
+ * [CaptureModeUiState.Unavailable] if capture mode selection is disabled or hidden.
  */
 fun CaptureModeUiState.Companion.from(
     systemConstraints: CameraSystemConstraints,
     cameraAppSettings: CameraAppSettings,
-    externalCaptureMode: ExternalCaptureMode
+    externalCaptureMode: ExternalCaptureMode,
+    optionVisibility: OptionVisibility<CaptureMode>? = null
```
temcguir, line 111: This was renamed to `optionVisibility` in commit `59f53fba`. Could we keep it consistent with `CaptureModeToggleUiState.Companion.from`?

ui/uistateadapter/capture/src/main/java/com/google/jetpackcamera/ui/uistateadapter/capture/CaptureModeUiStateAdapter.kt
```diff
@@ -257,50 +236,35 @@ private fun getAvailableCaptureModes(
...
-                        ?.supportedImageFormatsMap
-                        ?.anySupportsUltraHdr { it != affectsImageCapture } == true
-                ) {
-                    return if (affectsImageCapture) {
-                        DisabledReason.HDR_IMAGE_UNSUPPORTED_ON_SINGLE_STREAM
-                    } else {
-                        DisabledReason.HDR_IMAGE_UNSUPPORTED_ON_MULTI_STREAM
+            when (visibilityConfig) {
+                is OptionVisibility.Hidden ->
+                    return DisabledReason.IMAGE_CAPTURE_RESTRICTED
```
temcguir, line 252: Because both `CaptureModeToggleUiState.Companion.from` and `CaptureModeUiState.Companion.from` return `Unavailable` early when `OptionVisibility.Hidden`, `getCaptureModeDisabledReason` can never receive `OptionVisibility.Hidden`. 

This `is OptionVisibility.Hidden` branch for `IMAGE_ONLY` is unreachable dead code and can be removed.

ui/uistateadapter/capture/src/main/java/com/google/jetpackcamera/ui/uistateadapter/capture/CaptureModeUiStateAdapter.kt
```diff
@@ -311,40 +275,50 @@ private fun getCaptureModeDisabledReason(
                     .VIDEO_CAPTURE_EXTERNAL_UNSUPPORTED
             }
 
-            if (!hdrDynamicRangeSupported) {
-                if (systemConstraints.anySupportsHdrDynamicRange { it != currentLensFacing }) {
-                    return DisabledReason.HDR_VIDEO_UNSUPPORTED_ON_LENS
+            when (visibilityConfig) {
+                is OptionVisibility.Hidden -> {
+                    return DisabledReason.VIDEO_CAPTURE_RESTRICTED
+                }
```
temcguir, line 281: Similarly, this `is OptionVisibility.Hidden` check for `VIDEO_ONLY` is unreachable dead code and can be removed.

ui/uistateadapter/capture/src/main/java/com/google/jetpackcamera/ui/uistateadapter/capture/CaptureModeUiStateAdapter.kt
```diff
@@ -311,40 +275,50 @@ private fun getCaptureModeDisabledReason(
...
+            }
+            if (externalCaptureMode == ExternalCaptureMode.ImageCapture ||
+                externalCaptureMode == ExternalCaptureMode.MultipleImageCapture
+            ) {
+                return DisabledReason.VIDEO_CAPTURE_EXTERNAL_UNSUPPORTED
+            }
+            when (visibilityConfig) {
+                is OptionVisibility.Hidden -> {
+                    return DisabledReason.HYBRID_CAPTURE_RESTRICTED
+                }
```
temcguir, line 306: Similarly, this `is OptionVisibility.Hidden` check for `STANDARD` is unreachable dead code and can be removed.

temcguir (changes requested):

ui/uistateadapter/capture/src/main/java/com/google/jetpackcamera/ui/uistateadapter/capture/compound/QuickSettingsUiStateAdapter.kt
Kimblebee: I thought about that, but one issue is that the quick settings screen also currently the only entry point to the extended full settings screen.

I figured it may be better to just defer it since we may want to consider design and flow implications.

temcguir: Good point. One thing to keep in mind is that not every app will want a "More settings" screen, so `onNavigateToSettings` should probably be optional (`(() -> Unit)? = null`). If we do that, the button behavior could just be:

1. Has quick settings + `onNavigateToSettings` provided: opens Quick Settings (with the "More settings" button).
2. Has quick settings + no `onNavigateToSettings`: opens Quick Settings (without the "More settings" button).
3. No quick settings + `onNavigateToSettings` provided: skips the bottom sheet and calls `onNavigateToSettings()` directly.
4. No quick settings + no `onNavigateToSettings`: hide the button altogether.

I'm fine with deferring this to a follow-up if you want to capture that in the TODO.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewScreen.kt
Kimblebee: this specific request was addressed, we now verify `quickSettingsState.value` is `QuickSettingsUiState.Available` here.

However, returning ` Unavailable` in `QuickSettingsUiStateAdapter` in the case of all quick settings rows being unavailable was deferred on the basis of this [previous thread](https://github.com/google/jetpack-camera-app/pull/493#discussion_r4044227233) in this PR

temcguir (approved):

PR #499, opened 2026-04-23
Kim/docs/core camera

Kimblebee:


gemini_cli_interaction_guide.md
temcguir, line 1: I don't think you need to check this in.

temcguir (approved):
I noticed that we have a few other public classes in this module that weren't touched, such as `CameraXCameraSystem` and `InitialRecordingSettings`. Are these intentionally out of scope for this PR? If the goal is to document all public classes in `core/camera`, we should consider adding them here or in a follow-up PR.

PR #503, opened 2026-04-29
Add topStartButton parameter to PreviewLayout

davidjiagoogle:
This PR adds the topStartButton parameter to PreviewLayout and places it stacked on top of the viewfinder.

Tested with an example button:
Screenshot 2026-04-29 at 3 25 54 PM

### DIFF
--- a/ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/CaptureLayout.kt
+++ b/ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/CaptureLayout.kt
@@ -76,7 +76,8 @@ fun PreviewLayout(
     debugOverlay: @Composable (Modifier) -> Unit,
     debugVisibilityWrapper: (@Composable (@Composable () -> Unit) -> Unit),
     screenFlashOverlay: @Composable (Modifier) -> Unit,
-    snackBar: @Composable (Modifier, snackbarHostState: SnackbarHostState) -> Unit
+    snackBar: @Composable (Modifier, snackbarHostState: SnackbarHostState) -> Unit,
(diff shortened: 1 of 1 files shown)


ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/CaptureLayout.kt
```diff
@@ -102,7 +102,8 @@ fun PreviewLayout(
     debugOverlay: @Composable (Modifier) -> Unit,
     debugVisibilityWrapper: (@Composable (@Composable () -> Unit) -> Unit),
     screenFlashOverlay: @Composable (Modifier) -> Unit,
-    snackBar: @Composable (Modifier, snackbarHostState: SnackbarHostState) -> Unit
+    snackBar: @Composable (Modifier, snackbarHostState: SnackbarHostState) -> Unit,
+    topStartButton: @Composable (Modifier) -> Unit = {}
```
Kimblebee: Looks good, but I think `topStartButton` could be renamed to `topStartContent`, since technically "anything" could go there, or it could just be unused (as it currently is in this PR). Since the other component slots are bespoke, so i think its ok to be a little more generic with this slot's name

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/CaptureLayout.kt
Kimblebee, line 79: should incluide an `@param` kdoc for this new param. 

Also should specify that it is in the top start with respect for the viewfinder (since the active setting indicators occupy the _true_ top start)

Kimblebee (changes requested):

Kimblebee (approved):

PR #504, opened 2026-05-01
Refactor DI to use @BindsOptionalOf for library defaults

davidjiagoogle:
This PR refactors the dependency injection to use optional bindings for library defaults. It also removes injection for SaveMode in PreviewViewModel as it is app-level.


core/common/src/main/java/com/google/jetpackcamera/core/common/FilePathLibraryModule.kt
temcguir, line 1: This file introduces Hilt modules (`@Module`, `@InstallIn`) into `core:common`. Since `core:common` is a low-level module that many other modules depend on, adding Hilt here forces a transitive dependency on Hilt for all its consumers.

To keep `core:common` pure and reduce transient dependencies, we should avoid putting Hilt modules here.

core/common/src/main/java/com/google/jetpackcamera/core/common/JcaFilePathGenerator.kt
temcguir, line 1: Regarding moving `JcaFilePathGenerator` to `core:common`: Since this is a concrete implementation (and prefixed with `Jca`), it might be better to keep it in the `app` module.

This keeps `core:common` strictly for interfaces and generic utilities, and ensures that we don't bundle a specific default implementation into the core library. Consuming modules can just depend on the `FilePathGenerator` interface, and the `app` module can provide this concrete implementation via DI.

core/common/src/main/java/com/google/jetpackcamera/core/common/FilePathLibraryModule.kt
temcguir, line 1: Instead of using Hilt's `@BindsOptionalOf` in the library to handle defaults, we can achieve this using standard Kotlin constructor default parameters in the classes that actually need the dependency (like `CameraXCameraSystem`). This keeps the library pure and friendly for non-Hilt users.

Here is how that pattern would look:

**1. In the library class (e.g., `CameraXCameraSystem`):**
```kotlin
class CameraXCameraSystem constructor(
    // ...
    private val filePathGenerator: FilePathGenerator = StandardFilePathGenerator(), 
    // ...
) : CameraSystem { ... }
```

**2. In the `app` module (where Hilt handles assembly):**
The `@Provides` method for `CameraXCameraSystem` in the app layer can handle the override logic:
```kotlin
@Provides
@Singleton
fun providesCameraSystem(
    // ...
    customGenerator: Optional<FilePathGenerator> 
): CameraSystem {
    return CameraXCameraSystem(
        // ...
        // Use override if present, otherwise let Kotlin use the default
        filePathGenerator = customGenerator.orElse(StandardFilePathGenerator())
    )
}
```

This avoids adding Hilt modules to `core:common` entirely, reducing transient dependencies while still allowing the app to easily override the behavior.

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewViewModel.kt
```diff
@@ -90,12 +89,11 @@ private const val TAG = "PreviewViewModel"
 class PreviewViewModel @Inject constructor(
     private val cameraSystem: CameraSystem,
     private val savedStateHandle: SavedStateHandle,
-    @DefaultSaveMode private val defaultSaveMode: SaveMode,
```
temcguir, line 93: Instead of hardcoding `SaveMode.Immediate` directly in the ViewModel body, we should keep the ability to inject the default save mode so it remains configurable (e.g., for tests or different app configurations).

Instead of removing it, keep the `@DefaultSaveMode` qualifier in the constructor. We can move the definition of the `DefaultSaveMode` annotation class out of `core:common` and into this feature module (or another module visible to it). The `app` module can then continue to provide the actual binding.

data/settings/src/main/java/com/google/jetpackcamera/settings/SettingsLibraryModule.kt
temcguir, line 1: By moving and renaming `AppModule.kt` from the `app` module to `data:settings` (as `SettingsLibraryModule.kt`), this PR removes the central configuration from the application level. I recommend keeping `AppModule` at the app level to maintain central control over DI assembly.

Additionally, instead of creating this complex `@BindsOptionalOf` setup here in the library for `CaptureMode`, we can use a simpler strategy by providing the default directly in the class that consumes it.

Since `@DefaultCaptureModeOverride` is only used in `LocalSettingsRepository.kt`, we could just use a Kotlin constructor default argument there:

```kotlin
class LocalSettingsRepository @Inject constructor(
    // ...
    private val defaultCaptureModeOverride: CaptureMode = CaptureMode.STANDARD
)
```

This removes the need for the `@DefaultCaptureModeOverride` qualifier and this fallback Hilt module in the library. The `app` module can still override it by providing a binding for `CaptureMode` in its own Hilt graph.

temcguir (changes requested):

temcguir (commented):
This introduces some patterns that pull Hilt dependencies into our low-level core modules. To align with the goal of keeping our core camera library clean and free of DI, I suggest the following direction:

1. **Remove Hilt from `core:common`**: We should avoid adding new Hilt modules (like `FilePathLibraryModule`) to `core:common`. Furthermore, as a follow-up (or in this PR), we should remove the existing `CommonModule` from `core:common` and move it to the `app` module.
2. **Keep `AppModule` at the App Level**: I recommend keeping the central configuration in the `app` module instead of moving it to `data:settings`. This preserves the `app` module's role as the central assembly point.
3. **Use Kotlin Constructor Defaults**: To provide defaults for non-Hilt users without adding DI complexity, we can use standard Kotlin constructor default arguments in library classes (like `CameraXCameraSystem` and `LocalSettingsRepository`) instead of using `@BindsOptionalOf`.

I have left specific inline comments on the files with code examples to illustrate these suggestions.

PR #507, opened 2026-05-11
Decouple ui:controller:impl and ui:uistateadapter:capture from ui:components:capture

davidjiagoogle:
Breaks the transitive dependency chain to exclude heavy development/tooling dependencies when JCA is embedded as an external library.


ui/uistateadapter/capture/src/main/java/com/google/jetpackcamera/ui/uistateadapter/capture/DisabledReason.kt
temcguir, line 1: Encoding capture-specific camera constraints (like `HdrImageUnsupportedOnDevice`) directly into `ui:uistate` pollutes the generic state layer with domain-specific capture logic. 

Instead, the `DisableRationale` interface (defined in `SingleSelectableUiState.kt`) should drop the `testTag` property entirely. Since `ui:uistateadapter:capture` is the module evaluating these camera constraints, the specific `DisabledReason` implementations (and their associated string resources) should live in `ui:uistateadapter:capture`, implementing that generic `DisableRationale` interface. 

UI tests can find the disabled component and inspect its semantics or state description if they need to verify the rationale.

ui/uistateadapter/capture/src/main/res/values/strings.xml
temcguir, line 1: Because the specific `DisabledReason`s should be moved to `ui:uistateadapter:capture` where the constraints are evaluated, these string resources should also be moved to `ui:uistateadapter:capture` (e.g., `ui/uistateadapter/capture/src/main/res/values/strings.xml`). `ui:uistate` only needs to hold the generic interfaces.

ui/controller/impl/src/main/java/com/google/jetpackcamera/ui/controller/impl/CaptureControllerImpl.kt
temcguir: This file highlights why `ui:controller` was originally dependent on the components module: it is currently responsible for constructing UI state (`SnackbarData`), which forces it to know about UI concepts like test tags and string resources.

Since `CaptureController` already emits domain events (e.g., `CaptureEvent.VideoSaved`, `CaptureEvent.ImageCaptureFailed`), we should not need it to construct `SnackbarData` directly. 

We should remove the `SnackBarController` dependency from `CaptureControllerImpl` entirely. Instead, let the `PreviewViewModel` observe the `CaptureEvent` stream and construct the `SnackbarData` itself. 

Additionally, `SnackbarData` (in `SnackBarUiState.kt`) should drop the `testTag` property completely. For example, it could just hold a `SnackbarSeverity` and a `stringResource`. 

For testing, we can define a single `SNACKBAR_NODE_TAG` in `ui:components:capture`'s `TestTags.kt`. The Compose UI applies this single tag to the Snackbar component. UI tests can then find the Snackbar and verify its state by checking the text directly, rather than relying on unique tags for every single message type:

```kotlin
// Example UI test check:
composeTestRule.onNodeWithTag(SNACKBAR_NODE_TAG)
    .assertExists()
    .assertTextContains(getResString(R.string.toast_image_capture_success))
```

temcguir (changes requested):
Decoupling `ui:controller:impl` and `ui:uistateadapter:capture` from `ui:components:capture` is the right goal for allowing JCA to be embedded externally. 

However, moving test tags and UI string resources into `ui:uistate` highlights a deeper architectural issue: we are passing test tags through our state layer. Test tags in Compose are meant to identify structural UI nodes (e.g., a "Snackbar" component), not the data passing through them. 

The cleanest architecture is to remove the `testTag` property from both `SnackbarData` and `DisableRationale`. 
1. Our UI tests can find the UI element using a single static tag (like `SNACKBAR_NODE_TAG` defined in `ui:components`) and then inspect its text or semantics to verify the specific message or severity. 
2. `ui:controller` should stop constructing `SnackbarData` and just emit domain events.
3. `ui:uistateadapter:capture` should evaluate constraints and provide the string resources for the rationales.

This severs the dependencies on test tags across the board and makes the state layer generic. I have left a few inline comments showing how we can distribute these changes.

temcguir (commented):
Looks great. See my one comment on ComposeTestRule.

app/src/androidTest/java/com/google/jetpackcamera/utils/ComposeTestRuleExt.kt
temcguir, line 1: I noticed that the UI tests were updated to use `waitForNodeWithText` to globally search for the success/failure strings. This defeats the purpose of adding the `SNACKBAR_NODE_TAG` and could lead to flaky tests or false positives if that same text ever appears elsewhere on the screen.

Could you update the testing approach to ensure we are specifically inspecting the Snackbar? I recommend creating a new utility function in `ComposeTestRuleExt.kt` (e.g., `waitForSnackbarWithText`) that explicitly waits for a node matching *both* the `SNACKBAR_NODE_TAG` and the expected text.

Here is an example of what that could look like:

```kotlin
fun ComposeTestRule.waitForSnackbarWithText(
    @StringRes textResId: Int,
    timeoutMillis: Long = DEFAULT_TIMEOUT_MILLIS
) {
    val expectedText = getResString(textResId)
    waitUntil(timeoutMillis = timeoutMillis) {
        onAllNodes(hasTestTag(SNACKBAR_NODE_TAG) and hasText(expectedText))
            .fetchSemanticsNodes().isNotEmpty()
    }
}
```

You can then replace the `waitForNodeWithText` calls with `waitForSnackbarWithText` where appropriate in the test files. This will make the UI tests much more robust.

temcguir (approved):

PR #508, opened 2026-05-11
Configure Consumer Proguard Rules for UI Modules

davidjiagoogle:
Propagates custom shrinker rules to guide R8 compiler optimization downstream, enabling aggressive elimination of dead code and unused layout/viewfinder classes in consuming apps.


ui/components/capture/consumer-rules.pro
```diff
@@ -0,0 +1,4 @@
+# Allow R8 to aggressively strip unused compose layout and viewfinder classes
+-assumevalues class com.google.jetpackcamera.** {
+    *** getDebugMode() return false;
```
temcguir: There's a typo here that will prevent this rule from working. In Kotlin, a boolean property named `isDebugModeEnabled` generates a Java getter named `isDebugModeEnabled()`, not `getDebugMode()`. Since R8 operates on the bytecode, it won't find a match for this rule.

Also, applying this via the `**` wildcard is a bit dangerous for a library, as it will apply to the entire app namespace. It would be safer to target the exact class. I'd recommend updating this to:
```proguard
-assumevalues class com.google.jetpackcamera.model.DebugSettings {
    *** isDebugModeEnabled() return false;
}
```

ui/controller/impl/build.gradle.kts
```diff
@@ -31,6 +31,7 @@ android {
         lint.targetSdk = libs.versions.targetSdk.get().toInt()
 
         testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
+        consumerProguardFiles("consumer-rules.pro")
```
temcguir: As mentioned in `ui/components/capture`, we can remove this and the associated `consumer-rules.pro` file if we move the rule into `core/model`.

ui/uistateadapter/capture/build.gradle.kts
```diff
@@ -31,6 +31,7 @@ android {
         lint.targetSdk = libs.versions.targetSdk.get().toInt()
 
         testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
+        consumerProguardFiles("consumer-rules.pro")
```
temcguir: Same as above: we can drop this file and its `consumer-rules.pro` if we consolidate the rule into `core/model`.

ui/components/capture/build.gradle.kts
```diff
@@ -31,6 +31,7 @@ android {
         lint.targetSdk = libs.versions.targetSdk.get().toInt()
 
         testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
+        consumerProguardFiles("consumer-rules.pro")
```
temcguir: Since `DebugSettings` actually lives in `core/model`, we don't need to duplicate this `consumer-rules.pro` file across three different UI modules. 

A cleaner architectural approach would be to move the `consumer-rules.pro` file entirely into the `core/model` module and register it in `core/model/build.gradle.kts`. Because all of these UI modules depend on `core/model` anyway, any app that consumes them will automatically inherit the optimization rule without us having to duplicate it.

temcguir (changes requested):
The PR description mentions eliminating unused layout/viewfinder classes, but the rules being added are specifically targeting the `DebugSettings` state. I'd recommend updating the description to be a bit more accurate, something like: 

*'Propagates custom shrinker rules to downstream apps, allowing R8 to aggressively eliminate debug-specific code paths and layouts when `isDebugModeEnabled` evaluates to false in release builds.'*

temcguir (approved):

PR #514, opened 2026-05-15
Configurable Elapsed Time Updates in CaptureUiStateAdapter

temcguir:
### Problem
High-frequency updates to the elapsed time state (emitted at nanosecond precision) were causing massive recomposition overhead in the UI, specifically affecting `CaptureUiState.Ready`. This was identified as a major source of CPU contention on emulators, leading to test timeouts.

### Solution
This PR makes the time rounding precision configurable in `CaptureUiStateAdapter`. It now accepts a `timePrecision` parameter (defaulting to `SECONDS`). An internal function `roundVideoRecordingState` was extracted to handle the rounding logic for various `TimeUnit` values (Nanoseconds, Microseconds, Milliseconds, Seconds, Minutes, Hours).

### Tests
Added unit tests in `CaptureUiStateAdapterTest.kt` to verify rounding for Nanoseconds, Milliseconds, and Seconds. All tests passed.

### Impact
Reduces recomposition overhead while providing flexibility for different precision requirements in the future.


app/src/androidTest/java/com/google/jetpackcamera/utils/UiTestUtil.kt
```diff
@@ -64,15 +65,23 @@ val isEmulatorWithFakeFrontCamera: Boolean
...
+            extras.putString(MainActivity.KEY_DEBUG_SINGLE_LENS_MODE, "back")
         }
-    } else {
-        null
+
+        val resolver = InstrumentationRegistry.getInstrumentation().targetContext.contentResolver
+        val args = TestStorage(resolver).getInputArgs()
+        val disableAnimations = args["disable_animations"]?.toBoolean() ?: false
+        if (disableAnimations) {
+            extras.putBoolean("KEY_DISABLE_ANIMATIONS", true)
```
Kimblebee: is this necessary in this PR?

ui/uistateadapter/capture/src/main/java/com/google/jetpackcamera/ui/uistateadapter/capture/compound/CaptureUiStateAdapter.kt
```diff
@@ -167,9 +174,27 @@ fun captureUiState(
...
+
+/**
+ * Rounds the elapsed time of a [VideoRecordingState] to the given [timePrecision] to reduce UI recomposition frequency.
+ */
+internal fun roundVideoRecordingState(
+    videoRecordingState: VideoRecordingState,
+    timePrecision: TimeUnit
+): VideoRecordingState {
+    if (videoRecordingState !is VideoRecordingState.Active) return videoRecordingState
+
```
Kimblebee, line 193: we cna skip this if the timePrecision is already Nanoseconds.


```suggestion
if (timePrecision == TimeUnit.NANOSECONDS) return videoRecordingState    
```

ui/uistateadapter/capture/src/test/java/com/google/jetpackcamera/ui/uistateadapter/capture/CaptureUiStateAdapterTest.kt
Kimblebee, line 1: Though the pause button is currently only available on the debug screen, I think it should also check for `Active.Paused` state too since the elapsed time is still visible.

app/src/androidTest/java/com/google/jetpackcamera/utils/UiTestUtil.kt
temcguir: Good catch! This logic was accidentally included in this PR. I've removed it here and moved it to PR 519 where it belongs.

ui/uistateadapter/capture/src/main/java/com/google/jetpackcamera/ui/uistateadapter/capture/compound/CaptureUiStateAdapter.kt
temcguir: Done. Added a check to skip rounding if precision is already at nanoseconds.

ui/uistateadapter/capture/src/test/java/com/google/jetpackcamera/ui/uistateadapter/capture/CaptureUiStateAdapterTest.kt
temcguir: Done. Added a test case to verify behavior in the paused state.

ui/uistateadapter/capture/src/main/java/com/google/jetpackcamera/ui/uistateadapter/capture/compound/CaptureUiStateAdapter.kt
```diff
@@ -129,47 +137,69 @@ fun captureUiState(
...
                 trackedUiState.recentCapturedMedia,
-                cameraState.videoRecordingState
+                roundedVideoRecordingState
             ),
             screenFlashUiState = ScreenFlashUiState.from(trackedUiState)
         )
     }
 }
+
+/**
```
Kimblebee, line 185: nit: may be a good idea mention in the kdocs that the rounding logic rounds *down*  instead of the nearest second

Kimblebee (approved):

ui/uistateadapter/capture/src/main/java/com/google/jetpackcamera/ui/uistateadapter/capture/compound/CaptureUiStateAdapter.kt
temcguir: Done

PR #519, opened 2026-05-15
Implement animation disable mechanism for instrumentation tests

temcguir:
### Problem
Continuous animations (like screen flash, focus pulse, audio amplitude) keep the Compose runtime busy, causing Espresso's idle checks to fail or time out on slow emulators.

### Solution
Implemented a `LocalDisableAnimations` `CompositionLocal` to provide a flag to disable or snap animations.
- Provided from `MainActivity` based on an intent extra.
- Set in `UiTestUtil` via `TestStorage` arguments passed from the build configuration.
- Updated various components to respect this local and snap animations or use fixed states.

### Impact
Drastically reduces CPU load and Espresso timeouts by stopping background animations during tests.


app/src/main/java/com/google/jetpackcamera/MainActivity.kt
```diff
@@ -113,57 +115,62 @@ class MainActivity : ComponentActivity() {
...
-                                },
-                                onCaptureEvent = captureEventCallback
-                            )
+                            CircularProgressIndicator(modifier = Modifier.size(50.dp))
+                            Text(text = stringResource(R.string.jca_loading), color = Color.White)
+                        }
+                    }
+
+                    is Success -> {
+                        // TODO(kimblebee@): add app setting to enable/disable dynamic color
```
Kimblebee: can get rid of this todo

Kimblebee (approved):

app/src/main/java/com/google/jetpackcamera/MainActivity.kt
temcguir: Done.

PR #521, opened 2026-05-18
Add missing AndroidManifest.xml files to data:camera and ui:controller:testing

davidjiagoogle:
This PR adds missing package-less `AndroidManifest.xml` files to the following modules:
- `data/camera`
- `ui/controller/testing`

### Rationale
Downstream build systems (like Soong in AOSP platform builds) require a physical `AndroidManifest.xml` manifest file for every `android_library` target. Adding these files prevents compilation failures during platform integration, matching the convention used in other library modules.


temcguir (approved):
You can go ahead and merge.

PR #525, opened 2026-05-28
Refactor settings storage to Preferences DataStore to remove Protobuf

davidjiagoogle:
This PR refactors settings storage from Proto DataStore to Preferences DataStore. This allows us to completely remove the com.google.protobuf dependency and its generated Java code, saving ~134 KB of bytecode footprint in platform environments where dependencies are linked dynamically.


core/model/src/main/java/com/google/jetpackcamera/model/DebugSettings.kt
```diff
@@ -38,64 +32,81 @@ data class DebugSettings(
...
-            return Base64.encodeToString(protoValue.toByteArray(), Base64.NO_WRAP)
+            val lensStr = singleLensMode?.name ?: ""
+            val patternStr = when (val pattern = testPattern) {
+                is TestPattern.SolidColor ->
+                    "SolidColor(${pattern.red},${pattern.greenEven},${pattern.greenOdd},${pattern.blue})"
+                else -> pattern.toString()
+            }
+            return "debug:$isDebugModeEnabled;lens:$lensStr;pattern:$patternStr"
         }
     }
```
Kimblebee, line 111: The characters `:`, `;`, and `,` are standard URI reserved characters or delimiters. The strings should be URL-encoded before being used in navigation, and URL-decoded when parsed.

data/settings/src/main/java/com/google/jetpackcamera/settings/LocalSettingsRepository.kt
```diff
@@ -15,209 +15,141 @@
...
+                .toEnumOrDefault(VideoQuality.UNSPECIFIED),
+            audioEnabled = sharedPreferences.getBoolean(KEY_AUDIO_ENABLED, true),
+            captureMode = defaultCaptureModeOverride
+        )
+    }
+
+    private fun updateSetting(update: SharedPreferences.Editor.() -> SharedPreferences.Editor) {
+        sharedPreferences.edit().update().apply()
+        _defaultCameraAppSettings.value = readSettings()
+    }
```
Kimblebee: I think we should guard this block with a Mutex to guarantee atomic and sequential updates to the StateFlow.

data/settings/src/main/java/com/google/jetpackcamera/settings/LocalSettingsRepository.kt
```diff
@@ -15,209 +15,141 @@
...
 
-    override suspend fun updateDefaultLensFacing(lensFacing: LensFacing) {
-        jcaSettings.updateData { currentSettings ->
-            currentSettings.toBuilder()
-                .setDefaultLensFacing(lensFacing.toProto())
-                .build()
-        }
-    }
+    override suspend fun updateDefaultLensFacing(lensFacing: LensFacing) =
+        updateSetting { putString(KEY_LENS_FACING, lensFacing.name) }
```
Kimblebee: If a developer refactors the codebase and renames a Kotlin enum value (for example, renaming `LensFacing.BACK` to `LensFacing.REAR` ), all existing users' saved settings will silently fail to parse and revert to the default value.             

We should either:
- add a warning comment in the Enum classes
_or_
- define a stable serialization string/ID for each enum value instead of relying on `.name`.

core/settings/datastore-prefs/src/androidTest/java/com/google/jetpackcamera/core/settings/datastoreprefs/PrefsDataStoreSettingsDataSourceInstrumentedTest.kt
Kimblebee, line 1: we should add a test case for every setting update.

we don't have tests for:

- darkMode (tbh negligible)
- targetFrameRate
- aspectRatio
- streamConfig
- stabilizationMode
- maxVideoDuration
- videoQuality
- lowLightBoost
- audioenabled

data/settings/src/main/java/com/google/jetpackcamera/settings/LocalSettingsRepository.kt
```diff
@@ -15,209 +15,141 @@
...
+        private const val KEY_TARGET_FRAME_RATE = "target_frame_rate"
 
-    override suspend fun updateAudioEnabled(isAudioEnabled: Boolean) {
-        jcaSettings.updateData { currentSettings ->
-            currentSettings.toBuilder()
-                .setAudioEnabledStatus(isAudioEnabled)
-                .build()
+        private inline fun <reified T : Enum<T>> String?.toEnumOrDefault(default: T): T {
+            if (this == null) return default
+            return enumValues<T>().firstOrNull { it.name == this } ?: default
```
Kimblebee: This function allocates an array when called... and gets called 10x whenever a setting is updated.
i suggest using a call to `enumValueOf` protected with a try/catch to fetch the value without allocating memory for an array

```suggestion
        private inline fun <reified T : Enum<T>> String?.toEnumOrDefault(default: T): T {
            if (this == null) return default
            return try {
                                                             
                enumValueOf<T>(this)
            } catch (e: IllegalArgumentException) {
             // provide default if an entry isn't found
                default
            }
        }
```

data/settings/src/main/java/com/google/jetpackcamera/settings/LocalSettingsRepository.kt
Kimblebee, line 1: I think that removing the proto-datastore implementation was a good choice.

 think the implementation(s) for the `SettingsRepository` could just be moved into a separate sub-module  (e.g. `:data:settings-storage`). An developer could then choose to either import our current implementation or create their own.

data/settings/src/main/java/com/google/jetpackcamera/settings/LocalSettingsRepository.kt
```diff
@@ -15,209 +15,141 @@
  */
 package com.google.jetpackcamera.settings
 
-import androidx.datastore.core.DataStore
+import android.content.SharedPreferences
```
Kimblebee: I 100% agree that proto-datastore was a bit overkill for our data models.
Was the decision to use Shared Preferences over the Preferences Datastore Jetpack library intentional?

Kimblebee (changes requested):

data/settings/src/main/java/com/google/jetpackcamera/settings/LocalSettingsRepository.kt
davidjiagoogle: switched to Preferences DataStore which handles transaction safety and thread safety natively under the hood.

data/settings/src/main/java/com/google/jetpackcamera/settings/LocalSettingsRepository.kt
davidjiagoogle: Switched to datastore prefs

data/settings/src/main/java/com/google/jetpackcamera/settings/LocalSettingsRepository.kt
davidjiagoogle: Since we are using datastore-prefs, the storage implementaion is isolated in :core:settings:datastore-prefs now

data/settings/src/main/java/com/google/jetpackcamera/settings/LocalSettingsRepository.kt
davidjiagoogle: warning kdoc added

core/model/src/main/java/com/google/jetpackcamera/model/DebugSettings.kt
davidjiagoogle: Url-encoding is implemented in DebugSettingsNavType

core/settings/datastore-prefs/src/main/java/com/google/jetpackcamera/core/settings/datastoreprefs/LocalSettingsDataSource.kt
```diff
@@ -0,0 +1,168 @@
...
+import com.google.jetpackcamera.settings.model.CameraAppSettings
+import javax.inject.Inject
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.first
+import kotlinx.coroutines.flow.map
+
+/**
+ * Implementation of [SettingsDataSource] with locally stored Preferences DataStore.
+ */
+class LocalSettingsDataSource @Inject constructor(
```
temcguir: Since this module is specifically for Preferences DataStore, "LocalSettingsDataSource" is a bit too generic (e.g., a future Room implementation would also be "local"). 

Could we rename this to `PrefsDataStoreSettingsDataSource`? It's a bit verbose, but it makes it completely unambiguous what this data source is doing, especially when looking at the DI graph or stack traces. Don't forget to rename the file as well!

core/settings/datastore-prefs/build.gradle.kts
```diff
@@ -0,0 +1,91 @@
...
+        }
+    }
+}
+
+dependencies {
+    implementation(libs.kotlinx.coroutines.core)
+
+    // Hilt
+    implementation(libs.dagger.hilt.android)
+    kapt(libs.dagger.hilt.compiler)
```
temcguir: We should avoid putting Hilt dependencies in our `core:` modules. Keeping DI frameworks out of `core` makes the libraries truly reusable across different architectures (e.g., if a consuming project uses Koin or manual DI).

Please remove the Hilt dependencies and plugins from `core:settings:datastore-prefs/build.gradle.kts`, remove the `@Inject` annotation from `LocalSettingsDataSource`, and delete `DataStoreModule.kt` and `SettingsDataSourceModule.kt`.

Instead, move that wiring logic up to the `:app` module where the architectural orchestration belongs. You can create a new Hilt module in `:app` that looks roughly like this:

```kotlin
@Module
@InstallIn(SingletonComponent::class)
object AppSettingsModule {

    @Provides
    @Singleton
    fun providePreferencesDataStore(@ApplicationContext context: Context): DataStore<Preferences> {
        return PreferenceDataStoreFactory.create(
            produceFile = { context.preferencesDataStoreFile("app_settings.preferences_pb") }
        )
    }

    @Provides
    @Singleton
    fun provideSettingsDataSource(
        dataStore: DataStore<Preferences>,
        @DefaultCaptureModeOverride defaultCaptureMode: CaptureMode
    ): SettingsDataSource {
        return PrefsDataStoreSettingsDataSource(dataStore, defaultCaptureMode)
    }
}
```

core/settings/src/main/java/com/google/jetpackcamera/settings/SettingsDataSource.kt
```diff
@@ -0,0 +1,65 @@
...
+
+    suspend fun updateTargetFrameRate(targetFrameRate: Int)
+
+    suspend fun updateImageFormat(imageFormat: ImageOutputFormat)
+
+    suspend fun updateMaxVideoDuration(durationMillis: Long)
+
+    suspend fun updateVideoQuality(videoQuality: VideoQuality)
+
+    suspend fun updateAudioEnabled(isAudioEnabled: Boolean)
```
temcguir, line 110: These need kdocs

core/settings/datastore-prefs/src/androidTest/java/com/google/jetpackcamera/core/settings/datastoreprefs/LocalSettingsDataSourceInstrumentedTest.kt
```diff
@@ -0,0 +1,249 @@
...
+import org.junit.After
+import org.junit.Before
+import org.junit.Rule
+import org.junit.Test
+import org.junit.rules.TemporaryFolder
+import org.junit.runner.RunWith
+
+@OptIn(ExperimentalCoroutinesApi::class)
+@RunWith(AndroidJUnit4::class)
+class LocalSettingsDataSourceInstrumentedTest {
```
temcguir: To match the implementation rename suggested above, let's also rename this test class (and its filename) to `PrefsDataStoreSettingsDataSourceInstrumentedTest`.

core/settings/datastore-prefs/build.gradle.kts
```diff
@@ -0,0 +1,91 @@
...
+    androidTestImplementation(libs.androidx.junit)
+    androidTestImplementation(libs.truth)
+    androidTestImplementation(libs.kotlinx.coroutines.test)
+    androidTestImplementation(project(":core:settings:datastore-prefs:testing"))
+
+    // Access Model and Settings Interface
+    implementation(project(":core:model"))
+    implementation(project(":core:common"))
+    implementation(project(":core:settings"))
+    implementation(project(":data:settings"))
```
temcguir: Please remove `implementation(project(":data:settings"))`. The storage implementation shouldn't depend on our app's Repositories. Since the models and interfaces moved to `core:settings`, this is unused and breaks the architectural separation.

core/settings/datastore-prefs/src/main/java/com/google/jetpackcamera/core/settings/datastoreprefs/LocalSettingsDataSource.kt
```diff
@@ -0,0 +1,168 @@
...
+        CameraAppSettings(
+            cameraLensFacing = prefs[PreferenceKeys.KEY_LENS_FACING]
+                .toEnumOrDefault(LensFacing.BACK),
+            darkMode = prefs[PreferenceKeys.KEY_DARK_MODE].toEnumOrDefault(DarkMode.DARK),
+            flashMode = prefs[PreferenceKeys.KEY_FLASH_MODE].toEnumOrDefault(FlashMode.OFF),
+            aspectRatio = prefs[PreferenceKeys.KEY_ASPECT_RATIO]
+                .toEnumOrDefault(AspectRatio.NINE_SIXTEEN),
+            stabilizationMode = prefs[PreferenceKeys.KEY_STABILIZATION_MODE]
+                .toEnumOrDefault(StabilizationMode.AUTO),
+            targetFrameRate = prefs[PreferenceKeys.KEY_TARGET_FRAME_RATE] ?: 0,
```
temcguir: nit: Instead of hardcoding `0` here, we should use the constant `TARGET_FPS_AUTO` from the model to keep the fallback explicitly bound to the domain logic.

temcguir (changes requested):

core/model/src/main/java/com/google/jetpackcamera/model/DebugSettings.kt
```diff
@@ -38,64 +32,81 @@ data class DebugSettings(
...
-                isDebugModeEnabled = proto.isDebugModeEnabled,
-                singleLensMode = if (proto.hasSingleLensMode()) {
-                    LensFacing.fromProto(proto.singleLensMode)
-                } else {
-                    null
-                },
-                testPattern = TestPattern.fromProto(proto.testPattern)
-            )
-        }
+        fun parseFromString(value: String): DebugSettings {
```
temcguir, line 37: Since we are replacing Protobuf with a custom, hand-rolled string parser, we need unit tests to ensure this logic is robust. String parsing based on delimiters (`:`, `;`, `,`) and specific string formats (like `SolidColor(...)`) is prone to regressions if the model changes in the future.

Could you please add a `DebugSettingsTest.kt` in the `core:model` module? It should verify that `parseFromString(encodeAsString())` correctly rebuilds the object. Specifically, we should test:
1. Default values (all fields empty/off).
2. Simple `TestPattern` variants (like `ColorBars`).
3. The complex `TestPattern.SolidColor` variant to ensure the regex/splitting of the 4 color channels works correctly.
4. Edge cases like missing delimiters or malformed strings (to ensure it falls back gracefully).

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/navigation/DebugSettingsNavType.kt
```diff
@@ -31,27 +33,32 @@ internal object DebugSettingsNavType : NavType<DebugSettings>(isNullableAllowed
...
+            DebugSettings.parseFromString(str)
         }
     }
 
     /**
-     * Parses the Base64 encoded Proto string from the navigation route.
+     * Parses the URL-encoded serialized string from the navigation route.
      */
-    override fun parseValue(value: String): DebugSettings = DebugSettings.parseFromString(value)
+    override fun parseValue(value: String): DebugSettings {
```
temcguir, line 51: Similarly, we should add a quick unit test (`DebugSettingsNavTypeTest.kt`) to verify the URL encoding and decoding steps here. We want to ensure that `parseValue(serializeAsValue(debugSettings))` returns the original object and that the resulting string is safe for Compose Navigation routes.

temcguir (approved):

PR #527, opened 2026-06-02
Refactor JCA to extract OpenGL/EGL effects into dedicated :core:camera:effects module

davidjiagoogle:
This PR modularizes the OpenGL/EGL surface-processing and single-stream effects in the Jetpack Camera App (JCA), separating them into a dedicated `:core:camera:effects` module.

### Key Changes:
1. **New Interface**: Defined `SingleStreamEffectProvider` in `:core:camera` as the injection contract.
2. **Modularized Effects**: Moved the `effects/` package hierarchy (including `SingleSurfaceForcingEffect`, `CopyingSurfaceProcessor`, shaders, and GL helper classes) from `:core:camera` to `:core:camera:effects`.
3. **Decoupled Dependencies**: Transferred `libs.androidx.graphics.core` (OpenGL/EGL wrapper library) to `:core:camera:effects`'s `build.gradle.kts`.
4. **Hilt Injection**: Registered `SingleStreamEffectProviderImpl` using Hilt `@IntoSet` multi-binding. `CameraXCameraSystem` now injects `Set<SingleStreamEffectProvider>` and passes the effect provider to `CameraSession` dynamically.
5. **Configurable Exclusion**: By separating effects into a new module, we enable build configurations (like "lite" builds) to completely exclude the `:core:camera:effects` dependency, preventing the `androidx.graphics:graphics-core` library (~15KB) from being compiled and bundled into the production APK.

Tried out removing effects module in the app layer in the david/splitEffectsTryout branch


core/camera/effects/single-stream/src/main/java/com/google/jetpackcamera/core/camera/effects/EffectsModule.kt
```diff
@@ -0,0 +1,33 @@
...
+
+@Module
+@InstallIn(SingletonComponent::class)
+interface EffectsModule {
+    @Binds
+    @IntoSet
+    fun bindSingleStreamEffectProvider(
+        impl: SingleStreamEffectProviderImpl
+    ): SingleStreamEffectProvider
+}
```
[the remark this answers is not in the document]
davidjiagoogle: Dagger does not allow satisfying an @BindsOptionalOf defined in a parent component (SingletonComponent) from a subcomponent (ActivityRetainedComponent).

core/camera/src/main/java/com/google/jetpackcamera/core/camera/SingleStreamEffectProvider.kt
```diff
@@ -0,0 +1,31 @@
...
+ */
+package com.google.jetpackcamera.core.camera
+
+import androidx.camera.core.CameraEffect
+import kotlinx.coroutines.CoroutineScope
+
+/**
+ * Provider for the camera effect that forces a single surface for use cases.
+ */
+interface SingleStreamEffectProvider {
```
temcguir: Since the goal is to establish a pattern where users can inject their own arbitrary effects, the name `SingleStreamEffectProvider` is a bit too tied to our internal "single stream" effect. Let's rename this to something generic like `CameraEffectProvider` so it reads like a proper API for custom effects.

data/camera/src/main/java/com/google/jetpackcamera/data/camera/CameraSessionOptionalModule.kt
```diff
@@ -0,0 +1,29 @@
...
+import com.google.jetpackcamera.core.camera.SingleStreamEffectProvider
+import dagger.BindsOptionalOf
+import dagger.Module
+import dagger.hilt.InstallIn
+import dagger.hilt.components.SingletonComponent
+
+@Module
+@InstallIn(SingletonComponent::class)
+interface CameraSessionOptionalModule {
+    @BindsOptionalOf
```
temcguir: The PR description mentions using a Hilt `@IntoSet` multi-binding, but it looks like we're actually using `@BindsOptionalOf` to inject a single `java.util.Optional` here. 

Because this only injects a single instance, if developers want to inject their own custom effect (like a color filter or AR effect), they would have to completely override and replace our `EffectsModule`. It prevents multiple effects from coexisting.

To truly allow arbitrary effects without adding Hilt dependencies to `:core:camera`, we should match the Map-based registry pattern we already use for `LowLightBoost` and `ImagePostProcessor`:

1. Create a `CameraEffectFeatureKey` interface.
2. Delete this `CameraSessionOptionalModule` and replace it with a new module right here in `:data:camera` (e.g., `CameraEffectModule`) that uses `@Multibinds` to provide an empty Map or Set of `Map.Entry<CameraEffectFeatureKey, Provider<CameraEffectProvider>>` by default.
3. Have the `:core:camera:effects` module use `@IntoSet` to contribute its specific implementation to that Map.
4. Update `CameraModule` and `CameraXCameraSystem` to simply accept a `Map<CameraEffectFeatureKey, Provider<CameraEffectProvider>>`.

Because the `@Multibinds` module inside `:data:camera` guarantees the Map will always exist (even if it's empty when `:core:camera:effects` is excluded from a lite build), we completely eliminate the need for `Optional` injection and cleanly support any number of custom effects.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraSystem.kt
```diff
@@ -110,7 +110,8 @@ class CameraXCameraSystem(
     effectProviders:
     Map<LowLightBoostFeatureKey, @JvmSuppressWildcards Provider<LowLightBoostEffectProvider>>,
     val imagePostProcessors:
-    Map<ImagePostProcessorFeatureKey, @JvmSuppressWildcards Provider<ImagePostProcessor>>
+    Map<ImagePostProcessorFeatureKey, @JvmSuppressWildcards Provider<ImagePostProcessor>>,
+    private val singleStreamEffectProvider: SingleStreamEffectProvider?
```
temcguir: Following up on the injection comment in `CameraSessionOptionalModule`, this parameter should be updated to accept the Map pattern:
`cameraEffectProviders: Map<CameraEffectFeatureKey, @JvmSuppressWildcards Provider<CameraEffectProvider>>`

Having this map acts as a registry of available effects, which elegantly allows us to treat the active effect as a configurable setting rather than hardcoding it here. 

To complete this architecture, we should:
1. Add an `activeCameraEffect: CameraEffectFeatureKey?` field to `CameraAppSettings` (and `SessionSettings`).
2. Have `CameraSession` check `sessionSettings.activeCameraEffect`, look up the matching provider in this injected map, and apply only the selected effect to the `UseCaseGroup`.

This gives us a clean, unidirectional data flow where the UI can change the effect via the `SettingsRepository`, and the camera session automatically rebuilds with the correct effect from the registry. Note: It's totally fine to keep this map separate from `lowLightBoostEffectProvider`, since Low Light Boost requires complex UI lifecycle callbacks that generic effects don't need.

temcguir (changes requested):
The overall direction of moving effects into a dedicated module and injecting them to allow for "lite" builds is solid. However, there are a few architectural issues with the current implementation that prevent it from being truly extensible for arbitrary custom effects.

Most notably, the PR description states that this uses Hilt `@IntoSet` multi-binding, but the code actually relies on `@BindsOptionalOf` to inject a single `Optional<SingleStreamEffectProvider>`. This single-instance injection prevents multiple effects from coexisting and forces developers to completely overwrite the module if they want to inject their own custom effect.

To resolve this, I've left a few inline comments outlining an approach that aligns with our existing `LowLightBoost` and `ImagePostProcessor` implementations:

1. **Generalize the interface**: Rename `SingleStreamEffectProvider` to a generic `CameraEffectProvider`.
2. **Use a Map Registry**: Replace `CameraSessionOptionalModule` with a `@Multibinds` module inside `:data:camera` to provide a `Map<CameraEffectFeatureKey, Provider<CameraEffectProvider>>`. This guarantees the Map is always present (even if empty in lite builds) and completely eliminates the need for `Optional` injection.
3. **Drive via Settings**: Treat the active effect as a configurable UI state by adding it to `CameraAppSettings` (and `SessionSettings`), rather than hardcoding it in `startCamera`. `CameraSession` can then simply use the setting state to look up the correct provider from the injected Map registry.

feature/settings/src/main/java/com/google/jetpackcamera/settings/ui/SettingsComponents.kt
```diff
@@ -351,59 +350,59 @@ fun AspectRatioSetting(
...
                         ),
-                        selected = streamConfigUiState.currentStreamConfig ==
-                            StreamConfig.SINGLE_STREAM,
+                        selected = cameraEffectUiState.currentCameraEffect.isEmpty() ||
+                            cameraEffectUiState.currentCameraEffect == "none",
                         enabled = true,
-                        onClick = { setStreamConfig(StreamConfig.SINGLE_STREAM) }
+                        onClick = { setCameraEffect("none") }
                     )
+                    if (cameraEffectUiState.supportedEffects.contains("single_stream")) {
```
temcguir: Note: By checking for the exact raw string "single_stream" here, we are unfortunately defeating the purpose of the generic effect registry we just built. We should export the effect ID constants (like SingleStreamEffectKey.id) to a shared constants file so the UI module can reference them safely without relying on hardcoded magic strings.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraXCameraSystem.kt
```diff
@@ -269,12 +271,13 @@ class CameraXCameraSystem(
                                 supportedFixedFrameRates = supportedFixedFrameRates,
                                 supportedDynamicRanges = supportedDynamicRanges,
                                 supportedImageFormatsMap = mapOf(
```
temcguir, line 274: Currently, the codebase assumes that if *any* effect is active (`selectedCameraEffect != "none"`), it forces a "single-stream layout" and we must disable Ultra HDR. We see this with the `Map<Boolean, Set<ImageOutputFormat>>` here, and the duplicated `!= "none"` checks across the UI Adapters.

However, if we ever add an effect that only targets `PREVIEW` (e.g. an AR viewfinder overlay), `IMAGE_CAPTURE` wouldn't be affected and Ultra HDR could theoretically remain on. 

To future-proof this (and clean up duplicated magic string checks), I'd recommend an "Effect Capabilities" model:
1. Let `CameraEffectProvider` expose the CameraX `targets` it requires (e.g., `CameraEffect.PREVIEW | CameraEffect.IMAGE_CAPTURE`).
2. Replace this boolean map with a function like `fun getSupportedImageFormats(effectTargets: Int)`. If the active effect's targets include `IMAGE_CAPTURE`, we return `JPEG`. Otherwise, we can still support `JPEG_ULTRA_HDR`.
3. This allows us to remove the `selectedCameraEffect != "none"` string checks from `CaptureModeUiStateAdapter` and `HdrUiStateAdapter`, as they can just query the constraints dynamically based on the active effect's targets.

settings.gradle.kts
```diff
@@ -40,6 +40,7 @@ include(":data:camera")
 include(":core:camera:testing")
 include(":core:camera:low-light")
 include(":core:camera:low-light-playservices")
+include(":core:camera:effects")
```
temcguir: **Nit: Module Naming**

The name `:core:camera:effects` implies that this module contains the generic, base infrastructure for the entire effects system. However, the core effect interfaces (like `CameraEffectFeatureKey`) actually live in the base `:core:camera` module, meaning this new module is strictly just the implementation for the `SingleStreamEffect`.

Since our goal here is to make the single stream effect an optional dependency that apps can exclude to reduce APK size, naming it something more specific like `:core:camera:effects:single-stream` would make it much clearer to developers that this is a specific, removable plugin rather than mandatory core infrastructure.

temcguir (changes requested):

build.log
temcguir, line 1: I think this isn't meant to be here.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/effects/CameraEffectFeatureKey.kt
```diff
@@ -0,0 +1,30 @@
...
+interface CameraEffectFeatureKey {
+    val id: String
+}
+
+/**
+ * Key for the single stream effect.
+ */
+object SingleStreamEffectKey : CameraEffectFeatureKey {
+    override val id: String = "single_stream"
+}
```
temcguir: Since this specific key isn't actually used anywhere in `core:camera` (the camera system just queries the registry map by string ID), it shouldn't live in this core module. Let's move `SingleStreamEffectKey` entirely into the `:core:camera:effects:single-stream` module.

feature/settings/src/main/java/com/google/jetpackcamera/settings/ui/SettingsComponents.kt
```diff
@@ -351,59 +350,59 @@ fun AspectRatioSetting(
...
+                                BTN_DIALOG_STREAM_CONFIG_OPTION_SINGLE_STREAM_TAG
+                            ),
+                            text = stringResource(
+                                id = R.string.stream_config_selector_single_stream
+                            ),
+                            selected = cameraEffectUiState.currentCameraEffect == "single_stream",
+                            enabled = true,
+                            onClick = { setCameraEffect("single_stream") }
+                        )
+                    }
```
temcguir: We're still using the magic string `"single_stream"` here and in the viewmodel tests. 
Let's add a dependency on `:core:camera:effects:single-stream` to this feature's `build.gradle`, and replace these hardcoded strings with `SingleStreamEffectKey.id`. Since `feature:settings` is an app-level module, it's perfectly fine for it to depend directly on the implementation module to wire up the UI, and it prevents us from relying on magic strings.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/effects/CameraEffectFeatureKey.kt
```diff
@@ -0,0 +1,30 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.jetpackcamera.core.camera.effects
+
+/**
+ * A key used to identify a specific implementation of a camera effect.
+ */
+interface CameraEffectFeatureKey {
+    val id: String
```
temcguir: To give us strict type safety and completely eliminate magic strings across the app, let's update this to use a Kotlin value class instead of raw Strings. 

Could you replace the interface and add a `NONE` constant like this:
```kotlin
@JvmInline
value class CameraEffectId(val value: String)

interface CameraEffectFeatureKey {
    val id: CameraEffectId
}

val NONE_EFFECT_ID = CameraEffectId("none")

File: core/settings/src/main/java/com/google/jetpackcamera/settings/model/CameraAppSettings.kt [hunk 11]
```diff
@@ -42,7 +41,7 @@ data class CameraAppSettings(
     val cameraLensFacing: LensFacing = LensFacing.BACK,
     val darkMode: DarkMode = DarkMode.DARK,
     val flashMode: FlashMode = FlashMode.OFF,
-    val streamConfig: StreamConfig = StreamConfig.MULTI_STREAM,
+    val selectedCameraEffect: String = "",
```
temcguir: Let's update this property to use our new type-safe wrapper so it can never be assigned an invalid string:
`val selectedCameraEffect: CameraEffectId = NONE_EFFECT_ID`
*(Note: When you read/write this to the Preferences Datastore in `LocalSettingsRepository`, you'll just need to unwrap it by calling `.value` when saving, and wrap it in `CameraEffectId(...)` when reading).*

feature/settings/src/main/java/com/google/jetpackcamera/settings/ui/SettingsComponents.kt
```diff
@@ -351,59 +350,59 @@ fun AspectRatioSetting(
...
-                            id = R.string.stream_config_description_single_stream
+                            id = R.string.stream_config_selector_multi_stream
                         ),
-                        selected = streamConfigUiState.currentStreamConfig ==
-                            StreamConfig.SINGLE_STREAM,
+                        selected = cameraEffectUiState.currentCameraEffect.isEmpty() ||
+                            cameraEffectUiState.currentCameraEffect == "none",
                         enabled = true,
-                        onClick = { setStreamConfig(StreamConfig.SINGLE_STREAM) }
+                        onClick = { setCameraEffect("none") }
```
temcguir: With the new value class in place, we can eliminate the magic strings here. The `"none"` block can be updated to compare against our new constant:

```kotlin
selected = cameraEffectUiState.currentCameraEffect == NONE_EFFECT_ID,
enabled = true,
onClick = { setCameraEffect(NONE_EFFECT_ID) }

temcguir (approved):

PR #530, opened 2026-06-09
Refactor Quick Settings UI

Kimblebee:
## Summary

Refactors the Quick Settings UI to a flat connected button row layout using Material 3 Expressive `ToggleButton` groups. Improves conflict handling between HDR and Low Light Boost / Dual Camera mode, modernizes the test suite, and cleans up unused resources.

---

## Key Changes

* **Flat Button Row UX:** Replaces expandable sub-screens (Aspect Ratio, Capture Mode) with flat `SettingRow` / `QuickSettingsListRow` components using `ButtonGroupDefaults` connected shapes. Removed "Flip Camera" from Quick Settings (available directly on preview).
* **Conflict Resolution & State Handling:** Unsupported flash modes (such as Low Light Boost when HDR is active) now remain visible as disabled with a clear rationale (`LLB_DISABLED_BY_HDR`) rather than disappearing.
* **Architecture & Cleanup:**
  * Removed the obsolete `FocusedQuickSetting` enum and `setFocusedSetting()` controller method.
  * Deleted `NoOpQuickSettingsController` and unused drawables/strings.
  * Scoped state adapter functions (`FlashModeUiState.from`, `HdrUiState.from`) to `internal`.

---

### 1. Test Tags (Impacts Dependents & Test Automation)

| Removed Tag / Constant | New Replacement Tag |
| :--- | :--- | 
| `QUICK_SETTINGS_HDR_BUTTON` | `BTN_QUICK_SETTINGS_HDR_OPTION_ON` / `_OFF` | 
| `BTN_QUICK_SETTINGS_FOCUS_CAPTURE_MODE` | `BTN_QUICK_SETTINGS_CAPTURE_MODE_OPTION_*` | 
| `QUICK_SETTINGS_FLASH_BUTTON` | `BTN_QUICK_SETTINGS_FLASH_OPTION_*` | 
| `QUICK_SETTINGS_FLIP_CAMERA_BUTTON` | `FLIP_CAMERA_BUTTON` (main UI) | 
| `QUICK_SETTINGS_RATIO_BUTTON` | `QUICK_SETTINGS_RATIO_*_BUTTON` / `ROW_QUICK_SETTINGS_ASPECT_RATIO` | 
| `QUICK_SETTINGS_CLOSE_EXPANDED_BUTTON` | *None* (sub-panels removed) |

### 2. Public Signatures & State Models
* `QuickSettingsController`: Removed `setFocusedSetting(focusedQuickSetting: FocusedQuickSetting)`.
* `QuickSettingsUiState.Available` & `TrackedCaptureUiState`: Removed `focusedQuickSetting` parameter.
* `QuickSettingsBottomSheet`: Now takes `onEvent: (QuickSettingsEvent) -> Unit` instead of `QuickSettingsController`.
* `CaptureModeToggleButton`: Now takes `onChangeCaptureMode` and `onToggleWhenDisabled` lambdas.

---

## Testing

* **Unit Tests:**
  * Added `QuickSettingsControllerImplTest` and `QuickSettingsUiStateAdapterTest`
  * Updated `CaptureUiStateAdapterTest`, `FlashModeUiStateAdapterTest`, and `HdrUiStateAdapterTest` 
* **Instrumented Tests:**
   * Updated `ComposeTestRuleExt.kt`, `CaptureModeSettingsTest`, `ConcurrentCameraTest`, `NavigationTest`, and `SwitchCameraTest` to target the flat button row layout.

Screenshot 2026-07-29 at 10 22 23 AM

Screenshot 2026-07-29 at 10 21 33 AM


ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/quicksettings/ui/QuickSettingsComponents.kt
```diff
@@ -499,136 +239,263 @@ fun QuickSettingsBottomSheet(
...
-    // It handles the "overflow to the right" behavior by default.
-    LazyRow(
+    Row(
         modifier = modifier
-            .testTag(QUICK_SETTINGS_SCROLL_CONTAINER)
-            .fillMaxWidth(),
-        horizontalArrangement = Arrangement.spacedBy(8.dp, Alignment.CenterHorizontally),
-        contentPadding = PaddingValues(horizontal = 16.dp)
+            .fillMaxWidth()
+            .padding(vertical = 12.dp, horizontal = 16.dp),
```
temcguir: The vertical padding here is a bit too large compared to the mocks. The mocks specify a `4.dp` vertical padding (and `16.dp` horizontal) for the settings rows. 

Could you update this to `padding(vertical = 4.dp, horizontal = 16.dp)`?

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/quicksettings/ui/QuickSettingsComponents.kt
```diff
@@ -499,136 +239,263 @@ fun QuickSettingsBottomSheet(
...
+                .semantics(mergeDescendants = true) {}
+        ) {
+            Text(
+                text = title,
+                style = MaterialTheme.typography.titleMedium,
+                color = MaterialTheme.colorScheme.onSurface
+            )
+            Text(
+                text = stateSubtitle,
+                style = MaterialTheme.typography.bodyMedium,
```
temcguir: The typography styles for the title and subtitle seem to be swapped and slightly off from the design specs. 

According to the mocks:
- The Row Title (`title`) should be `MaterialTheme.typography.labelMedium` (12px, Medium).
- The Row Subtitle (`stateSubtitle`) should be `MaterialTheme.typography.titleMedium` (16px, SemiBold).

Currently, this is using `titleMedium` for the top text and `bodyMedium` for the subtext. Could you update these styles to match the spec?

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/quicksettings/ui/QuickSettingsComponents.kt
```diff
@@ -698,55 +566,12 @@ private fun QuickSettingToggleButton(
     }
```
temcguir: We shouldn't be rendering individual text labels underneath each toggle icon. The mocks consolidate the text description entirely into the row's title and subtitle stack on the left side.

Let's remove this `Spacer` and `Text` composable entirely so we just show the unlabelled icons on the right. (This will also mean we can clean up the `text` parameter being passed into this component).

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/quicksettings/ui/QuickSettingsComponents.kt
temcguir: The shapes and container colors for the toggle buttons aren't quite matching the custom styling in the mocks. The design calls for a dynamic shape and color depending on the selection state:
- **Unselected:** Should be perfectly circular (e.g., `CircleShape`) with a 20% white opacity for the container color.
- **Selected:** Should morph into a rounded rectangle (`RoundedCornerShape(12.dp)`) using the primary color.

Could we update the `shapes` and `colors` parameters to dynamically use these values instead of the default `IconButtonDefaults`?

ui/controller/impl/src/main/java/com/google/jetpackcamera/ui/controller/impl/QuickSettingsControllerImpl.kt
temcguir: Do we need to take `externalCaptureMode` as a dependency constructor parameter here?

Upstream state resolution already applies `externalCaptureMode` on the `CameraSystem` defaults, and `HdrUiStateAdapter` accurately reacts to it by emitting omitting/restricting the UI state. Because the UI adapter is already preventing interaction under these conditions, repeating defensive guards here in the controller is redundant.

We should be able to remove `externalCaptureMode` entirely from this file (and its fake) and just blindly forward changes to the `CameraSystem` – letting the `UiStateAdapter`s handle correct UI visibility/blocking.

ui/uistateadapter/capture/src/main/java/com/google/jetpackcamera/ui/uistateadapter/capture/FlashModeUiStateAdapter.kt
```diff
@@ -27,127 +28,169 @@ import com.google.jetpackcamera.ui.uistate.SingleSelectableUiState
...
+                continue // Hide LLB if Dual Camera is active
+            }
+        }
+
+        // 4. Determine if Enabled or Disabled
+        val isLlbHdrConflict = mode == FlashMode.LOW_LIGHT_BOOST && isHdrOn
+
+        if (currentLensSupportedFlashModes.contains(mode) && !isLlbHdrConflict) {
+            displayableModes.add(SingleSelectableUiState.SelectableUi(mode)) // Enabled
+        } else {
```
temcguir: Since `LOW_LIGHT_BOOST` is hardcoded into `ORDERED_UI_SUPPORTED_FLASH_MODES`, this loop successfully catches when a device's lens doesn't physically support LLB hardware. 

However, by emitting a `SingleSelectableUiState.Disabled` anyway, we accidentally force the UI to render a permanently grayed-out LLB button onto the screen. This needlessly clutters the interface with an unusable option for devices that don't even have the required LLB hardware/software options.

Could we conditionally skip `displayableModes.add` entirely if `reason == DisabledReason.FLASH_UNSUPPORTED_ON_LENS`, so unsupported hardware options are seamlessly hidden instead of permanently disabled?

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/quicksettings/QuickSettingsScreen.kt
```diff
@@ -15,281 +15,234 @@
...
+@Composable
+private fun HybridQuickSettings(
+    quickSettingsUiState: QuickSettingsUiState.Available,
+    quickSettingsController: QuickSettingsController,
+    onNavigateToSettings: () -> Unit,
+    showMoreSettingsButton: Boolean
+) {
+    Column {
+        Text(
+            modifier = Modifier.padding(16.dp),
```
temcguir: There's a minor padding issue here for the sheet titles. Compose's Material 3 `ModalBottomSheet` inherits a default top padding to accommodate the drag handle. Because this `Text` composable has a symmetric `16.dp` padding, it adds 16dp of *extra* padding underneath the drag handle, pushing the title awkwardly far down. 

To perfectly match the 40dp visual gap in the mocks from the *very top* of the sheet container (and just 8dp below the title), we can just let the drag handle handle the top sizing, and only provide horizontal/bottom padding here:

`modifier = Modifier.padding(start = 16.dp, end = 16.dp, bottom = 8.dp)`

(This applies to all three sheet composables: Photo, Video, and combined!)

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/quicksettings/QuickSettingsScreen.kt
```diff
@@ -15,281 +15,234 @@
...
-
-                FocusedQuickSetting.CAPTURE_MODE -> focusedCaptureModeButtons(
-                    onUnFocus = onUnFocus,
-                    onSetCaptureMode = quickSettingsController::setCaptureMode,
-                    captureModeUiState = quickSettingsUiState.captureModeUiState
-                )
+    if (quickSettingsUiState is QuickSettingsUiState.Available &&
+        quickSettingsUiState.quickSettingsIsOpen
+    ) {
+        val sheetState = rememberModalBottomSheetState()
```
temcguir: By default, `rememberModalBottomSheetState()` sets `skipPartiallyExpanded = false`. Since our Quick Settings content extends slightly past 50% of the screen height, it currently opens into a "Partially Expanded" resting state that cuts off the "More Settings" controls at the bottom until the user manually pulls the sheet the rest of the way up.

We can fix this by forcing it to measure and reveal its full height automatically:

`val sheetState = rememberModalBottomSheetState(skipPartiallyExpanded = true)`

ui/uistateadapter/capture/src/main/java/com/google/jetpackcamera/ui/uistateadapter/capture/compound/QuickSettingsUiStateAdapter.kt
```diff
@@ -32,13 +31,9 @@ import com.google.jetpackcamera.ui.uistate.capture.compound.QuickSettingsUiState
...
- * @param cameraAppSettings The current application settings for the camera.
- * @param systemConstraints The constraints of the camera system.
  * @param aspectRatioUiState The UI state for the aspect ratio setting.
  * @param hdrUiState The UI state for the HDR setting.
  * @param quickSettingsIsOpen Indicates whether the quick settings panel is open.
- * @param focusedQuickSetting The currently focused quick setting, if any.
- * @param externalCaptureMode The external capture mode, if any.
  * @return A [QuickSettingsUiState.Available] instance containing the consolidated states.
  */
 fun QuickSettingsUiState.Companion.from(
```
temcguir, line 39: I noticed we are missing a specific test file (`QuickSettingsUiStateAdapterTest.kt`) to verify this output. 

Could we add a test to verify that the companion `from` method is continuing to properly map its injected states (aspectRatioUiState, captureModeUiState, flashModeUiState, flipLensUiState, hdrUiState)?

ui/controller/impl/src/main/java/com/google/jetpackcamera/ui/controller/impl/QuickSettingsControllerImpl.kt
temcguir, line 1: Can you make sure we have unit test coverage in `QuickSettingsControllerImplTest.kt` to ensure the standard toggles correctly mutate state against Fake dependencies?

ui/components/capture/src/main/res/values/strings.xml
```diff
@@ -120,9 +110,29 @@
...
     <!-- More Settings Text -->
-    <string name="quick_settings_more_text">More</string>
+    <string name="quick_settings_more_text">More Settings</string>
     <string name="quick_settings_more_description">Settings icon</string>
 
     <!-- ImageWell -->
     <string name="image_well_content_description">View recently saved media</string>
+    <string name="quick_settings_title_video_settings">Video Settings</string>
+    <string name="quick_settings_title_photo_settings">Photo Settings</string>
+    <string name="quick_settings_title_photo_and_video_settings">Photo and Video Settings</string>
```
temcguir: Minor phrasing nit: `quick_settings_title_photo_settings` uses capital "Settings" ("Photo Settings"), but the visual mocks use a lowercase "s" ("Photo settings"). We should update the casing to strictly match.

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/quicksettings/ui/QuickSettingsComponents.kt
```diff
@@ -863,34 +691,75 @@ private fun QuickSettingToggleButtonPreview() {
...
+@Preview(name = "JCA Setting Row - Dark Mode", showBackground = true, backgroundColor = 0xFF000000)
+@Composable
+private fun PreviewSettingRowDark() {
+    MaterialTheme(colorScheme = darkColorScheme()) {
+        Surface(
+            modifier = Modifier.fillMaxWidth(),
+            color = Color.Black // Consistent with Camera UI
+        ) {
+            SettingRow(
+                title = "Video Resolution",
```
temcguir: Do we not have string resources for all of the strings in this composable `@Preview`? If so, we should use them rather than using string literals.

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/quicksettings/QuickSettingsScreen.kt
```diff
@@ -15,281 +15,234 @@
...
-                        SingleSelectableUiState.SelectableUi(LensFacing.BACK),
-                        SingleSelectableUiState.SelectableUi(LensFacing.FRONT)
-                    )
-                ),
-                hdrUiState = HdrUiState.Unavailable,
-                quickSettingsIsOpen = true
-            ),
-            onNavigateToSettings = {},
-            quickSettingsController = NoOpQuickSettingsController()
+private fun VideoQuickSettings(
```
temcguir: Since `ImageAndVideoQuickSettings`, `VideoQuickSettings`, and `ImageQuickSettings` share the exact same structural boilerplate (a title, a `<Row>` list, and a conditional `QuickNavSettings` button at the bottom), could we consolidate them into a shared base `<QuickSettingsLayout>` composable?

For example:
```kotlin
@Composable
private fun QuickSettingsLayout(
    @StringRes titleRes: Int,
    showMoreSettingsButton: Boolean,
    onNavigateToSettings: () -> Unit,
    content: @Composable ColumnScope.() -> Unit
) {
    Column {
        Text(
            modifier = Modifier.padding(start = 16.dp, end = 16.dp, bottom = 8.dp),
            text = stringResource(id = titleRes),
            style = MaterialTheme.typography.titleLarge
        )
        content()
    }
    if (showMoreSettingsButton) {
        QuickNavSettings(onNavigateToSettings = onNavigateToSettings)
    }
}

File: ui/uistateadapter/capture/src/main/java/com/google/jetpackcamera/ui/uistateadapter/capture/FlashModeUiStateAdapter.kt [hunk 11]
```diff
@@ -27,127 +28,169 @@ import com.google.jetpackcamera.ui.uistate.SingleSelectableUiState
...
+        // 2. Hide if not designated as visible by the developer.
+        // todo(kc): supply visible flash modes from developer options
+        /*if (!visibleFlashModes.contains(mode)) {
+            continue
+        }*/
+
+        // 3. Special handling for LOW_LIGHT_BOOST based on other settings.
+        if (mode == FlashMode.LOW_LIGHT_BOOST) {
+            if (cameraAppSettings.concurrentCameraMode == ConcurrentCameraMode.DUAL) {
+                continue // Hide LLB if Dual Camera is active
```
temcguir: Nit: same point on formatting as elsewhere – can we move all of these trailing comments (lines 101, 109, 121, etc) to their own dedicated lines above the relevant declarations?

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/quicksettings/ui/QuickSettingsComponents.kt
```diff
@@ -863,34 +691,75 @@ private fun QuickSettingToggleButtonPreview() {
...
     }
 }
+
+@Preview(name = "JCA Setting Row - Dark Mode", showBackground = true, backgroundColor = 0xFF000000)
+@Composable
+private fun PreviewSettingRowDark() {
+    MaterialTheme(colorScheme = darkColorScheme()) {
+        Surface(
+            modifier = Modifier.fillMaxWidth(),
+            color = Color.Black // Consistent with Camera UI
```
temcguir: Nit: please move the trailing comment to its own dedicated line above the code rather than inline.

temcguir (changes requested):

ui/controller/impl/src/main/java/com/google/jetpackcamera/ui/controller/impl/QuickSettingsControllerImpl.kt
```diff
@@ -39,14 +39,14 @@ import kotlinx.coroutines.launch
...
  * @param trackedCaptureUiState The state flow to update with quick settings information.
+ * @param externalCaptureMode The external capture mode active.
  * @param cameraSystem The camera system to control.
- * @param externalCaptureMode The current external capture mode.
  * @param coroutineContext The [CoroutineContext] for launching coroutines.
  */
 class QuickSettingsControllerImpl(
     private val trackedCaptureUiState: MutableStateFlow<TrackedCaptureUiState>,
-    private val cameraSystem: CameraSystem,
     private val externalCaptureMode: ExternalCaptureMode,
```
temcguir: It looks like the order of `externalCaptureMode` and `cameraSystem` just got swapped.

To clarify my previous comment: we shouldn't need `externalCaptureMode` passed into the Quick Settings controller anymore. Because the UI State Adapters react to the external capture mode (by hiding/disabling conflicting UI elements based on the initial app settings), the UI is already structurally prevented from sending conflicting interactions down to this controller. 

Repeating defensive guards for `externalCaptureMode` here inside the controller is redundant. We should totally remove `externalCaptureMode` from this file and `FakeQuickSettingsController`!

ui/controller/impl/src/main/java/com/google/jetpackcamera/ui/controller/impl/QuickSettingsControllerImpl.kt
```diff
@@ -59,12 +57,6 @@ class QuickSettingsControllerImpl(
         }
     }
 
-    override fun setFocusedSetting(focusedQuickSetting: FocusedQuickSetting) {
```
temcguir, line 62: Just re-flagging that we still need updates to `QuickSettingsControllerImplTest.kt`!

The control flows inside `QuickSettingsControllerImpl` (such as `setFocusedSetting`) were heavily modified in this PR. We need ensuring unit tests added to verify that state mutates properly against our Fake dependencies.

ui/uistateadapter/capture/src/main/java/com/google/jetpackcamera/ui/uistateadapter/capture/compound/QuickSettingsUiStateAdapter.kt
temcguir: Friendly ping on this. I still don't see a `QuickSettingsUiStateAdapterTest.kt` file in the diff. 

Because we've refactored how `QuickSettingsUiState` is mapped from the injected sub-states (aspectRatio, captureMode, flashMode, etc.), it's vital we add test coverage verifying that this adapter effectively outputs the correct `QuickSettingsUiState.Available` or `Unavailable` states.

ui/components/capture/src/main/res/values/strings.xml
```diff
@@ -120,9 +110,36 @@
...
```
temcguir: It looks like these strings are still using a capital "S" for "Settings". To strictly match the visual mocks, could we update these to use a lowercase "s" (i.e., `"Video settings"`, `"Photo settings"`, and `"Photo and Video settings"`)?

ui/uistateadapter/capture/src/main/java/com/google/jetpackcamera/ui/uistateadapter/capture/FlashModeUiStateAdapter.kt
```diff
@@ -27,69 +29,124 @@ import com.google.jetpackcamera.ui.uistate.SingleSelectableUiState
...
+                }
+                isLlbHdrConflict -> DisabledReason.LLB_DISABLED_BY_HDR
+                else -> throw IllegalStateException("Unexpected disabled state")
+            }
+            displayableModes.add(
+                SingleSelectableUiState.Disabled(
+                    value = mode,
+                    disabledReason = reason
+                )
+            )
```
temcguir: Returning to this thread: since we are unconditionally invoking `displayableModes.add()` with a `SingleSelectableUiState.Disabled` state down here, the UI will still render these unsupported flashes as permanently grayed-out buttons. 

When a mode is completely unsupported by the hardware on the current lens (e.g., `LLB_UNSUPPORTED_ON_LENS` or `FLASH_UNSUPPORTED_ON_LENS`), we should skip adding it to `displayableModes` entirely, perhaps with a `continue` early above. This ensures these unsupported hardware options are seamlessly hidden rather than cluttering the UI as disabled options.

*(Note: The new unit test added for `FLASH_UNSUPPORTED_ON_LENS` in `FlashModeUiStateAdapterTest.kt` around line 364 will also need to be updated to expect the mode to be missing, rather than asserting against a `DisabledReason`).*

temcguir (approved):

ui/uistateadapter/capture/src/main/java/com/google/jetpackcamera/ui/uistateadapter/capture/FlashModeUiStateAdapter.kt
Kimblebee: I think makes sense, so the settings would only be completely hidden if the current lens or entire device *doesn't* support it. 
We should reserve the disabled state for config conflicts that can be resolved within a specific lens/capture mode.

app/src/androidTest/java/com/google/jetpackcamera/utils/ComposeTestRuleExt.kt
```diff
@@ -437,84 +436,72 @@ inline fun <reified T> ComposeTestRule.checkComponentStateDescriptionState(
...
-                getResString(CaptureR.string.quick_settings_description_capture_mode_video_only) ->
-                    return@let CaptureMode.VIDEO_ONLY
-            }
-        }
-        throw (AssertionError("unable to determine capture mode from quick settings"))
+    if (aspectRatioRowExists) {
+        return@visitQuickSettings CaptureMode.IMAGE_ONLY
     }
+
+    return@visitQuickSettings CaptureMode.VIDEO_ONLY
```
temcguir: This fallback logic creates a risk for test false positives. By unconditionally returning `CaptureMode.VIDEO_ONLY` when the `STANDARD` and `IMAGE_ONLY` rows aren't found, we're assuming the underlaying state based entirely on the *absence* of UI elements.

If a UI bug occurs where the Quick Settings sheet just fails to load any setting rows whatsoever, tests checking that the capture mode correctly switched to `VIDEO_ONLY` will fall through to this line and falsely pass!

Instead of a silent fallback, we should actively verify a unique semantic state for `VIDEO_ONLY` (or verify that a setting exclusive to video mode exists) and throw an `AssertionError("Unable to determine capture mode from quick settings")` at the bottom of this function if no explicit modes can be verified, matching how `getCurrentFlashMode()` handles missing states.

app/src/androidTest/java/com/google/jetpackcamera/utils/ComposeTestRuleExt.kt
```diff
@@ -437,84 +436,72 @@ inline fun <reified T> ComposeTestRule.checkComponentStateDescriptionState(
...
+        false
     }
 }
 
-fun ComposeTestRule.getCurrentLensFacing(): LensFacing = visitQuickSettings {
-    onNodeWithTag(QUICK_SETTINGS_FLIP_CAMERA_BUTTON).fetchSemanticsNode(
-        "Flip camera button is not visible when expected."
+fun ComposeTestRule.getCurrentLensFacing(): LensFacing {
+    onNodeWithTag(
+        com.google.jetpackcamera.ui.components.capture.FLIP_CAMERA_BUTTON
```
temcguir: Per our `styleguide.md`, please avoid using fully qualified symbols inline. Please add `com.google.jetpackcamera.ui.components.capture.FLIP_CAMERA_BUTTON` to the file's `import` block instead.

app/src/androidTest/java/com/google/jetpackcamera/utils/ComposeTestRuleExt.kt
```diff
@@ -437,84 +436,72 @@ inline fun <reified T> ComposeTestRule.checkComponentStateDescriptionState(
...
-        throw AssertionError("Unable to determine flash mode from quick settings")
     }
+    throw AssertionError("Unable to determine flash mode from quick settings")
 }
 
 fun ComposeTestRule.getCurrentCaptureMode(): CaptureMode = visitQuickSettings {
-    waitUntil(timeoutMillis = DEFAULT_TIMEOUT_MILLIS) {
-        onNodeWithTag(BTN_QUICK_SETTINGS_FOCUS_CAPTURE_MODE).isDisplayed()
+    val standardRowExists = onAllNodesWithTag(
+        com.google.jetpackcamera.ui.components.capture.ROW_QUICK_SETTINGS_CAPTURE_MODE
```
temcguir: Per our `styleguide.md`, please avoid using fully qualified symbols inline. Please add `com.google.jetpackcamera.ui.components.capture.ROW_QUICK_SETTINGS_CAPTURE_MODE` to the file's `import` block instead.

app/src/androidTest/java/com/google/jetpackcamera/utils/ComposeTestRuleExt.kt
```diff
@@ -437,84 +436,72 @@ inline fun <reified T> ComposeTestRule.checkComponentStateDescriptionState(
...
-        for (description in node.config[SemanticsProperties.ContentDescription]) {
-            // check description is one of the capture modes
-            when (description) {
-                getResString(CaptureR.string.quick_settings_description_capture_mode_standard) ->
-                    return@let CaptureMode.STANDARD
 
-                getResString(CaptureR.string.quick_settings_description_capture_mode_image_only) ->
-                    return@let CaptureMode.IMAGE_ONLY
+    val aspectRatioRowExists = onAllNodesWithTag(
+        com.google.jetpackcamera.ui.components.capture.ROW_QUICK_SETTINGS_ASPECT_RATIO
```
temcguir: Per our `styleguide.md`, please avoid using fully qualified symbols inline. Please add `com.google.jetpackcamera.ui.components.capture.ROW_QUICK_SETTINGS_ASPECT_RATIO` to the file's `import` block instead.

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/CaptureScreenComponents.kt
```diff
@@ -905,9 +905,17 @@ fun FlipCameraButton(
             onClick = onClick,
             enabled = enabledCondition
         ) {
+            val contentDescription = when (flipLensUiState.selectedLensFacing) {
+                com.google.jetpackcamera.model.LensFacing.FRONT -> stringResource(
+                    R.string.quick_settings_front_camera_description
+                )
+                com.google.jetpackcamera.model.LensFacing.BACK -> stringResource(
```
temcguir, line 912: Per our `styleguide.md`, please avoid using fully qualified symbols inline. Please add `com.google.jetpackcamera.model.LensFacing` to the file's `import` block rather than fully qualifying `FRONT` and `BACK`.

ui/uistateadapter/capture/src/main/java/com/google/jetpackcamera/ui/uistateadapter/capture/HdrUiStateAdapter.kt
```diff
@@ -55,42 +55,42 @@ internal fun HdrUiState.Companion.from(
...
         cameraAppSettings
     )
-    val activeEffectTargets = cameraConstraints?.effectTargetsMap?.get(
+    val affectsImageCapture = cameraConstraints?.effectTargetsMap?.get(
         cameraAppSettings.selectedCameraEffect
-    ) ?: emptySet()
-    val affectsImageCapture = activeEffectTargets.contains(
-        com.google.jetpackcamera.model.CameraEffectTarget.IMAGE_CAPTURE
-    )
+    )?.contains(com.google.jetpackcamera.model.CameraEffectTarget.IMAGE_CAPTURE) == true
```
temcguir, line 60: Per our `styleguide.md`, please avoid using fully qualified symbols inline. Please add `com.google.jetpackcamera.model.CameraEffectTarget.IMAGE_CAPTURE` to the file's `import` block instead.

I see this was already present before, but would be nice to fix up since you're editing it.

temcguir (approved):

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/quicksettings/ui/QuickSettingsComponents.kt
temcguir, line 195: We should pivot away from using the native `ModalBottomSheet` here entirely so the UI library stays perfectly compatible with embedding apps.

The issue with M3's `ModalBottomSheet` is that it spawns a new OS-level window under the hood. If an external client app embeds our UI while already operating in a dialog or complex window lifecycle, it creates nested windows. That structure usually leads to severe downstream bugs like Z-ordering conflicts, obscured Snackbars, and broken back-press handling.

We should switch this to a `BottomSheetScaffold` since it renders within the same view hierarchy and completely avoids the window nesting bugs. To make it easy for consumers so they don't have to manually wire up scaffolds and scrims over their camera views, we can encapsulate it in a reusable wrapper component. Doing this also lets us hook our manual scrim directly into `LocalDisableAnimations` to keep our tests stable.

Here is a rough idea of how we can structure that in the UI library:

```kotlin
@Composable
fun QuickSettingsScaffold(
    quickSettingsUiState: QuickSettingsUiState,
    onDismiss: () -> Unit,
    content: @Composable () -> Unit 
) {
    val sheetState = rememberStandardBottomSheetState(
        initialValue = SheetValue.Hidden,
        skipHiddenState = false
    )
    val scaffoldState = rememberBottomSheetScaffoldState(bottomSheetState = sheetState)
    val disableAnimations = LocalDisableAnimations.current

    val isOpen = quickSettingsUiState is QuickSettingsUiState.Available && quickSettingsUiState.quickSettingsIsOpen
    LaunchedEffect(isOpen) {
        if (isOpen) sheetState.expand() else sheetState.hide()
    }

    BottomSheetScaffold(
        modifier = modifier,
        scaffoldState = scaffoldState,
        sheetPeekHeight = 0.dp, 
        sheetContent = { 
            // QuickSettingsContent(...)
        }
    ) { innerPadding ->
        Box(modifier = Modifier.padding(innerPadding).fillMaxSize()) {
            content()
            
            // Manual scrim with LocalDisableAnimations support
            AnimatedVisibility(
                visible = sheetState.targetValue != SheetValue.Hidden,
                enter = if (disableAnimations) EnterTransition.None else fadeIn(),
                exit = if (disableAnimations) ExitTransition.None else fadeOut()
            ) {
                Box(
                    modifier = Modifier
                        .fillMaxSize()
                        .background(Color.Black.copy(alpha = 0.4f))
                        .clickable(
                            interactionSource = remember { MutableInteractionSource() },
                            indication = null
                        ) {
                            onDismiss()
                        }
                )
            }
        }
    }
}
```

Then over in `PreviewScreen.kt` (and for any external callers), it keeps the usage really clean and avoids forcing them to build their own scaffolds:

```kotlin
QuickSettingsScaffold(
    quickSettingsUiState = quickSettingsState.value,
    onDismiss = { controller.toggleQuickSettings() }
) {
    LayoutWrapper(
        // Camera layout goes untouched inside here
    )
}
```

It probably also makes sense to add somethign to the `styleguide.md` so we don't accidentally re-add the `ModalBottomSheet` again in the future.

temcguir (approved):

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/quicksettings/ui/QuickSettingsComponents.kt
Kimblebee: addressed this in #569

PR #537, opened 2026-07-07
Refactor: Move Hilt modules from core to data layer

davidjiagoogle:
Refactor Hilt modules to enforce modularization boundaries:

- Created `:data:common` and moved `CommonModule` out of `:core:common`.
- Created `:data:camera:low-light` and moved `LowLightBoostModule` out of `:core:camera:low-light`.
- Created `:data:camera:postprocess` and moved `PostProcessModule` out of `:core:camera:postprocess`.
- Moved `:core:camera:low-light-playservices` to `:data:camera:low-light-playservices`.
- Removed unused Hilt/kapt plugins and dependencies from `:core:common`, `:core:settings`, `:core:camera:low-light`, and `:core:camera:postprocess`.
- Removed unused `hilt.navigation.compose` dependency from `:ui:components:capture` and explicitly declared `androidx.core:core-ktx`.
- Updated all consumer module dependencies and import statements across app, feature, data, and core modules.


core/camera/build.gradle.kts
```diff
@@ -142,6 +142,7 @@ dependencies {
     implementation(project(":core:settings"))
     implementation(project(":core:model"))
     implementation(project(":core:common"))
+    implementation(project(":data:common"))
```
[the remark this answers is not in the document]
davidjiagoogle: Moved the @qualifier annotations back to core:common and added javax.inject:javax.inject to circumvent this

core/camera/low-light-playservices-di/build.gradle.kts
temcguir, line 1: Providing application-wide `CoroutineDispatcher`s and root `CoroutineScope`s is inherently an app-level policy, rather than a discrete capability library. To centralize application architecture, let's delete this `data:common` module and simply move `CommonModule.kt` directly into our `app` module where global configuration belongs.

core/camera/low-light/low-light-di/build.gradle.kts
temcguir, line 1: Similar to the playservices implemention, we shouldn't create a `data` layer module just for Hilt wiring. Instead, rename this to `core:camera:low-light:low-light-di` to house `LowLightBoostModule.kt`. This isolates the Dagger `@Multibinds` logic in a dedicated DI wrapper, leaving the base `low-light` module pure.

core/camera/low-light-playservices/build.gradle.kts
temcguir, line 1: Let's revert the move of `low-light-playservices` to the `data` layer; it provides a core capability implementation and should remain in our `core` libraries as `core:camera:low-light-playservices`. 

To ensure this library is completely DI-agnostic, let's strip the `dagger.hilt.android` plugin from it entirely. Let's create a dedicated `core:camera:low-light-playservices-di` module to house `PlayServicesLowLightBoostModule.kt`. This moves everything into core so others aren't depending on our data layer, but still allows for other apps to choose whether to use our Hilt implementation or not.

core/camera/postprocess/postprocess-di/build.gradle.kts
temcguir, line 1: As with the low-light modules, let's move this out of the `data` layer and rename it to `core:camera:postprocess:postprocess-di`. This gives us a clean DI wrapper for the `PostProcessModule.kt` bindings.

core/common/build.gradle.kts
```diff
@@ -69,6 +67,7 @@ dependencies {
     implementation(libs.androidx.appcompat)
     implementation(libs.kotlinx.atomicfu)
     implementation(libs.androidx.tracing)
+    implementation("javax.inject:javax.inject:1")
```
temcguir: Let's remove this dependency entirely. To be truly pure and unopinionated, our shared `core` libraries shouldn't assume the consumer uses *any* dependency injection (even JSR-330). We should delete the `@Inject` constructors in the core classes, and instead construct them manually inside our `@Provides` blocks within the `-di` and `app` modules.

app/src/main/java/com/google/jetpackcamera/di/AppQualifiers.kt
temcguir, line 1: Let's remove this file from the `core` library. Since our `app` module centrally handles DI via Hilt, the responsibility for Dagger `@Qualifier` annotations belongs in the `app` module. 

We should move these qualifiers (like `@IODispatcher`) into the `app` module (e.g., `app/src/main/java/com/google/jetpackcamera/app/di/AppQualifiers.kt`). The app layer can continue using them to grab the correct dispatchers from its Hilt graph and pass them as standard arguments when building core classes.

temcguir, line 1: nit: Assuming we move this file to the `app` module, we should update the KDocs. The newly added KDocs here (e.g. `Qualifier for the default file path generator.`) are redundant and merely restate the annotation class name. The styleguide dictates that documentation should provide more value than the signature alone (e.g. explaining *why* we need a specific qualifier).

temcguir, line 1: nit: Section 2 of our styleguide enforces using the most restrictive visibility modifier possible (`internal`, `private`). Since these DI modules and qualifiers are strictly internal wiring and shouldn't be exposed as a public API surface, they should be declared as `internal class` and `internal annotation class`.

temcguir (changes requested):

core/common/build.gradle.kts
temcguir: The `@Inject` imports were correctly removed from the classes, however `implementation("javax.inject:javax.inject:1")` is still listed. Please remove.

app/src/main/java/com/google/jetpackcamera/di/AppQualifiers.kt
temcguir, line 1: These new qualifiers are still exposed as public API surfaces. Please update them to `internal annotation class` as per the styleguide.

temcguir (approved):
Two minor pieces of feedback from the last check were missed. Could you address these before we merge?

PR #538, opened 2026-07-08
fix debug screen zoom control bug

Kimblebee:
fix debug screen zoom control bug where debug zoom controls were missing from accessibility tree, which led to automated test failures on some devices.

### DIFF
--- a/app/src/androidTest/java/com/google/jetpackcamera/ExternalAutomationCompatibilityTest.kt
+++ b/app/src/androidTest/java/com/google/jetpackcamera/ExternalAutomationCompatibilityTest.kt
@@ -21,6 +21,7 @@ import androidx.test.rule.GrantPermissionRule
 import androidx.test.uiautomator.By
 import androidx.test.uiautomator.UiDevice
 import androidx.test.uiautomator.Until
+import com.google.common.truth.Truth.assertWithMessage
 import com.google.jetpackcamera.ui.components.capture.CAPTURE_BUTTON
 import com.google.jetpackcamera.ui.components.capture.FLIP_CAMERA_BUTTON
 import com.google.jetpackcamera.ui.debug.DEBUG_OVERLAY_BUTTON
@@ -30,7 +31,6 @@ import com.google.jetpackcamera.ui.debug.DEBUG_OVERLAY_SET_ZOOM_RATIO_TEXT_FIELD
 import com.google.jetpackcamera.utils.TEST_REQUIRED_PERMISSIONS
 import com.google.jetpackcamera.utils.debugExtra
 import com.google.jetpackcamera.utils.runMainActivityScenarioTest
-import org.junit.Assert.assertNotNull
 import org.junit.Before
 import org.junit.Rule
 import org.junit.Test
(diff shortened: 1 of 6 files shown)


app/src/androidTest/java/com/google/jetpackcamera/ExternalAutomationCompatibilityTest.kt
```diff
@@ -0,0 +1,123 @@
...
+            // Verify the confirm button is visible to UI Automator
+            val confirmButton = device.findObject(By.res(DEBUG_OVERLAY_SET_ZOOM_RATIO_SET_BUTTON))
+            assertWithMessage(
+                "Confirm button not found by UI Automator"
+            ).that(confirmButton).isNotNull()
+        }
+    }
+
+    companion object {
+        private const val TIMEOUT_MS = 5000L
```
temcguir: Since we are waiting for the Capture and Flip buttons right after `runMainActivityScenarioTest` launches the app, we are subject to the same cold-start emulator delays (like virtual camera HAL initialization) that the rest of the app tests face.

Instead of defining a custom 5000ms timeout here, could we import and use `APP_START_TIMEOUT_MILLIS` (which is 20,000ms) from `UiTestUtil.kt` for the initial button checks? We can also use `DEFAULT_TIMEOUT_MILLIS` for the subsequent dialog checks. Using the centralized timeouts helps keep our CI flakiness mitigations consistent across the codebase!

app/src/androidTest/java/com/google/jetpackcamera/ExternalAutomationCompatibilityTest.kt
```diff
@@ -0,0 +1,123 @@
...
+    @get:Rule
+    val permissionsRule: GrantPermissionRule =
+        GrantPermissionRule.grant(*(TEST_REQUIRED_PERMISSIONS).toTypedArray())
+
+    private lateinit var device: UiDevice
+
+    @Before
+    fun setUp() {
+        device = UiDevice.getInstance(InstrumentationRegistry.getInstrumentation())
+    }
```
temcguir, line 54: Let's add an `@After` block to clean up the device state. Since the `setZoomRatio` test opens a dialog and interacts with a text field, we should ensure the screen is cleared for the next test.

```kotlin
@After
fun tearDown() {
    device.pressBack()
    device.pressHome()
}

File: app/src/androidTest/java/com/google/jetpackcamera/ExternalAutomationCompatibilityTest.kt [hunk 3]
```diff
@@ -0,0 +1,123 @@
...
+
+    @Test
+    fun verifyUiAutomator_canFindCaptureControls() {
+        runMainActivityScenarioTest {
+            // Wait for the capture button to be visible to UI Automator via Resource ID
+            val captureButton = device.wait(Until.findObject(By.res(CAPTURE_BUTTON)), TIMEOUT_MS)
```
temcguir: To ensure maximum compatibility across different Android OS versions in ATP/TAP, it's safer to provide the fully-qualified package format for UIAutomator's `By.res()`. Stricter OS versions will sometimes fail to match bare resource strings.
```kotlin
val pkg = InstrumentationRegistry.getInstrumentation().targetContext.packageName
val captureButton = device.wait(Until.findObject(By.res(pkg, CAPTURE_BUTTON)), TIMEOUT_MS)
```

temcguir (approved):

PR #539, opened 2026-07-13
modernize elapsed time composable

Kimblebee:
Screenshot 2026-07-13 at 1 27 17 PM


ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/CaptureScreenComponents.kt
```diff
@@ -141,15 +151,28 @@ fun ElapsedTimeText(
...
-            modifier = modifier,
-            text = state.elapsedTimeNanos.nanoseconds
-                .toComponents { minutes, seconds, _ -> "%02d:%02d".format(minutes, seconds) },
-            textAlign = TextAlign.Center,
-            style = LocalTextStyle.current.copy(
-                fontFeatureSettings = "tnum"
+        val formattedTime = remember(state.elapsedTimeNanos) {
+            state.elapsedTimeNanos.nanoseconds
+                .toComponents { minutes, seconds, _ -> "%d:%02d".format(minutes, seconds) }
+        }
```
temcguir, line 167: According to the mocks for the paused state, the badge should read `PAUSED 0:02` when video recording is paused, rather than just the time. 

To conditionally prepend the `"PAUSED "` string here, we'll need to know if the recording is paused. Since `ElapsedTimeUiState.Enabled` doesn't currently expose this, we should:
1. Add an `isPaused: Boolean` property to `ElapsedTimeUiState.Enabled` (or create a new `ElapsedTimeUiState.Paused` variant).
2. Update `ElapsedTimeUiState.Companion.from(cameraState:)` in `ElapsedTimeUiStateAdapter.kt` to map `VideoRecordingState.Active.Paused` so the paused state isn't lost. 

Once that's added, we can prepend the "PAUSED" text here conditionally based on that state.

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/CaptureScreenComponents.kt
```diff
@@ -950,3 +973,51 @@ private fun FocusMeteringIndicator(
         }
     }
 }
+
+@Preview(name = "Elapsed Time", showBackground = true, backgroundColor = 0xFF000000)
```
temcguir, line 1011: Once we add the paused state to `ElapsedTimeUiState` to support the `"PAUSED"` text prefix, let's also add a preview scenario for it here (e.g. `PAUSED 0:30`) to ensure it matches the mocks.

temcguir (changes requested):

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/CaptureScreenComponents.kt
```diff
@@ -141,15 +151,28 @@ fun ElapsedTimeText(
...
-            text = state.elapsedTimeNanos.nanoseconds
-                .toComponents { minutes, seconds, _ -> "%02d:%02d".format(minutes, seconds) },
-            textAlign = TextAlign.Center,
-            style = LocalTextStyle.current.copy(
-                fontFeatureSettings = "tnum"
+        val formattedTime = remember(state.elapsedTimeNanos) {
+            state.elapsedTimeNanos.nanoseconds
+                .toComponents { minutes, seconds, _ -> "%d:%02d".format(minutes, seconds) }
+        }
+        Box(
```
temcguir, line 174: Should we include a UI test file for this composable (e.g., `ElapsedTimeTextTest.kt`) to run basic accessibility checks, similar to the baseline established for the capture button in PR #461? 

Running `composeTestRule.onRoot().tryPerformAccessibilityChecks()` would be great to lock in the contrast ratio of the white text against the new `#ED0000` background to ensure it always passes ATF guidelines. 

Additionally, we should improve the screen reader experience. Right now, TalkBack will read the literal text formatting (like "0 colon 30"). We could add a `clearAndSetSemantics { contentDescription = ... }` modifier here to override that and provide a more meaningful announcement like "Recording time: 0 minutes and 30 seconds" (or "Recording paused..."). We could then add a quick assertion in the test using `assertContentDescriptionEquals()`. What do you think?

Kimblebee: sounds like a good idea!

temcguir (approved):

PR #541, opened 2026-07-21
Fix crash on rapid double-clicking video record button

davidjiagoogle:
## Summary
Fixes an issue (b/535645067) where double-clicking or rapidly tapping the video recording button in quick succession crashed the app with `IllegalStateException: A recording is already in progress`.

## Changes Made
1. **Added `CaptureButtonUiState.Enabled.Recording.Pending`**: Created a new UI state representing the initialization window when video recording is starting.
2. **Mapped `VideoRecordingState.Starting`**: Mapped `VideoRecordingState.Starting` to `CaptureButtonUiState.Enabled.Recording.Pending` instead of `Idle` in `CaptureButtonUiStateAdapter.kt`.
3. **Updated `CaptureButtonComponents.kt`**: Added `Recording.Pending` to the ignored/no-op states during click handling (`onKeyUp`), preventing rapid consecutive clicks from sending duplicate `StartRecordingEvent`s.
4. **Defensive Error Handling in `CameraSession.kt`**: Replaced throwing an unhandled `IllegalStateException` with logging a warning when receiving a duplicate `StartRecordingEvent` during active recording.
5. **Unit Tests**: Updated `CaptureButtonUiStateAdapterTest.kt` to assert the new `Pending` state.


ui/uistate/capture/src/main/java/com/google/jetpackcamera/ui/uistate/capture/CaptureButtonUiState.kt
```diff
@@ -61,6 +61,11 @@ sealed interface CaptureButtonUiState {
         sealed interface Recording : Enabled {
             override val isEnabled: Boolean get() = true
 
+            /**
+             * The video recording request has been sent and is pending initialization.
+             */
+            data object Pending : Recording
```
temcguir: Given we are mapping this directly from `VideoRecordingState.Starting`, could we rename `Pending` to `Starting` here (and throughout the PR)? It feels like it captures the actual state of the camera backend much more accurately than "Pending". 

```kotlin
            /**
             * The video recording request has been sent and is starting.
             */
            data object Starting : Recording
```

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/CaptureButtonComponents.kt
```diff
@@ -272,6 +272,7 @@ internal fun CaptureButton(
                     }
 
                     CaptureButtonUiState.Enabled.Recording.LockedRecording -> onStopRecording()
+                    CaptureButtonUiState.Enabled.Recording.Pending,
```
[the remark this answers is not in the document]
temcguir: I agree with @gemini-code-assist's inline comment to allow stopping the recording by releasing the long-press while in the `Starting` state. 

However, applying the bot's fix will expose a coroutine race condition in `CaptureControllerImpl.kt`. Because `CameraSession.kt` silently ignores `StopRecordingEvent`s if it isn't actively recording yet, a rapid start/stop sequence will freeze the camera into an endless recording loop. If the UI dispatcher executes the stop coroutine before the start coroutine sends its event, the `Stop` event is sent to the channel first (and ignored), and then the `Start` event triggers the recording infinitely.

Could we also include a quick fix in `CaptureControllerImpl.kt` in this PR to synchronously cancel the job?
```kotlin
    // CaptureControllerImpl.kt
    override fun stopVideoRecording() {
        Log.d(TAG, "stopVideoRecording")
        // Cancel job synchronously before sending Stop to prevent the race condition
        recordingJob?.cancel() 
        recordingJob = null
        scope.launch {
            cameraSystem.stopVideoRecording()
        }
    }
```

temcguir (changes requested):
It would be awesome to add a quick unit test to `PreviewViewModelTest.kt` to ensure this specific fast-release cancellation doesn't regress. 

First, add a diagnostic counter to `FakeCameraSystem.kt`:
```
    var numVideoRecordingStarts = 0
    
        saveLocation: SaveLocation,
        onVideoRecord: (OnVideoRecordEvent) -> Unit
    ) {
        if (!useCasesBinded) {
            throw IllegalStateException("Usecases not bound")
        }
        numVideoRecordingStarts++
        recordingInProgress = true
    }
```
Then we can add the following test directly inside `PreviewViewModelTest.kt` to verify that rapidly firing "stop" immediately cancels the "start" job:
```
    @Test
    fun fastStartAndStopVideoRecording_cancelsRecording() = runTest(StandardTestDispatcher()) {
        startCameraUntilRunning()
        
        // Start and stop immediately without advancing the dispatcher
        previewViewModel.captureController.startVideoRecording()
        previewViewModel.captureController.stopVideoRecording()
        
        // Let the coroutines execute
        advanceUntilIdle()
        
        // Verify that because we cancelled the job synchronously, the start implementation
        // was bypassed completely to protect us from the channel race condition!
        assertThat(cameraSystem.numVideoRecordingStarts).isEqualTo(0)
    }
```

ui/uistateadapter/capture/src/test/java/com/google/jetpackcamera/ui/uistateadapter/capture/CaptureButtonUiStateAdapterTest.kt
```diff
@@ -95,7 +95,7 @@ class CaptureButtonUiStateAdapterTest {
     }
 
     @Test
-    fun from_cameraRunning_recordingStarting_returnsIdleAndEnabled() {
+    fun from_cameraRunning_recordingStarting_returnsPendingRecording() {
```
temcguir: nit: Should this now be called `from_cameraRunning_recordingStarting_returnsStartingRecording`?

temcguir (approved):

PR #548, opened 2026-07-31
Replace Context Receivers with Context Parameters

Kimblebee:
TLDR: Upgrades kotlin version to 2.4 with the following implications:
* deprecates Context Receivers in favor of Context Parameters
* upgrades Hilt version to 2.60.1 to resolve compatibility issues
* removes all experimental context receiver flags (kotlin 2.2 required experimental flags for context parameters)

ktlint is also upgraded to 1.8.0 😋 

#### The changes to `CameraSession` / `ConcurrentCameraSession`
* During this dependency update, previous experimental `Context Receivers` used in these files ~ _surprise surprise_ ~ were deprecated in favor of the new ✨ **Context Parameters** ✨ as of Kotlin `2.2.0`! (_Read official docs_ [_**here**_](https://kotlinlang.org/docs/context-parameters.html))!  

#### Alternative Considered:
I considered reverting to the Extension Functions pattern for the following reasons:
  * Current standard
  * For 11 out of the 12 refactored functions, they only required one receiver (`CameraSessionContext`). For these, standard Kotlin extension functions (`CameraSessionContext.foo()`) work perfectly fine, and are 100% stable since Kotlin 1.0.
      * caveat to this is that the one (`Preview.Builder.updateCameraStateWithCaptureResults`) would need one of the receivers to be explicitly passed as a regular parameter. 
 
But TBH since Context Parameters is stable in `2.4.0`, i think it is OK to keep this PR as is! 🆗 

butterfly hatching from chrysalis


core/camera/src/main/java/com/google/jetpackcamera/core/camera/CameraSession.kt
```diff
@@ -119,253 +119,260 @@ private val QUALITY_RANGE_MAP = mapOf(
     SD to Range.create(241, 719)
 )
 
-context(CameraSessionContext)
 @ExperimentalCamera2Interop
+context(c: CameraSessionContext)
```
davidjiagoogle, line 123: Could you find out how the new context parameter should be handled on other platforms we mirror our code on?

PR #567, opened 2026-08-19
Make MediaRepository and ImageWellController optional in CaptureControllerImpl

davidjiagoogle:
Allow CaptureControllerImpl to be used in client applications  that manage their own media persistence and do not display an image well thumbnail by making mediaRepository and imageWellController optional parameters defaulting to null.

### DIFF
--- a/ui/controller/impl/src/main/java/com/google/jetpackcamera/ui/controller/impl/CaptureControllerImpl.kt
+++ b/ui/controller/impl/src/main/java/com/google/jetpackcamera/ui/controller/impl/CaptureControllerImpl.kt
@@ -66,12 +66,12 @@ private const val IMAGE_CAPTURE_TRACE = "JCA Image Capture"
 class CaptureControllerImpl(
     private val trackedCaptureUiState: MutableStateFlow<TrackedCaptureUiState>,
     private val cameraSystem: CameraSystem,
-    private val mediaRepository: MediaRepository,
+    private val mediaRepository: MediaRepository? = null,
     private val saveMode: SaveMode,
     private val externalCaptureMode: ExternalCaptureMode,
     private val externalCapturesCallback: () -> Pair<SaveLocation, IntProgress?>,
     override val captureEvents: Channel<CaptureEvent>,
-    private val imageWellController: ImageWellController,
+    private val imageWellController: ImageWellController? = null,
     coroutineContext: CoroutineContext
 ) : CaptureController {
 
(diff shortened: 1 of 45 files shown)


ui/controller/impl/src/main/java/com/google/jetpackcamera/ui/controller/impl/CaptureControllerImpl.kt
temcguir: We should be removing these from dependencies.

temcguir: This comment needs to be updated. snackBarController isn't here anymore.

ui/controller/impl/src/main/java/com/google/jetpackcamera/ui/controller/impl/CaptureControllerImpl.kt
```diff
@@ -66,12 +66,12 @@ private const val IMAGE_CAPTURE_TRACE = "JCA Image Capture"
 class CaptureControllerImpl(
     private val trackedCaptureUiState: MutableStateFlow<TrackedCaptureUiState>,
     private val cameraSystem: CameraSystem,
-    private val mediaRepository: MediaRepository,
+    private val mediaRepository: MediaRepository? = null,
```
temcguir: Instead of making MediaRepository nullable, we should just replace this with a lambda action (or set of lambda actions: onImageSuccess/onVideoSuccess? You would need to look at where it's used) so that we can remove the dependency completely on the data/media module. Then when constructing the controller, the user can pass in an action from their ViewModel. This keeps us from leaking data layer constructs into core UI libraries. @gemini-code-assist can you suggest how to do this?

ui/controller/impl/src/main/java/com/google/jetpackcamera/ui/controller/impl/CaptureControllerImpl.kt
```diff
@@ -50,28 +48,29 @@ private const val TAG = "CaptureButtonControllerImpl"
...
     private val trackedCaptureUiState: MutableStateFlow<TrackedCaptureUiState>,
     private val cameraSystem: CameraSystem,
-    private val mediaRepository: MediaRepository,
     private val saveMode: SaveMode,
     private val externalCaptureMode: ExternalCaptureMode,
     private val externalCapturesCallback: () -> Pair<SaveLocation, IntProgress?>,
     override val captureEvents: Channel<CaptureEvent>,
-    private val imageWellController: ImageWellController,
+    private val imageWellController: ImageWellController? = null,
+    private val onImageCached: ((Uri) -> Unit)? = null,
```
temcguir: nit: is there a reason we called this "Cached"? Since this is the "CaptureController", maybe "onImageCaptureComplete" and "onVideoCaptureComplete" might be more appropriate? @gemini-code-assist what do you think? Can you suggest some other options if these don't sound good?

temcguir (approved):

PR #569, opened 2026-08-27
bottom sheet refactor

Kimblebee:
## Summary
Refactors Quick Settings to use Material 3's in-hierarchy `BottomSheetScaffold` instead of `ModalBottomSheet`. 

### Why the Transition from `ModalBottomSheet` to `BottomSheetScaffold`?
- **In-Hierarchy Composition:** `BottomSheetScaffold` renders the sheet within the primary Compose layout tree, maintaining a single window lifecycle and preserving seamless viewfinder rendering.

### Key Changes
- **In-Hierarchy Scaffolding:** Replaced root `Scaffold` with `BottomSheetScaffold` in `CaptureLayout.kt` (`sheetPeekHeight = 0.dp`), unifying `SnackbarHost` and removing nested sheet wrappers.
- **Clickable Drag Handle:** Added `BottomSheetDefaults.DragHandle` with `onDismissQuickSettings` click action and `QUICK_SETTINGS_DRAG_HANDLE` test tag.
- **State & Back Handling:** Managed `isQuickSettingsOpen` in `PreviewScreen.kt` with `LaunchedEffect` sync and `BackHandler(enabled = isQuickSettingsOpen)`.
- **Test Determinism:** Added `closeQuickSettings()` and updated `visitQuickSettings` in `ComposeTestRuleExt.kt` to click the drag handle pill, eliminating flaky synthetic gesture timeouts.

### Lifecycle Behavior & Trade-offs
- **Persistent Composition (Never Decomposed):** Unlike `ModalBottomSheet` (which dynamically mounts/unmounts from composition upon opening/closing), `BottomSheetScaffold` is a persistent layout container. When hidden (`sheetPeekHeight = 0.dp`), the sheet container **remains in the composition tree** and is translated off-screen rather than decomposed. Automated tests therefore check visibility (`.isNotDisplayed()`) rather than complete semantics node absence.
- **State Synchronization:** Traded the declarative `onDismissRequest` modal API for bidirectional `LaunchedEffect` synchronization between the boolean UI state (`isQuickSettingsOpen`) and `SheetState`.
- **Drag Handle Affordance:** Material 3's default drag handle is purely a drag gesture affordance; a `.clickable` modifier was explicitly attached to provide a deterministic tap-to-dismiss target for accessibility and test automation.


feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewScreen.kt
```diff
@@ -300,6 +305,39 @@ private fun ContentScreen(
         )
     }
 
+    var isQuickSettingsOpen by rememberSaveable { mutableStateOf(false) }
```
temcguir: Right now, `isQuickSettingsOpen`, the `LaunchedEffect` synchronization, and `BackHandler` are all defined here in `PreviewScreen.kt`.

Because `feature:preview` is an internal app module and not part of the shared UI library, any external consumer building their own camera screen using `ui:components:capture` would have to copy and wire up this entire block of `BackHandler`, `rememberSaveable`, and sheet synchronization in their own codebase.

Could we push the `BackHandler` and this state synchronization down into `PreviewLayout` (in `ui:components:capture`)? We could still allow callers to optionally pass in a state or dismiss callback if they need custom control, but having `PreviewLayout` handle its own `BackHandler` and scaffold state by default would make our UI library much more plug-and-play for embedding apps.

app/src/androidTest/java/com/google/jetpackcamera/utils/ComposeTestRuleExt.kt
```diff
@@ -194,7 +193,16 @@ fun ComposeTestRule.waitForNodeWithTagToDisappear(
     timeoutMillis: Long = DEFAULT_TIMEOUT_MILLIS
 ) {
     waitUntil(timeoutMillis = timeoutMillis) {
-        onNodeWithTag(tag).isNotDisplayed()
+        val nodes = onAllNodesWithTag(tag).fetchSemanticsNodes()
+        if (nodes.isEmpty()) {
+            true
+        } else {
+            try {
+                onNodeWithTag(tag).isNotDisplayed()
```
temcguir: In `waitForNodeWithTagToDisappear`, `onNodeWithTag(tag).isNotDisplayed()` already returns a `Boolean` rather than throwing when the node is visible. Catching `AssertionError` and returning `true` can accidentally mask other failures (for instance, if multiple nodes match the tag, `onNodeWithTag` throws an `AssertionError`). 

We should be able to simplify this to `nodes.isEmpty() || onNodeWithTag(tag).isNotDisplayed()`.

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/CaptureLayout.kt
```diff
@@ -115,13 +150,12 @@ fun PreviewLayout(
...
                             quickSettingsToggleButton = quickSettingsButton,
                             captureModeToggleSwitch = captureModeToggle,
-                            bottomSheetQuickSettings = quickSettingsOverlay,
                             zoomControls = zoomLevelDisplay,
                             elapsedTimeDisplay = elapsedTimeDisplay
                         )
                     }
                     // controls overlay
-                    snackBar(Modifier, snackbarHostState)
+                    snackBar(Modifier, scaffoldState.snackbarHostState)
```
temcguir, line 164: Because `BottomSheetScaffold` doesn't provide a scrim by default, the background (viewfinder and controls) stays fully interactive when Quick Settings is open. Tapping in the area above the sheet will pass straight through to the viewfinder and trigger tap-to-focus rather than closing the sheet.

Could we add a transparent `Box` overlay in this container when `targetValue == SheetValue.Expanded` that intercepts touches and calls `onDismissQuickSettings`? (Note: we specifically want `== SheetValue.Expanded` rather than `!= SheetValue.Hidden` because swiping down settles at `PartiallyExpanded`, and we don't want the overlay blocking touches once closed).

```kotlin
if (scaffoldState.bottomSheetState.targetValue == SheetValue.Expanded) {
    Box(
        modifier = Modifier
            .fillMaxSize()
            .clickable(
                interactionSource = remember { MutableInteractionSource() },
                indication = null,
                onClick = onDismissQuickSettings
            )
    )
}
```

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/quicksettings/QuickSettingsScreen.kt
```diff
@@ -50,25 +52,54 @@ import com.google.jetpackcamera.ui.uistate.capture.FlipLensUiState
...
+        }
+    }
+}
+
 /**
  * The UI bottom sheet component for quick settings.
  */
 @OptIn(ExperimentalMaterial3Api::class)
 @Composable
 fun QuickSettingsBottomSheet(
```
temcguir: Since `PreviewScreen` now uses `QuickSettingsScaffoldContent` with `BottomSheetScaffold`, this `QuickSettingsBottomSheet` composable (and `QuickSettingsModalBottomSheet` in `QuickSettingsComponents.kt`) is no longer used anywhere in the app except in the preview at the bottom of this file.

Could we remove `QuickSettingsBottomSheet` and `QuickSettingsModalBottomSheet` entirely, and update `ExpandedQuickSettingsUiPreview` to preview `QuickSettingsContent` instead?

feature/preview/src/main/java/com/google/jetpackcamera/feature/preview/PreviewScreen.kt
```diff
@@ -300,6 +305,39 @@ private fun ContentScreen(
...
+        }
+    }
+
+    // Intercept back navigation only while Quick Settings is actively open.
+    // Note: Checking `isQuickSettingsOpen` instead of `bottomSheetState.isVisible` ensures back handling
+    // is immediately relinquished back to the Activity the instant the sheet begins closing.
+    BackHandler(enabled = isQuickSettingsOpen) {
+        isQuickSettingsOpen = false
+    }
+
```
temcguir: If you swipe the bottom sheet away, it currently takes 2 clicks on the quick settings button to reopen it. This happens because swiping down with `sheetPeekHeight = 0.dp` snaps the sheet to `SheetValue.PartiallyExpanded` (at 0px height) rather than `SheetValue.Hidden`. That means `bottomSheetState.isVisible` stays `true` (and the bot's suggestion of checking `targetValue == SheetValue.Hidden` will also stay `false`), leaving `isQuickSettingsOpen` stuck at `true`.

Since `rememberStandardBottomSheetState` already uses `rememberSaveable` internally, we can eliminate the dual-state sync and both `LaunchedEffect` blocks by using `bottomSheetState` as our single source of truth:

```kotlin
val scaffoldState = rememberBottomSheetScaffoldState(
    bottomSheetState = rememberStandardBottomSheetState(
        initialValue = SheetValue.Hidden,
        skipHiddenState = false
    )
)
val isQuickSettingsOpen = scaffoldState.bottomSheetState.targetValue == SheetValue.Expanded

BackHandler(enabled = isQuickSettingsOpen) {
    scope.launch { scaffoldState.bottomSheetState.hide() }
}
```

Then in `captureButtonLambda`, `quickSettingsButtonLambda`, and `onDismissQuickSettings`, we can just call `scope.launch { scaffoldState.bottomSheetState.hide() }` (or `.expand()`) directly.

temcguir (changes requested):

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/CaptureLayout.kt
```diff
@@ -82,15 +101,31 @@ fun PreviewLayout(
...
+            scaffoldState = scaffoldState,
+            sheetPeekHeight = 0.dp,
+            sheetDragHandle = {
+                BottomSheetDefaults.DragHandle(
+                    modifier = Modifier
+                        .testTag(QUICK_SETTINGS_DRAG_HANDLE)
+                        .clickable(
+                            onClickLabel = "Close quick settings",
+                            onClick = onDismissQuickSettings
+                        )
```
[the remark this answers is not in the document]
Kimblebee: updated to `"Close expanded settings button"` resource value

temcguir (approved):

PR #570, opened 2026-09-01
Implement AF scene change preemption and focus reticle enhancements

temcguir:
# Implement AF scene change preemption and focus reticle enhancements

## Overview
This PR updates the tap-to-focus pipeline in Jetpack Camera App with scene change detection, visual updates to the focus reticle, and accessibility support.

### 1. Focus Lock Preemption (`CONTROL_AF_SCENE_CHANGE`)
When tap-to-focus engages (yielding an implicit focus lock), JCA now monitors the camera stream for the `CONTROL_AF_SCENE_CHANGE_DETECTED` flag. If the scene changes for 3 consecutive frames (after a 3000ms delay post-lock), the lock is automatically released and continuous auto-focus resumes. This prevents the camera from getting stuck out-of-focus if a user pans away after tapping.

**Unsupported Hardware Fallback:** If the optical hardware does not report `CONTROL_AF_SCENE_CHANGE` capabilities in the frame metadata, the pipeline runs an active coroutine monitor which securely cancels the focus lock after a fixed 15-second timeout, preventing permanent lock behavior on heavily fragmented hardware.

### 2. Focus Reticle UI Updates
- The focus indicator now persists on screen while focus is locked, holding at a slightly scaled-down geometry.
- During the scanning phase, the reticle smoothly pulses.
- The indicator uses an expanding-dissolve exit animation upon cancellation or failure.

### 3. Accessibility (A11y)
Added explicit TalkBack semantics to the focusing pipeline. The `CameraXViewfinder` is bound as a semantic node so synthetic double-taps correctly target the geometric center. Focus state transitions (Focusing, Focused, Focus failed, Focus released) use explicit TalkBack announcements via `view.announceForAccessibility(...)`. We intentionally use imperative announcements rather than an in-tree `liveRegion` because the focus indicator composable is dismissed upon cancellation or clearing, which would unmount an in-tree `liveRegion` before TalkBack can announce "Focus released".

### 4. Accessibility Testing Framework (ATF)
Integrates the Accessibility Testing Framework (ATF) alongside the existing UI testing suite. The PR enables accessibility checks at the test-rule level and adds tests covering the updated focus reticle `contentDescription`.

## Testing
- **Unit Tests:** Added `FocusMeteringTest.kt` to verify the 3000ms delay and frame debouncing logic.
- **Instrumented Tests:** Extended `CaptureScreenComponentsTest.kt` to assert TalkBack `contentDescription` properties across different focus states.
- **On-Device:** Tested TalkBack accessibility node geometry and 3-second hardware-cancellation on local physical devices.


core/camera/src/main/java/com/google/jetpackcamera/core/camera/FocusMetering.kt
```diff
@@ -115,6 +127,28 @@ internal suspend fun CameraSessionContext.processFocusMeteringEvents(
...
+
+                        try {
+                            cameraControl.cancelFocusAndMetering().await()
+                        } catch (_: CameraControl.OperationCanceledException) {
+                            // Ignored if cancelled by a subsequent action
+                        } catch (e: Exception) {
+                            Log.w(TAG, "cancelFocusAndMetering failed", e)
+                        }
+                        updateFocusState(FocusState.Status.CANCELLED)
+                    }
```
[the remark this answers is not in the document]
temcguir: Thanks for the catch! I've updated the logic to address the edge case for older Android versions (API < 28) and instances where `captureResults` flow is null. I added an explicit `if-else` wrapper that degrades to a 15-second fallback delay (`delay(15000L)`) before cancelling the focus and metering action. This guarantees the lock releases cleanly avoiding indefinite suspensions.

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/CaptureScreenComponents.kt
```diff
@@ -931,95 +938,163 @@ fun FlipCameraButton(
...
+                    }
                     .border(
-                        1.dp,
+                        2.dp,
                         Color.White,
                         CircleShape
                     )
-                    .size(TAP_TO_FOCUS_INDICATOR_SIZE)
+                    .size(TAP_TO_FOCUS_INDICATOR_SIZE * animatedScale)
             )
```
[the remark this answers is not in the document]
temcguir: Done. I've updated the focus indicator `Box` modifier to include a `.semantics` block. Now, whenever the `focusStatusDescription` is not empty, it'll apply that directly to the `contentDescription` for TalkBack and other accessibility services to parse.

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/CaptureScreenComponents.kt
```diff
@@ -931,95 +938,168 @@ fun FlipCameraButton(
...
+                view.announceForAccessibility(context.getString(R.string.focus_scanning))
+            FocusMeteringUiState.Status.SUCCESS ->
+                view.announceForAccessibility(context.getString(R.string.focus_locked))
+            FocusMeteringUiState.Status.FAILURE ->
+                view.announceForAccessibility(context.getString(R.string.focus_failed))
+            FocusMeteringUiState.Status.CANCELLED ->
+                view.announceForAccessibility(context.getString(R.string.focus_released))
+            null -> {}
+        }
+    }
```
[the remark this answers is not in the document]
temcguir: @gemini-code-assist Compose's `LocalAccessibilityManager` does not actually expose a `.makeAnnouncement(...)` or equivalent method for imperative announcements. It only supports querying timeouts.

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/CaptureScreenComponents.kt
```diff
@@ -931,95 +938,168 @@ fun FlipCameraButton(
...
-                            alpha
+                    .semantics {
+                        if (focusStatusDescription.isNotEmpty()) {
+                            contentDescription = focusStatusDescription
                         }
-                    )
+                    }
+                    .graphicsLayer {
+                        alpha = reticleAlpha
+                    }
```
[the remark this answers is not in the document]
temcguir: Good catch! Deferred the state read to the draw phase using a provider lambda (`pulseAlphaProvider() -> Float`) inside the `.graphicsLayer {}` modifier, which successfully avoids the composition loop and halts the high-frequency UI invalidations during the animation cycle.

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/CaptureScreenComponents.kt
```diff
@@ -931,95 +938,168 @@ fun FlipCameraButton(
...
-            remember(
-                coordinateTransformer.transformMatrix,
-                focusMeteringUiState.surfaceCoordinates
-            ) {
-                Matrix().run {
-                    setFrom(coordinateTransformer.transformMatrix)
-                    invert()
-                    map(focusMeteringUiState.surfaceCoordinates)
-                }
+    }
```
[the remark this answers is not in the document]
temcguir: Done, implemented this side-by-side with the lambda provider below so the state extraction only evaluates during the draw phase!

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/CaptureScreenComponents.kt
```diff
@@ -931,95 +938,168 @@ fun FlipCameraButton(
...
+    androidx.compose.runtime.SideEffect {
+        if (focusMeteringUiState is FocusMeteringUiState.Specified) {
+            lastSpecifiedState = focusMeteringUiState
+        }
+    }
+
+    val activeState =
+        (focusMeteringUiState as? FocusMeteringUiState.Specified)
+            ?: lastSpecifiedState
+            ?: return
```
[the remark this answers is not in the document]
temcguir: Done. Since `SideEffect` evaluates after a successful composition phase, mutating the `MutableState` caching variable continuously scheduled unnecessary re-evaluations under the Snapshot system every 16ms. I swapped it out for a non-reactive inline cache (`remember { arrayOfNulls(1) }`) that won't trigger reactivity ping-pongs.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/FocusMetering.kt
```diff
@@ -130,3 +169,53 @@ private fun SurfaceRequest.createTransformationInfoFlow(
...
+        coroutineScope {
+            val fallbackTimeoutJob = launch {
+                delay(FOCUS_LOCK_FALLBACK_TIMEOUT_MILLIS)
+                this@coroutineScope.cancel(
+                    FallbackTimeoutException("SceneChange fallback triggered")
+                )
+            }
+
+            var consecutiveFrames = 0
+            sceneChangeStatusFlow.first { sceneChangeStatus ->
```
davidjiagoogle: Inside first { ... }, fallbackTimeoutJob.cancel() is invoked on every single frame as long as sceneChangeStatus != null (30–60 FPS) until 3 consecutive scene change frames are found. While cancelling an already-cancelled job is a no-op, consider guarding it so it is only cancelled once:
```
if (fallbackTimeoutJob.isActive) {
    fallbackTimeoutJob.cancel()
}
```

core/camera/src/main/java/com/google/jetpackcamera/core/camera/FocusMetering.kt
```diff
@@ -130,3 +169,53 @@ private fun SurfaceRequest.createTransformationInfoFlow(
...
 }
+
+@RequiresApi(Build.VERSION_CODES.P)
+internal suspend fun awaitClearFocusLock(sceneChangeStatusFlow: Flow<Int?>) {
+    class FallbackTimeoutException(
+        message: String
+    ) : CancellationException(message)
+
+    try {
+        coroutineScope {
```
davidjiagoogle: Calling this@coroutineScope.cancel(FallbackTimeoutException(...)) to break out of the scope when the timeout expires works, but cancelling an enclosing coroutineScope for expected control flow is somewhat unconventional in Kotlin Coroutines. If desired, this can alternatively be modelled by running a concurrent race via select or using withTimeoutOrNull before entering the flow collection once capability is determined. If keeping the current approach, adding a short comment explaining that cancellation is intentionally scoped to abort the first frame collector is recommended.

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/CaptureScreenComponents.kt
```diff
@@ -931,95 +947,177 @@ fun FlipCameraButton(
...
  * to screen coordinates. This should come from [CameraXViewfinder].
  */
+@VisibleForTesting
 @Composable
-private fun FocusMeteringIndicator(
+internal fun FocusMeteringIndicator(
     focusMeteringUiState: FocusMeteringUiState,
     coordinateTransformer: CoordinateTransformer
 ) {
+    val lastSpecifiedState = remember { arrayOfNulls<FocusMeteringUiState.Specified>(1) }
```
davidjiagoogle, line 957: Directly mutating lastSpecifiedState[0] in the body of the composable is a side effect during the composition phase. If composition is discarded or restarted, unexpected state mutations can occur. To record the last specified state safely after a successful composition pass without triggering unnecessary recompositions, consider wrapping the update in a SideEffect:
```
SideEffect {
    if (focusMeteringUiState is FocusMeteringUiState.Specified) {
        lastSpecifiedState[0] = focusMeteringUiState
    }
}
```

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/CaptureScreenComponents.kt
```diff
@@ -931,95 +947,177 @@ fun FlipCameraButton(
...
+    val focusStatusDescription = when (currentStatus) {
+        FocusMeteringUiState.Status.RUNNING -> focusScanningDescription
+        FocusMeteringUiState.Status.SUCCESS -> focusLockedDescription
+        FocusMeteringUiState.Status.FAILURE -> focusFailedDescription
+        FocusMeteringUiState.Status.CANCELLED -> focusReleasedDescription
+        null -> ""
+    }
+
+    // Accessibility: announce focus state transitions to TalkBack
+    val view = LocalView.current
```
davidjiagoogle, line 1014: Currently, Modifier.semantics { liveRegion = LiveRegionMode.Polite } is not present on the reticle, and only view.announceForAccessibility is called. In Compose, using Modifier.semantics { liveRegion = LiveRegionMode.Polite } (or LocalAccessibilityManager.current) is generally preferred over calling LocalView.current.announceForAccessibility directly. If imperative announcements are preferred, either update the PR description to avoid confusion or add liveRegion to the semantics node.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/FocusMetering.kt
temcguir: I refactored `awaitClearFocusLock` into a two-phase flow using `withTimeoutOrNull` for the initial capability probe before transitioning to the indefinite scene change collector. This removes the custom scope cancellation and the `fallbackTimeoutJob` handling.

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/CaptureScreenComponents.kt
temcguir: Done.

ui/components/capture/src/main/java/com/google/jetpackcamera/ui/components/capture/CaptureScreenComponents.kt
temcguir: Sorry the PR description was misleading. We intentionally opted for imperative announcements via `LocalView.current.announceForAccessibility(...)` because the indicator composable is dismissed (`isVisible = false`) when focus transitions to `CANCELLED` or idle. If we relied on a `liveRegion` on the indicator node, it unmounts from the tree before TalkBack can announce "Focus released". I’ve updated the PR description to clarify.

core/camera/src/main/java/com/google/jetpackcamera/core/camera/FocusMetering.kt
temcguir: I refactored `awaitClearFocusLock` into the two-phase `withTimeoutOrNull` flow suggested in your other comment. Since `withTimeoutOrNull` directly manages the timeout lifecycle now, `fallbackTimeoutJob` and manual cancellations are no longer present.

davidjiagoogle (approved):

PR #576, opened 2026-09-18
Fix blank Settings screen after process recreation

### LINKED ISSUE #572: [Bug] Settings screen renders blank after process recreation due to uninitialized CameraSystemConstraints
When the app process is terminated while in the background on the Settings screen (e.g., via OS memory reclamation, developer options "Don't keep activities", or when revoking a permission in System Settings), returning to the app restores `SettingsScreen` from the navigation backstack, but all settings items fail to render. The user is presented with a blank screen showing only the top header ("Settings") and the back button.

Navigating back to `PreviewScreen` and returning to `SettingsScreen` recovers the UI.

## Steps to Reproduce
1. Open the app and navigate to the **Settings** screen.
2. Background the app.
3. Terminate the app process in the background (e.g., revoke any permission in Android System Settings, enable "Don't keep activities" in Developer Options, or run `adb shell am kill com.google.jetpackcamera`).
4. Reopen the app from Recents.
5. Observe the `SettingsScreen`.

## Expected Behavior
`SettingsScreen` should render all available settings options upon process recreation, or display a loading indicator while fetching hardware constraints. Furthermore, if a critical permission (such as `CAMERA`) was revoked while backgrounded, the app should guard navigation and route back to the permissions onboarding flow.

## Actual Behavior
The settings list is completely empty; only the top app bar is drawn.

## Root Cause
1. **Coupled Constraints Lifecycle**: `SettingsViewModel` combines with `constraintsRepository.systemConstraints.filterNotNull()` to construct `SettingsUiState.Enabled`.
2. **In-Memory Singleton Reset**: `SettableConstraintsRepositoryImpl` is an in-memory singleton whose `_systemConstraints` initializes to `null` on new process creation.
3. **Bypassed Preview Initialization**: `constraintsRepository.updateSystemConstraints(...)` is currently only called inside `PreviewViewModel.init` when the viewfinder is active. Because Compose Navigation restores `SettingsScreen` as the top backstack destination after process death, `PreviewScreen` and `PreviewViewModel` never run.
4. **Permanent Flow Suspension**: `constraintsRepository.systemConstraints.filterNotNull()` suspends indefinitely waiting for a non-null emission. `SettingsViewModel.settingsUiState` remains stuck on its initial `SettingsUiState.Disabled` state, which renders an empty `Column` in `SettingsScreen`.
5. **Missing Mandatory Permission Guard**: If `CAMERA` permission was revoked during process termination, `SettingsScreen` has no top-level guard checking whether baseline permissions are still held.

## Proposed Solutions
1. **Decouple Constraint Discovery from the Preview Lifecycle**:
   - Allow `ConstraintsRepository` (or `CameraSystem`) to query camera characteristics directly via `ProcessCameraProvider` on demand in the background. Querying camera characteristics (lenses, supported resolutions, flash, video capabilities) does not require active camera streaming or runtime camera permissions.
2. **Handle Loading / Fallback States in Settings**:
   - Add a proper `Loading` state or display default/fallback constraints in `SettingsViewModel` rather than hanging indefinitely on an empty `SettingsUiState.Disabled`.
3. **Root Navigation Permission Guard**:
   - On process restoration, verify baseline mandatory permissions (`CAMERA`). If missing, clear the backstack and route to `PermissionsRoute` to prevent child screens from restoring in an unexecutable state.

M1n9yu23:
### Description

Fixes #572.

`SettableConstraintsRepositoryImpl` was only ever populated by `PreviewViewModel`, so when the process was recreated with `SettingsScreen` on top of the back stack, `systemConstraints` stayed null and the settings list never rendered.

- `CameraXCameraSystemRepository` now implements `ConstraintsRepository`, and `SettableConstraintsRepository` and `ConstraintsModule` are removed. `CameraModule` scopes the concrete repository and provides both `CameraSystemRepository` and `ConstraintsRepository` from it, so they resolve to the same activity-retained instance. `PreviewViewModel` no longer copies constraints; it reads `cameraSystemRepository.systemConstraints` directly for `captureUiState` / `debugUiState`.
- Camera system initialization now starts when `systemConstraints` is first accessed. I went with this instead of the `onStart` + `stateIn` approach I described in the issue: `stateIn` launches a sharing coroutine in the `@Singleton` `@DefaultCoroutineScope`, and that coroutine stays alive regardless of the `SharingStarted` policy, so it outlives the `@ActivityRetainedScoped` repository. `runTest` caught that as an `UncompletedCoroutinesError` on the first attempt. `initializationDeferred.start()` needs no extra coroutine and exposes the camera system's own `StateFlow` as is.
- `SettingsUiState.Disabled` is renamed to `Loading`, and `SettingsScreen` renders a progress indicator for it instead of an empty `Column`.
- The CAMERA check moved from `previewScreen` into a `CameraPermissionGuard` in `JcaApp`. It covers whichever destination is restored, clears the back stack before routing to `PermissionsRoute`, and does nothing while the current destination is already the permissions screen.
- This came up while testing the restore path: once Settings rendered, changes made there never reached the camera system, since the camera had already been initialized before `PreviewViewModel` existed. `PreviewViewModel` now diffs default settings against the snapshot the camera system was initialized with, rather than against its own first emission. The snapshot is exposed as `CameraSystemRepository.getInitialDefaultCameraAppSettings()`.

As discussed in the issue, `SettingsScreen` now initializes CameraX on its own.

`SettableConstraintsRepositoryImpl` was also doubling as a fake in `PreviewViewModelTest`, `CameraAppSettingsViewModelTest` and `CaptureUiStateAdapterTest`. Those now use a new `FakeConstraintsRepository` in `:data:settings:testing` or `FakeCameraSystem.setSystemConstraints`.

### Testing

- `./gradlew test`
- `./gradlew spotlessCheck`
- `./gradlew assembleStableDebug`
- Manually verified the three paths from the issue on an emulator: `adb shell am kill`, "Don't keep activities", and revoking CAMERA in system settings.
- Added unit tests for the lazy initialization, the `Loading` state, and default settings changed before `PreviewViewModel` exists.


data/camera/src/main/java/com/google/jetpackcamera/data/camera/CameraXCameraSystemRepository.kt
```diff
@@ -69,6 +70,7 @@ class CameraXCameraSystemRepository(
     }
 
     override val systemConstraints: StateFlow<CameraSystemConstraints?> by lazy {
+        initializationDeferred.start()
```
temcguir: Because `scope` uses a `SupervisorJob()` (`@DefaultCoroutineScope` in `CommonModule`) and `initializationDeferred` is a root `async` block (`CoroutineStart.LAZY`), any exception thrown inside `cameraSystem.initialize()` is stored inside the `Deferred` and is never delivered to a `CoroutineExceptionHandler` or `uncaughtExceptionHandler` unless `.await()` is called.

When the process is restored directly onto `SettingsScreen`, `getCameraSystem()` isn't called, so if `cameraSystem.initialize()` fails, nothing logs the error and `systemConstraints` stays `null` forever (leaving `SettingsScreen` stuck on the loading spinner). Could we log/handle initialization failures inside the `async` block (or launch a coroutine that awaits/logs on error) so failures aren't silently swallowed?

data/camera/src/main/java/com/google/jetpackcamera/data/camera/CameraXCameraSystemRepository.kt
```diff
@@ -83,22 +85,27 @@ class CameraXCameraSystemRepository(
     private val _cameraPropertiesJSON = MutableStateFlow<String?>(null)
     override val cameraPropertiesJSON: StateFlow<String?> = _cameraPropertiesJSON.asStateFlow()
 
-    private val initializationDeferred: Deferred<Unit> =
+    private val initializationDeferred: Deferred<CameraAppSettings> =
```
temcguir: Minor ordering nit: `systemConstraints` references `initializationDeferred` on line 73 before `initializationDeferred` is declared here on line 88. That only compiles because `by lazy` defers evaluation into a lambda. Could you move `initializationDeferred` above `systemConstraints`?

feature/settings/src/main/java/com/google/jetpackcamera/settings/SettingsScreen.kt
```diff
@@ -141,13 +147,23 @@ private fun SettingsScreen(
...
+            ) {
+                CircularProgressIndicator(modifier = Modifier.size(50.dp))
+            }
+
+            is SettingsUiState.Enabled -> Column(
+                modifier = Modifier
+                    .padding(innerPadding)
+                    .verticalScroll(rememberScrollState())
+                    .background(color = MaterialTheme.colorScheme.background)
+            ) {
```
temcguir: Heads up that `#563` recently refactored `SettingsScreen` to take composable slots (`cameraSettingsSlot`, `recordingSettingsSlot`, `appSettingsSlot`) and deleted `SettingsList`, where each slot currently has an early return `val enabledState = uiState as? SettingsUiState.Enabled ?: return`.

When you rebase onto `main`, `SettingsScreen` can check `when (uiState)` to render the loading indicator for `SettingsUiState.Loading` and the slot `Column` for `SettingsUiState.Enabled`, which also lets you remove the per-slot `as? SettingsUiState.Enabled ?: return` guards.

feature/settings/src/main/java/com/google/jetpackcamera/settings/SettingsScreen.kt
```diff
@@ -141,13 +147,23 @@ private fun SettingsScreen(
...
-            if (uiState is SettingsUiState.Enabled) {
+        when (uiState) {
+            SettingsUiState.Loading -> Box(
+                modifier = Modifier
+                    .padding(innerPadding)
+                    .fillMaxSize()
+                    .background(color = MaterialTheme.colorScheme.background),
+                contentAlignment = Alignment.Center
+            ) {
+                CircularProgressIndicator(modifier = Modifier.size(50.dp))
```
temcguir: Small style nit: please extract `50.dp` on `Modifier.size(50.dp)` into a file-level constant (e.g. `private val LOADING_INDICATOR_SIZE = 50.dp`) or use the default `CircularProgressIndicator()` size.

data/settings/src/main/java/com/google/jetpackcamera/settings/ConstraintsRepository.kt
```diff
@@ -16,26 +16,8 @@
...
 
 import com.google.jetpackcamera.settings.model.CameraSystemConstraints
-import javax.inject.Inject
-import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
-import kotlinx.coroutines.flow.asStateFlow
 
 interface ConstraintsRepository {
     val systemConstraints: StateFlow<CameraSystemConstraints?>
 }
```
temcguir: Could you add brief KDoc comments to `ConstraintsRepository` and `systemConstraints`?

data/settings/testing/src/main/java/com/google/jetpackcamera/settings/testing/FakeConstraintsRepository.kt
```diff
@@ -0,0 +1,34 @@
...
+class FakeConstraintsRepository(
+    initialConstraints: CameraSystemConstraints? = null
+) : ConstraintsRepository {
+    private val _systemConstraints = MutableStateFlow(initialConstraints)
+    override val systemConstraints: StateFlow<CameraSystemConstraints?> =
+        _systemConstraints.asStateFlow()
+
+    fun setSystemConstraints(systemConstraints: CameraSystemConstraints?) {
+        _systemConstraints.value = systemConstraints
+    }
```
temcguir: Could you add brief KDoc comments to `FakeConstraintsRepository` and `setSystemConstraints`?

temcguir (changes requested):

